PR #57, opened 2014-11-03
Set initial mode as preferred.

michalsrb:
If there is any resolution specified with -geometry or -screen parameters,
report this resolution as preferred one. That way desktop environments won't
change it immediately after start.

Most desktop environments like to do some equivalent of "xrandr --output VNC-0 --auto" on start. Any user specified resolution is then replaced with the highest available one: 1920x1200
This patch marks user specified resolution (either a standard one, or new custom one) as preferred.


twaugh:
Is work required on this?

CendioOssman:
Yes. This only covers -geometry given on the command line to Xvnc. The more common case of the client resizing the session is still not handled.

CendioOssman:
I shuffled things around and got a new version in c08aeb6. This one makes sure to recreate the mode list whenever a client changes anything. It's a bit heavy handed, but I don't think it's any worse than before.

PR #94, opened 2014-12-26
Remove support for openSUSE 11.4

bphinz:
Project Evergreen is no longer providing security updates for Xorg source.


CendioOssman:
Huh? Weren't those packages for SUSE Enterprise rather than OpenSUSE?

bphinz:
Initially, yes.  At some point though I lost my subscription and switched the build hosts over to openSUSE.  I probably should have changed the folder name & build tag on the rpms at that point to better reflect the change, but in any case I don't have access to the Enterprise version anymore.  FWIW, 1.4.0 (all couple days of it) was the only release that we ever included these in.

PR #101, opened 2015-01-14
Add user-specified geometry to RandR list.

twaugh:
When Xvnc is started with a '-geometry' parameter, the specified geometry is not added to the RandR list. This causes problems when running e.g. GNOME, which adjusts resolution based on RandR.

How to test:

vncserver -geometry 1440x900 :1
DISPLAY=:1 randr | grep 1440

or run a GNOME session and see what resolution you get when connecting with 'vncviewer -RemoteResize=0'.


twaugh:
Oh, never mind, just saw pull request #57.

PR #103, opened 2015-01-20
Allow java viewer to load without manifest properly

hifi:
The java viewer would not run without a manifest file and caused an exception that is not an IOException like the original catch was. Bugged me while writing another thing.

### DIFF
--- a/java/com/tigervnc/vncviewer/CConn.java
+++ b/java/com/tigervnc/vncviewer/CConn.java
@@ -781,7 +781,7 @@ public class CConn extends CConnection implements
       Attributes attributes = manifest.getMainAttributes();
       pkgDate = attributes.getValue("Package-Date");
       pkgTime = attributes.getValue("Package-Time");
-    } catch (IOException e) { }
+    } catch (java.lang.Exception e) { }
 
     Window fullScreenWindow = Viewport.getFullScreenWindow();
     if (fullScreenWindow != null)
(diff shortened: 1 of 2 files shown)


hifi:
Updated to include the about dialog as well. The null version entries are unhelpful but still better than a silent exception.

bphinz:
Thanks for these patches.  Hopefully I will have some time to start looking
them tonight.  The resize patch looks useful, but that can be a tricky
thing to handle in the java viewer because of the different use modes
(embedded applet/standalone), and fullscreen is a bit flakey on some
platforms so the resize logic is overly complex in some spots.

-brian

hifi:
The UI bits are required for the resize stuff to take action since they are disabled by default so they shouldn't affect embedded use. That includes resize to local screen that hasn't been rebased yet which is quite tricky to do properly and it might need a few iterations.

bphinz:
Can you provide some detail about how this error presented?  What do you mean about running without a manifest file?

Thanks,
-brian

hifi:
If you build and run the classes directly without wrapping everything in a JAR. That's a bit faster when you dev with an IDE that automatically recompiles classes on-the-fly.

hifi:
Any reason why this hasn't been merged? It's a trivial fix for trivial issue.

Thanks.

bphinz:
Yep, should've been merged long ago.  Sorry.

PR #119, opened 2015-02-13
Fix performance regressions

### LINKED ISSUE #63: Severe performance regression in TigerVNC 1.4 relative to 1.2 and 1.3
Unfortunately, I find that the refactored encoder and decoder have significantly regressed in performance relative to TigerVNC 1.3.x-- by about 20-30% on the encoding side and 30-40% on the decoding side.  I consider this to be a bug, as it affects the suitability of TigerVNC as a solution for VirtualGL.

Some notes:
* I notice that the subrectangle mix is a bit different in the new encoder, and this may be partly responsible for the difference in performance.  Also note that the compression ratio isn't as good with the new encoder.
* All results were visually verified-- that is, before the benchmark run, I ran `compare-encodings` with the -o option to generate encoded session files, and I played those back through vncviewer to verify that they were correct.  This was done both with the CUT active and inactive, and for both the TigerVNC 1.2/1.3 encoder and the new 1.4 encoder.
* The same encoded datasets were used to test both versions of the viewer, so as to eliminate any differences in the encoder that might have affected those tests.

If you want to reproduce my results, you will need:

Encoder tests:
* The TurboVNC Benchmark Tools: https://sourceforge.net/p/turbovnc/code/HEAD/tree/vncbenchtools/trunk/ (these now contain a mockup of both the TigerVNC 1.2/1.3 encoder and the next-gen TigerVNC 1.4 encoder.)
* The datasets:  http://www.turbovnc.org/rfbsessions.tar.bz2
* My convenience scripts for running the tests and parsing the output into a spreadsheet-friendly form:
http://www.turbovnc.org/files/TurboVNC-Benchmark-Tools-Scripts.tar.bz2

Decoder tests:
* Turbo-encoded versions of the datasets: http://www.turbovnc.org/rfbsessions.turbo.hiqual.tar.bz2  (NOTE: these were all generated using the TurboVNC Benchmark Tools with the -o option.)
* Patches to the TigerVNC 1.3.x and 1.4.x viewers to enable them to play back these processed datasets and generate performance profiling information: http://www.turbovnc.org/files/TigerVNC-Viewer-Benchmark-Patches.tar.bz2
* My convenience scripts for running the tests and parsing the output into a spreadsheet-friendly form: http://www.turbovnc.org/files/TigerVNC-Viewer-Benchmark-Scripts.tar.bz2

<bountysource-plugin>

---
Want to back this issue? **[Place a bounty on it!](https://www.bountysource.com/issues/5837937-severe-performance-regression-in-tigervnc-1-4-relative-to-1-2-and-1-3?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github).
</bountysource-plugin>

CendioOssman:
Fix some performance regressions that were introduced when we cleaned up the codecs. Fixes #63.

### DIFF
--- a/cmake/StaticBuild.cmake
+++ b/cmake/StaticBuild.cmake
@@ -47,14 +47,18 @@ if(BUILD_STATIC)
 
     set(GNUTLS_LIBRARIES "${GNUTLS_LIBRARIES} -Wl,-Bdynamic")
 
-    # GnuTLS uses various crypto-api stuff
     if (WIN32)
+      # GnuTLS uses various crypto-api stuff
       set(GNUTLS_LIBRARIES "${GNUTLS_LIBRARIES} -lcrypt32")
+      # And sockets
+      set(GNUTLS_LIBRARIES "${GNUTLS_LIBRARIES} -lws2_32")
     endif()
 
-    # nanosleep() lives here on Solaris
     if(${CMAKE_SYSTEM_NAME} MATCHES "SunOS")
+      # nanosleep() lives here on Solaris
       set(GNUTLS_LIBRARIES "${GNUTLS_LIBRARIES} -lrt")
+      # and socket functions are hidden here
+      set(GNUTLS_LIBRARIES "${GNUTLS_LIBRARIES} -lsocket")
     endif()
 
     # GnuTLS uses gettext and zlib, so make sure those are always
--- a/common/rdr/CMakeLists.txt
+++ b/common/rdr/CMakeLists.txt
@@ -4,6 +4,7 @@ add_library(rdr STATIC
   Exception.cxx
   FdInStream.cxx
   FdOutStream.cxx
+  FileInStream.cxx
   HexInStream.cxx
   HexOutStream.cxx
   InStream.cxx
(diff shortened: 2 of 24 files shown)


dcommander:
Will retest on my end

dcommander:
encperf:

For the 24-bit datasets, encperf generates similar results to my hacked-up compare-encodings benchmark, but on the 16-bit and 8-bit datasets, the raw equivalent byte count appears to be off (I'll generate a new issue for that.)  I've generated PR #124 to add the display of the encoded byte count and raw-equivalent byte count to encperf's output (I need this data, because I often want to compute compression ratios for groups of datasets.)

decperf:

Everything LGTM.  Performance numbers match the decoding numbers from my custom benchmark code.

dcommander:
Actually, I now see what the issue is with the raw-equivalent bytes.  The computation of them is using the client-side pixel format instead of the server-side pixel format, and the client-side pixel format in encperf is always 32-bit.  compare-encodings behaves differently.  It will use the same pixel format for client and server unless you are generating a session capture, in which case it will always convert to 24-bit.  I would propose that encperf behave similarly, or at least that an option be added to it so that it can use the same server & client pixel format, even if that behavior isn't the default.

PR #121, opened 2015-02-18
Xvnc: abort argument processing on error.

twaugh:
This prevents e.g. 'Xvnc -screen x' crashing.

### DIFF
--- a/unix/xserver/hw/vnc/xvnc.c
+++ b/unix/xserver/hw/vnc/xvnc.c
@@ -408,7 +408,7 @@ ddxProcessArgument(int argc, char *argv[], int i)
 	    {
 		ErrorF("Invalid pixmap depth %d\n", depth);
 		UseMsg();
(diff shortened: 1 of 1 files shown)


unix/xserver/hw/vnc/xvnc.c
```diff
@@ -391,13 +400,15 @@ ddxProcessArgument(int argc, char *argv[], int i)
...
-	if (++i >= argc) UseMsg();
+	fail_unless_args(argc, i, 1);
+	++i;
 	while ((i < argc) && (depth = atoi(argv[i++])) != 0)
 	{
 	    if (depth < 0 || depth > 32)
 	    {
 		ErrorF("Invalid pixmap depth %d\n", depth);
 		UseMsg();
+		return;
```
CendioOssman: Should be `return 0;` I guess?

PR #124, opened 2015-02-21
Tweaks to encperf to add some functionality that exists in the TurboVNC Benchmark Tools

dcommander:


CendioOssman:
Looks good, but the commit messages are a bit broken with regard to git. Could you rebase them and adjust the messages?

dcommander:
Broken in what way?

dcommander:
More specifically, I think it's github that's broken.  The convention is to provide a summary on the first line of the commit, which I did.

CendioOssman:
The first lines are way too long, which makes them spill over. This happens both with github and git itself:

```
commit ad27d29482cbbfb2e21393b05b7a5e5987dc8d24
Author: DRC <information@virtualgl.org>
Date:   Sat Feb 21 12:01:47 2015 -0600

    encperf: explicitly add the compression level to the list of encodings so it
's obvious how to override it

commit e461b0f0d246aa25f9ae2c4d92363ec44f4bb454
Author: DRC <information@virtualgl.org>
Date:   Sat Feb 21 11:57:27 2015 -0600

    encperf: allow number of iterations to be specified

commit bd0ca8ce60137d31dcd7046e4ff65c1fff1a144c
Author: DRC <information@virtualgl.org>
Date:   Sat Feb 21 11:44:26 2015 -0600

    encperf: report encoded bytes and raw-equivalent bytes as well as compressio
n ratio I like to have these statistics so I can compute overall compression rat
ios for groups of datasets in a spreadsheet.
```

I believe the rule is to keep the first line under 50 characters. E.g.: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

dcommander:
Fixed.  I also added a "translate" option to disable pixel translation, so as to make the output for 16-bit and 8-bit datasets match that of the TurboVNC Benchmark Tools, and corrected a warning in the printf format revealed when compiling with clang.

CendioOssman:
Great. Looks good.

PR #126, opened 2015-02-24
Avoid warning when compiling with gcc's -Wformat-security option.

twaugh:
The -Wformat-security option warns whenever a non-const string is given as a format string.


CendioOssman:
I don't see what purpose that log line has? Just remove it instead?

twaugh:
It shows what the local port is, which otherwise doesn't appear in logging. Perhaps that's not useful though?

CendioOssman:
I'd say that's an internal detail. So until someone can state why that needs to be logged (by default even), then I say remove it.

(and even if it should be logged, it definitely should have more in the log message to give some context)

twaugh:
Sounds fair. Removed.

PR #127, opened 2015-02-24
Add support for X server 1.17.X

michalsrb:
* xalloc and xfree were removed after being marked as deprecated for long time.
* unix/xserver117.patch is a copy of unix/xserver116.patch which works fine with X server 1.17.


twaugh:
Didn't you find you needed to also add libos.la to the libvnc.la link? I get symbol resolution failures here if not.

CendioOssman:
#134 got merged, so this one is no longer needed.

PR #128, opened 2015-02-24
Don't use GetMaster() as that symbol is no longer exported by xorg-x11.

twaugh:


CendioOssman:
Won't this break with older servers though?

And the PR should be against master. We'll move anything relevant over to the stable branches if needed.

twaugh:
I'll make a new PR against master then.

I'm not sure GetMaster() has ever been exported. It has never had _X_EXPORT in its declaration.

PR #129, opened 2015-02-24
Link libvnc.so with '-z now' to make symbol resolution failures obvious.

twaugh:


CendioOssman:
This is also against the wrong branch. I don't know if you need to create a new PR, or if you can tweak this one.

twaugh:
I'll make a new PR.

PR #130, opened 2015-02-24
Updated to build against xorg-x11-server 1.17.x.

twaugh:
Here's a change on 1.4-branch to allow building against xorg-x11-server-1.17.x. The xserver116.patch was re-based against xorg-x11-server-1.17.1.

I've changed the xmalloc/xfree calls, no longer provided by xorg-x11, to use calloc().

See also #127.


twaugh:
I'll close this, as it's against 1.4-branch. I've added a new one for master.

PR #131, opened 2015-02-24
Link the module against libos.la.

twaugh:
Without doing this, and when linking with '-z now', I see:
(EE) Failed to load /usr/lib64/xorg/modules/extensions/libvnc.so: /usr/lib64/xorg/modules/extensions/libvnc.so: undefined symbol: _Z10fileexistsPc

This is branched from 1.4-branch.

### DIFF
--- a/unix/xserver/hw/vnc/Makefile.am
+++ b/unix/xserver/hw/vnc/Makefile.am
@@ -5,6 +5,7 @@ RFB_LIB=$(LIB_DIR)/rfb/librfb.la
 RDR_LIB=$(LIB_DIR)/rdr/librdr.la
 NETWORK_LIB=$(LIB_DIR)/network/libnetwork.la
 XREGION_LIB=$(LIB_DIR)/Xregion/libXregion.la
+OS_LIB=$(LIB_DIR)/os/libos.la
 COMMON_LIBS=$(NETWORK_LIB) $(RFB_LIB) $(RDR_LIB) $(XREGION_LIB)
 
 # Hack to get the C headers to work when included from C++ code
@@ -58,7 +59,7 @@ libvnc_la_CPPFLAGS = $(XVNC_CPPFLAGS) -I$(TIGERVNC_SRCDIR)/common -UHAVE_CONFIG_
 
 libvnc_la_LDFLAGS = -module -avoid-version
 
-libvnc_la_LIBADD = libvnccommon.la $(COMMON_LIBS)
+libvnc_la_LIBADD = libvnccommon.la $(COMMON_LIBS) $(OS_LIB)
 
 EXTRA_DIST = Xvnc.man
 


CendioOssman:
That seems like a bug somewhere. `fileexists()` is only called from `CSecurityTLS`, which is a class that should only be present in the client. Can you figure out why it is getting pulled in? I'm not seeing it here in my build.

twaugh:
Well, fileexists() is used by CSecurityTLS, which is built into librfb, which libvnc links against. So the symbol is not resolved unless libos is also linked in.

I guess you're not seeing it because you aren't linking with '-z now'. In this particular instance, the symbol should never be needed, but without '-z now' other symbol resolution problems may be hidden until resolution time.

CendioOssman:
I checked the binaries. There is no such symbol in my output files. And yes, it is in librfb.a. But it's not like all of it gets pulled into libvnc.so (or Xvnc). Only the parts that resolve needed symbols are picked.

Try building a librfb.a without CSecurityTLS and see what unresolved symbols you get?

bphinz:
I definitely have to apply that same patch when building several (all?) of
the distro packages or else there are unresolved symbols in Xvnc.

twaugh:
It only comes from CSecurityTLS.

```
$ for a in common/rfb/CMakeFiles/rfb.dir/*.o; do nm $a | grep fileexists && echo $a; done
                 U _Z10fileexistsPc
common/rfb/CMakeFiles/rfb.dir/CSecurityTLS.cxx.o
```
```

**Comment by CendioOssman** [comment]
I can't reproduce this, so I cannot debug it. One of you guys needs to follow the dependency chain and see why CSecurityTLS is getting pulled in when it shouldn't. Just remove it from librfb.a and try a link and the linker should tell you why. If you get yet another client class then repeat the process until you strike gold. :)

**Comment by CendioOssman** [comment]
Any progress on this?

**Comment by bphinz** [comment]
On Tue, Mar 17, 2015 at 12:27 PM, Pierre Ossman (Work account) <

> Any progress on this?
>
I have not had a chance to look at it yet.

-brian

**Comment by CendioOssman** [comment]
So I had a look at this by building the el6 SRPM for the current nightly. Some first notes:

 - It requires "cmake28" rather than just "cmake". The former is not in RHEL, whilst the latter is. Same thing for the %{cmake28} macros.
 - It lacks a requirement on g++. Perhaps that is automatically pulled in by mock and such.

Anyhoo, I found the actual issue. It's some delightful behaviour from libtool where it includes `librfb.a` with `--whole-archive`. This pulls in everything, not just what it needs.

**Comment by CendioOssman** [comment]
Apparently this is how "convenience libraries" are defined. They don't really state how it separates those from "normal" static libraries though. But digging through the source code I figured out that it depends on if `libdir` is set in the .la file. Setting that makes the link behave sane.

So, the proper fix for this is:

```diff
diff --git a/cmake/Modules/CMakeMacroLibtoolFile.cmake b/cmake/Modules/CMakeMacroLibtoolFile.cmake
index f2ea7f8..a54d994 100644
--- a/cmake/Modules/CMakeMacroLibtoolFile.cmake
+++ b/cmake/Modules/CMakeMacroLibtoolFile.cmake
@@ -130,7 +130,7 @@ macro(libtool_create_control_file _target)
   file(APPEND ${_laname} "shouldnotlink=no\n\n")
   file(APPEND ${_laname} "dlopen=''\n")
   file(APPEND ${_laname} "dlpreopen=''\n\n")
-  file(APPEND ${_laname} "libdir=''\n\n")
+  file(APPEND ${_laname} "libdir='/usr/lib'\n\n")
 
 
   # Add custom command to symlink the static library so that autotools finds
```

bphinz:
On Wed, Mar 18, 2015 at 12:10 PM, Pierre Ossman (Work account) <

> So I had a look at this by building the el6 SRPM for the current nightly.
> Some first notes:
>
>    - It requires "cmake28" rather than just "cmake". The former is not in
>    RHEL, whilst the latter is. Same thing for the %{cmake28} macros.
>
> cmake28 is in EPEL.  The cmake supplied by RHEL is cmake2.6 which
according to BUILDING.txt (and personal experience) is too old.

>
>    - It lacks a requirement on g++. Perhaps that is automatically pulled
>    in by mock and such.
>
> Thanks.  I'll add it to the requirements.

CendioOssman:
I got cmake 2.8 in RHEL 6. Perhaps they've updated it at some point?

twaugh:
How about on architectures whose libdir is /usr/lib64?

bphinz:
On Wed, Mar 18, 2015 at 12:42 PM, Pierre Ossman (Work account) <

> I got cmake 2.8 in RHEL 6. Perhaps they've updated it at some point?
>
How about that!  It's even newer than the one in epel...  Thanks for
pointing that out, I'll update that as well.

-brian

CendioOssman:
> How about on architectures whose libdir is /usr/lib64?

I don't think it actually uses the value for anything. We don't ever install those libraries. It just needs to be non-zero. And given libtool's colourful history with lib/lib64, I think lib is the safer option.

CendioOssman:
Have you guys tested my suggestion so we can resolve this issue?

twaugh:
Sorry, not yet. I'll try to get to that soon.

twaugh:
Yes, that suggestion works here.

CendioOssman:
Committed in de2df7e6. Closing this PR.

PR #135, opened 2015-02-25
Don't use GetMaster() as that symbol is no longer exported by xorg-x11.

twaugh:
This is against master, replacing #128 (which was against 1.4-branch).

### DIFF
--- a/unix/xserver/hw/vnc/InputXKB.c
+++ b/unix/xserver/hw/vnc/InputXKB.c
@@ -212,10 +212,7 @@ void vncPrepareInputDevices(void)
 
 unsigned vncGetKeyboardState(void)
 {
-	DeviceIntPtr master;
-
-	master = GetMaster(vncKeyboardDev, KEYBOARD_OR_FLOAT);
-	return XkbStateFieldFromRec(&master->key->xkbInfo->state);
+	return XkbStateFieldFromRec(&vncKeyboardDev->master->key->xkbInfo->state);
 }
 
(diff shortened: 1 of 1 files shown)


CendioOssman:
This won't work on anything older than 1.10 as far as I can tell.

twaugh:
Just saw this:
http://lists.x.org/archives/xorg-devel/2015-February/045659.html

So maybe we can drop this PR after all?

CendioOssman:
Works for me. I assume everyone out there is already doing some patches to the current versions of Xorg.

bphinz:
On Thu, Feb 26, 2015 at 10:51 AM, Pierre Ossman (Work account) <

> Works for me. I assume everyone out there is already doing some patches to
> the current versions of Xorg.
>
So should I be applying that patch to our builds?

-brian

CendioOssman:
I assume so. I figured we did already or libvnc.so would never work. Or maybe this isn't an issue with older Xorg servers.

bphinz:
On Tue, Mar 17, 2015 at 12:40 PM, Pierre Ossman (Work account) <

> I assume so. I figured we did already or libvnc.so would never work. Or
> maybe this isn't an issue with older Xorg servers.
>
OK, will do.

CendioOssman:
Hang on. No, you should not apply this. It is only relevant for libvnc.so. And it only affects Xorg, not Xvnc or libvnc.so. And since we do not ship Xorg, we cannot fix anything. It's the distributions that need to have the patch.

PR #141, opened 2015-03-01
Don't violate RFB spec by sending UTF-8

bphinz:
Java viewer was violating the RFB specification by sending UTF-8 encoded clipboard text rather than ISO-8859-1.

### DIFF
--- a/java/com/tigervnc/rfb/CMsgWriter.java
+++ b/java/com/tigervnc/rfb/CMsgWriter.java
@@ -20,6 +20,8 @@
 
 package com.tigervnc.rfb;
 
+import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
 import com.tigervnc.rdr.*;
 
 abstract public class CMsgWriter {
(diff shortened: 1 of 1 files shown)


CendioOssman:
I assume that `String` is some kind of unicode data type? If so, won't it throw exceptions if you try to convert to latin-1 and they symbol cannot be converted?

Also, you need to fix the incoming clipboard handling in `CMsgReader` as well.

bphinz:
On Mon, Mar 2, 2015 at 4:00 AM, Pierre Ossman (Work account) <

> I assume that String is some kind of unicode data type? If so, won't it
> throw exceptions if you try to convert to latin-1 and they symbol cannot be
> converted?
>
Surprisingly, Charset.encode() does not throw an exception.  For malformed
or unmappable characters it falls back to a default replacement string ('?'
IIRC).

> Also, you need to fix the incoming clipboard handling in CMsgReader as
> well.
>
I left it that way intentionally because I (incorrectly) thought that UTF-8
was a superset of ISO-8859-1.  I see now that's only true for the first 127
characters.  Thanks.

Issue #108, opened 2015-01-23
IPv6 breaks TCP filter (HostsFile)


austinbutler:
This is how I am starting x0vncserver:
```
/usr/bin/x0vncserver -display=:0 -hostsfile=/home/austin/.vnc/hosts -passwordfile=/home/austin/.vnc/passwd
```

This is the contents of the hosts file:
```
+192.168.1.2
```

When I try to connect from that IP, this is the result:
```
Thu Jan 22 16:16:16 2015
 Geometry:    Desktop geometry is set to 1920x1080+0+0
 Main:        XTest extension present - version 2.2
 Main:        Listening on port 5900

Thu Jan 22 16:16:20 2015
 Main:        Client connection rejected
```

Why is that? My real aim is to have localhost and that IP allowed, so either from my desktop on the internal network or, if outside, through an SSH tunnel.

But I tried this, and it didn't work so I tried the very simple rule above:
```
+192.168.1.2
+127.0.0.1
-
```

Everything works fine without the hostsfile parameter. This is on Arch with TigerVNC built from Github.

CendioOssman:
I can't see anything obviously wrong with your file. Could you run `x0vncserver` with `-Log *:stderr:100` and see if it tells us anything more?

austinbutler:
It reads out the parameters, but otherwise the error is the same unfortunately.

```
Sat Jan 24 10:42:33 2015
 Config:      set display(String) to :0
 Config:      set HostsFile(String) to /home/austin/.vnc/hosts
 Config:      set PasswordFile(String) to /home/austin/.vnc/passwd
 Geometry:    Desktop geometry is set to 1920x1080+0+0
 Main:        XTest extension present - version 2.2
 VNCServerST: creating single-threaded server x0vncserver
 Main:        Listening on port 5900

Sat Jan 24 10:42:44 2015
 Main:        Client connection rejected
```

CendioOssman:
Ah. Found it. It's the new IPv6 support that is causing problems. The filter code only supports IPv4. 

@twaugh I guess this needs some attention as well. :)

twaugh:
Thanks, I'll take a look.

CendioOssman:
Closing this now as we've merged the proper fix.

Issue #77, opened 2014-11-21
Xtigervnc fails to start VNC extension with -localhost ; assumes IPv6 is present


osd1000:
Ubuntu 14.04.1 LTS
Package version 1.3.90+20141120git59c5a55c-3ubuntu1

When I start Xvnc with the -localhost argument, I get the error:

vncext:      vncExtInit: unable to bind listening socket: Cannot assign
              requested address (99)

Using strace I see:
bind(6, {sa_family=AF_INET6, sin6_port=htons(5903), inet_pton(AF_INET6, ":
:1", &sin6_addr), sin6_flowinfo=0, sin6_scope_id=0}, 28) = -1 EADDRNOTAVAIL (Can
not assign requested address)

...which will not work on this system because IPv6 is disabled on all network interfaces.

Thanks,

Owen

CendioOssman:
I've added parameters to disable IPv6 now. Should probably still implement the dual socket approach, but this provides a workaround.

twaugh:
I have work in progress on this, but have been pressed for time this month due to other issues.

twaugh:
I've been working on this today. Here's where I'm up to:
https://github.com/twaugh/tigervnc/commits/multisocket

The TcpFilter class still needs some work, but what do you think about the general approach?

CendioOssman:
A whole bunch of comments added on the commits. :)

twaugh:
I've updated (forced update) the branch, addressing your comments. See what you think.

CendioOssman:
Some minor comments, but in general it looks very good.

CendioOssman:
Closing this now as we've merged the proper fix.

PR #150, opened 2015-03-13
IPv6 support using per-family sockets (#77, #108)

twaugh:
Here's a rebased version of the 'multisocket' branch, with fixes from local testing. It needs work for win32 targets.

I've tested it with an IPv6-capable host on an IPv4-only interface, and on an IPv4-only Linux host (booted with ipv6.disabled=1).


twaugh:
Should mention: this is intended to fix both #77 and #108.

PR #161, opened 2015-04-01
Warn about not reseting Xvnc and terminate instead.

michalsrb:
Some time ago I was debugging problem with "Xvnc -query hostname": You get login screen, can log in and out, but then Xvnc doesn't display login screen again. It just keeps running displaying nothing.

It is because the internal reset is intentionally prevented. When allowed, Xvnc and especially normal X server with VNC extension fail in various ways during the reset. I gave up on trying to fix the resetting for now.

I think Xvnc should warn about not supporting reset. Also in my opinion it is better to terminate than do nothing when reset is expected. User can still use "-noreset" to prevent the resetting/terminating.

### DIFF
--- a/unix/xserver/hw/vnc/xvnc.c
+++ b/unix/xserver/hw/vnc/xvnc.c
@@ -1566,7 +1566,12 @@ vfbScreenInit(ScreenPtr pScreen, int argc, char **argv)
 
 
 static void vfbClientStateChange(CallbackListPtr *a, void *b, void *c) {
-  dispatchException &= ~DE_RESET;
+    if (dispatchException & DE_RESET) {
+        ErrorF("Warning: VNC extension does not support -reset, terminating instead. Use -noreset to prevent termination.\n");
+
+        dispatchException |= DE_TERMINATE;
+        dispatchException &= ~DE_RESET;
+    }
 }
  
 #if XORG >= 113


CendioOssman:
In principle this is probably a good idea. But this change will just break so many systems that are currently working.

What do you think about switching the default to `-noreset` at the same time?

michalsrb:
Having `-noreset` as default would keep the behavior similar to what it is now. On the other hand normal X server has `-reset` as default, so that would make Xvnc different.

The situation that I want to improve (Xvnc with `-query` hanging in weird state after log out) gets improved in both cases, using `-query` forces restart no matter if `-reset` or `-noreset` is given. So this fix would still work.

If you think it's better to make `-noreset` default, I'll add that.

CendioOssman:
I suspect most people have no clue about the whole reset/noreset thing. All they want is that the server keeps on running, so that's what we should preserve.

michalsrb:
Ok, the best way to change the default would be changing `char dispatchExceptionAtReset = DE_RESET;` to `char dispatchExceptionAtReset = 0;`. That is file from xserver, so the change would have to go to the unix/xserver*.patch patches. I guess I shouldn't just append new chunk to them. Are those patches generated from some public branch of xserver?

CendioOssman:
The diffs should be against the public releases of Xorg. They might not be consistently updated against the latest micro release though.

Updating the initial value is probably the most correct, yes. But given that we have a bunch of patches, perhaps it's the lesser evil to just override it before we call `ProcessCommandLine()`?

PR #165, opened 2015-04-20
Syslog logging

michalsrb:
This is a syslog logger. Currently enabled for Xvnc, but it could be used in other places as well. The main motivation is to have nice logging when Xvnc is started from xinet.d.

Logger_syslog.cxx and Logger_syslog.h are new files. I am not sure if I placed correct comment to the start of the file.

### DIFF
--- a/common/rfb/CMakeLists.txt
+++ b/common/rfb/CMakeLists.txt
@@ -29,6 +29,7 @@ set(RFB_SOURCES
   Logger.cxx
   Logger_file.cxx
   Logger_stdio.cxx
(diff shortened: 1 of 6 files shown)


CendioOssman:
Looks good, except for the hard coded `"Xvnc"`. You should force a name parameter to `initSyslogLogger()`, like it's done for the file logger.

michalsrb:
Actually I just found out that one can give null to the first parameter of `openlog` and it will use program's name automatically. Would that be ok?

CendioOssman:
Yeah, that works.

PR #175, opened 2015-06-13
Create el.po

skarmoutsosv:


CendioOssman:
Thank you for your contribution. Unfortunately we cannot accept it this way. We've put our translations under the control of the translation project. Because of this the po files are owned by them so any changes needs to go via their systems. Ideally they'd like you to be a part of their team, but I believe they also accept one-off translation. You can find the contact information for their Greek team here: http://translationproject.org/team/el.html

skarmoutsosv:
I am already member of the translation project's greek team and as you suggested, I uploaded the file to the translation project which was accepted. 

I also contribute translations using https://l10n.gnome.org/ and https://www.transifex.com/
So, I was wondering why github can not be one of the translation methods to use.

CendioOssman:
It's a requirement from the translation project that they are the official source of the po files. And I can understand that. Having to keep an eye on every projects' repo for changes would be a lot of hassle.

PR #179, opened 2015-06-25
Key mapping issue for japanese keyboard for mac osx fixed 

brishtiteveja:
There were some key mapping issue with my Japanese keyboard with tiger-vnc. According to this post,
http://woods-squiggle.blogspot.com/2014/12/keyboard-tigervnc-1480-osx.html
I became able to solve it.

### DIFF
--- a/vncviewer/cocoa.mm
+++ b/vncviewer/cocoa.mm
@@ -37,6 +37,71 @@
 
 #define NoSymbol 0
 
+static const int kvk_map_qwerty[][3] = {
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm afraid we cannot accept this code. It changes the viewer to use Apple's incompatible version of VNC, which will screw things up when connecting to a proper VNC server. TigerVNC is first and foremost a VNC project, not an Apple Remote Desktop client.

PR #195, opened 2015-09-15
Allow X clients to change sharing and security options.

michalsrb:
This is meant more like proof-of-concept or suggestion rather than final implementation.

The VNC extension allows X clients (vncconfig for example) to read and change the VNC parameters. But most of the parameters can not be actually changed because of possible security problems. I would like to allow changing of some of the parameters if enabled via extra parameter.

Use case: I am working on a setup where Xvnc is spawned with login screen from display manager and allows single client without any authentication. After some user logs into a session, he could optionally use some tool inside the session to set up security and then allow additional clients to connect.

The list of the allowed parameters is getting bit long and ugly, maybe the `VoidParameter` class could have some extra property telling by what means it can be changed?

### DIFF
--- a/unix/xserver/hw/vnc/vncExt.c
+++ b/unix/xserver/hw/vnc/vncExt.c
@@ -51,6 +51,7 @@ static int vncErrorBase = 0;
 static int vncEventBase = 0;
 
 int vncNoClipboard = 0;
+int vncTrustXClient = 0;
 
 static char* clientCutText = NULL;
 static int clientCutTextLen = 0;
@@ -188,6 +189,17 @@ static int ProcVncExtSetParam(ClientPtr client)
    */
   if (strncasecmp(param, "desktop", 7) != 0 &&
       strncasecmp(param, "AcceptPointerEvents", 19) != 0 &&
+      (!vncTrustXClient || strncasecmp(param, "AlwaysShared", 12) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "NeverShared", 11) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "DisconnectClients", 17) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "SecurityTypes", 13) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "PlainUsers", 10) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "PasswordFile", 12) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "RFBAuth", 7) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "Password", 8) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "MaxDisconnectionTime", 20) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "MaxConnectionTime", 17) != 0) &&
+      (!vncTrustXClient || strncasecmp(param, "MaxIdleTime", 11) != 0) &&
       (vncNoClipboard || strncasecmp(param, "SendCutText", 11) != 0) &&
       (vncNoClipboard || strncasecmp(param, "AcceptCutText", 13) != 0))
     goto deny;
--- a/unix/xserver/hw/vnc/vncExt.c
+++ b/unix/xserver/hw/vnc/vncExt.c
@@ -208,7 +208,7 @@ static int ProcVncExtSetParam(ClientPtr client)
   rep.success = 1;
 
   // Send DesktopName update if desktop name has been changed
-  if (strncasecmp(param, "desktop", 7) != 0)
+  if (strncasecmp(param, "desktop", 7) == 0)
     vncUpdateDesktopName();
 
 deny:
--- a/unix/xserver/hw/vnc/vncExtInit.h
+++ b/unix/xserver/hw/vnc/vncExtInit.h
@@ -36,6 +36,7 @@ extern "C" {
 
 // vncExt.c
 extern int vncNoClipboard;
+extern int vncTrustXClient;
 
 int vncAddExtension(void);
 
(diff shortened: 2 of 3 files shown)


hildred:
I have been contemplating something similar as I have a similar use case, and I will definitely do some playing with this patch, but it does strike me as a bit coarse. Let's throw up a use case: first come first serve with tls and pam and reconnection. We start by assuming company policy requires tls and forbids sharing passwords or accounts (if you wish to allow someone to see your desktop or shell that is fine, but they may not control it, must use their own username and password, and encryption is required). Next I write pam_vnc_r.so a pam session module that disables passwordless authentication and changes the username to the name of the user that just logged in. So we setup Xvnc to require tls, not require passwords and allow changing of security options. we then add pam_vnc_r.so to the desktop login, and when desktop login is completed the user is then allowed to reconnect without allowing other users to, and he is able to allow other users to have view only shared sessions (lets ignore for a moment that this does not work, but that is becouse name based shared sessions are not implemented as they would be pointless without this base infistructure), but he is also allowed to disable tls, enable full shared sessions and otherwise violate policy or do dumb thinks like lockhimself out so he cannot reconnect.

The only reasonable option I can see is to have pluggable policy, where the determination of what is allowed is determined by a user supplied library that is loaded at runtime.

By the way, I will have a proof of concept pam module implementation for you to play with an a day or so.

michalsrb:
Interesting. My use case is not expecting any company policy, the user would be free to set any security settings he wishes. The tool may offer only settings that are secure and that would not lock him out, but if he sets them with vncconfig or with his own tool, then it is his own problem. Nothing is preventing him from starting x0vncserver or similar program anyway. So in this case there is no need for pam module, the tool is simply started in the session's autostart and if it detects vnc extension, it can configure it. Either immediately or after user interaction...

hildred:
The advantage of the pam module is speed. There is an inherent race condition between when the user connects to vnc, logs in and the authentication types change. by using pam that window is narrowed, perhaps if used as an auth module to before the password is entered. doing it as a userspace tool would leave the race window open until the desktop is running.

michalsrb:
My plan is to start Xvnc with something like "-NeverShared=1 -DisconnectClients=0 -SecurityTypes=TLSNone". So that one and only one client can connect at time, others will be refused. Then the user can log in, get the session started and then the sharing can be configured to whichever security he wants and after that is done the NeverShared can be unset to allow others in. So there is no race condition. And the sharing can be enabled at any time, not only after the log in.

michalsrb:
I will summarize why I would like to get this accepted into upstream:

- The infrastructure to configure VNC parameters using the X VNC extension is already in place, but the amount of configurable options is very limited. It would be useful to my use case to allow some additional options.
- Those additional options I am suggesting are related to sharing and security settings; Any X application could already grab screen content and send it to anybody (e.g. x0vncserver), so this doesn't give them more power than they already had. Nevertheless I suggest to allow those extra options only when the trustxclient parameter is given.

Alternatively we could add a parameter that will list which options can be changed over the VNC extension, that would allow more fine-grained control.

CendioOssman:
I'll try to have another look and either merge or comment. :)

CendioOssman:
Looks good. Just two comments really:

 a) Why not allow all settings when this is enabled?
 b) Could we find something more generic than trust**x**client? I'm thinking it would be nice if we could carry this over when we do a Wayland server, or to WinVNC. I can't think of anything on the top of my head though.

michalsrb:
a) My idea was that when the session is trusted, it can manage its own sharing and security parameters. But even then some other parameters should be still protected, because the Xvnc process may run under other user than the session. (E.g. some user that is member shadow group for *Plain securities or even root.) The sessions still should not be able to change parameters like `PAMService` or `Log`.

b) `trustclient`? Or something like `AllowOverride` that takes list of names of parameters that can be changed from inside. The default could be the ones that are allowed now. If that sounds good, I can prepare it.

CendioOssman:
> a) My idea was that when the session is trusted, it can manage its own sharing and security parameters. But even then some other parameters should be still protected, because the Xvnc process may run under other user than the session. (E.g. some user that is member shadow group for *Plain securities or even root.) The sessions still should not be able to change parameters like PAMService or Log.
> 

We need a better definition of "trusted" in this case, because I'm not seeing the difference between being allowed to change `SecurityTypes` and `PAMService`.

> b) trustclient? Or something like AllowOverride that takes list of names of parameters that can be changed from inside. The default could be the ones that are allowed now. If that sounds good, I can prepare it.

`AllowOverride` sounds promising in that case. But a name might also be obvious once a good definition for a) is in place.

michalsrb:
> We need a better definition of "trusted" in this case, because I'm not seeing the difference between being allowed to change SecurityTypes and PAMService.

I think that typically Xvnc runs under different user than the X applications connected to it. Especially when using normal X server with VNC extension in it. Lets consider Xvnc running under root while displaying a desktop session of a regular user.

When the X applications are **not trusted**, then they can not change the VNC sharing and security configuration. This could be for example school computer where teacher can supervise screens of students. Students are not allowed to change almost any of the VNC settings from within their sessions, so they cannot deny the teacher access and can not share it to classmates either. (Current state)

When the X applications are **trusted**, then the VNC is there as a service for the user and he can manage it. The applications that are displaying in Xvnc are trusted to change the security settings from within the session.

In both cases the X applications must not be able to make Xvnc do things which they wouldn't have permissions to do themselves. By changing the `PAMService` parameter the regular user could make the root process Xvnc run pam_authenticate with some other service configuration. What if that configuration contains self-destruct of the computer after 3 failed login attempts... It only depends on what pam modules and configurations are present on the machine.

CendioOssman:
There might be other things to consider under that scenario. E.g. `PasswordFile` probably has to be restricted or that root Xvnc process could be tricked to open arbitrary dangerous files.

It's still a bit fuzzy though. At least to me. So I'm liking the `AllowOverride` idea better in that case. That also lets the administrator choose the minimum set of options to trust users with.

michalsrb:
You are right, `PasswordFile` should be also restricted. I created pull request for the alternative with AllowOverride parameter: #282

CendioOssman:
Great. Closing this one then.

Issue #216, opened 2015-10-12
configuration files for /usr/bin/vncserver


graysky2:
I am writing a systemd service for vncserver (https://github.com/TigerVNC/tigervnc/pull/217) and would like to make it as general as possible (minimal number of switched in the ExecStart line.  My question is: does /usr/bin/vncserver source some config file where various options reside and are called by vncserver upon startup?  For example, I want to avoid passing the `-alwaysshared -localhost -geometry 2000x1200` flags to /usr/bin/vncserver but I do want them capture in a user config file like $HOME/.vnc/config or the like.  Thanks.

CendioOssman:
No there is no such configuration right now.

graysky2:
Would you be opposed to depending on a module to parse a config file?  For example: Config::General

hildred:
@graysky2, you do realize that library is written in perl and that Xvnc is written in C and C++. I don't think that would be the best place for embedding a perl interpreter unless we had additional plans for using it.

bphinz:
I think he's suggesting adding a dependency on Config::General from within
the vncserver perl script.

CendioOssman:
That doesn't seem to be installed by default on at least Fedora, so I'm a bit cautious about depending on it.

It there some simpler module that is included with Perl itself? Or how difficult would it be to include a small parser in the vncserver script?

graysky2:
@pbhinz - Yes, you are indeed correct: only needed for the vncserver perl script.
@CendioOssman - Yes, I am making ready a PR that will have this parsing built into the script without any extra dependencies.  The code works well in my test cases and will allow for easy incorporation of the systemd user service from https://github.com/TigerVNC/tigervnc/pull/217

graysky2:
Ready for review: https://github.com/TigerVNC/tigervnc/pull/221

PR #217, opened 2015-10-12
add systemd service files

graysky2:
Since systemd is ubiquitous these days, upstream should provide reliable service files.  This is my attempt at that.  I'm not sure where you want your Makefile to drop them though since both files require users to edit them.

### DIFF
--- a/contrib/systemd/system/vncserver@:1.service
+++ b/contrib/systemd/system/vncserver@:1.service
@@ -1,0 +1,30 @@
+# The vncserver service unit file system mode
+#
+# 1. Copy this file to /etc/systemd/system/vncserver@:<display>.service
+# 2. Edit User=
+#   ("User=foo")
+# 3. Edit the vncserver parameters appropriately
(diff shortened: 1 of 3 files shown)


CendioOssman:
Have you looked at the service files that the distributions have created? It would be nice if everyone could agree on the same thing. :)

graysky2:
Yes, I have based this PR on the one Fedora has provided (the system service) but I recently discovered a more simplistic method in the user session service.  To me, that is really the way to go.

If https://github.com/TigerVNC/tigervnc/issues/216 is ever implemented, I would recommend simply supplying the user service in `/usr/lib/systemd/user/` which wouldn't require any modification from the user wanting to run it.

graysky2:
I am going to close this PR and make it part of a larger one wherein the service, changes to the perl script and man page are updated for your review.

graysky2:
Superseded by https://github.com/TigerVNC/tigervnc/pull/221

PR #221, opened 2015-10-14
Enable parsing of ~/.vnc/config for vncserver and include a systemd service unit for user mode

graysky2:
PR contains the needed modification to perl script vncserver to parse a config file (~/.vnc/config) rather than hard-coding switches into the invocation of /usr/bin/vncserver.  This is both helpful and needed for the included general systemd service for user mode operation.  Updates to the manpage for vncserver and the corresponding CMakeLists are also included.

### DIFF
--- a/contrib/systemd/user/vncserver@.service
+++ b/contrib/systemd/user/vncserver@.service
@@ -1,0 +1,26 @@
+#
+# /usr/lib/systemd/user/vncserver@.service
+#
+# 1. Switches for vncserver should be entered in ~/.vnc/config rather than
+#    hard-coded into this unit file. See the vncserver(1) manpage.
+#
+# 2. Users wishing for the server to continue running after the owner logs
+#    out MUST enable 'linger' with loginctl like this:
+#    `loginctl enable-linger username`
+#    
+# 3. The server can be enabled and started like this once configured:
+#    `systemctl --user start vncserver@:<display>.service`
+#    `systemctl --user enable vncserver@:<display>.service`
+
+[Unit]
+Description=Remote desktop service (VNC)
+After=syslog.target network.target
+
+[Service]
+Type=forking
+ExecStartPre=/bin/sh -c '/usr/bin/vncserver -kill %i > /dev/null 2>&1 || :'
+ExecStart=/usr/bin/vncserver %i
+ExecStop=/usr/bin/vncserver -kill %i
+
+[Install]
+WantedBy=default.target
--- a/unix/vncserver
+++ b/unix/vncserver
@@ -81,6 +81,17 @@ $defaultXStartup
        "xterm -geometry 80x24+10+10 -ls -title \"\$VNCDESKTOP Desktop\" &\n".
        "twm &\n");
 
+$defaultConfig
+    = ("## Supported server options to pass to vncserver upon invocation can be listed\n".
+       "## in this file. See the following manpages for more: vncserver(1) Xvnc(1).\n".
+       "## Several common ones are shown below. Uncomment and modify to your liking.\n".
+       "##\n".
+       "# securitytypes=vncauth,tlsvnc\n".
+       "# desktop=sandbox\n".
+       "# geometry=2000x1200\n".
+       "# localhost\n".
+       "# alwaysshared\n");
+
 chop($host = `uname -n`);
 
 if (-d "/etc/X11/fontpath.d") {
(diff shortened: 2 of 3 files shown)


bphinz:
I have a couple of issues with this.

(1) I'm pretty strongly opposed to having the systemd service files be installed as a component of the software.  Putting it into the contrib directory is one thing, but installing it alongside Xvnc pretty much makes it a first class citizen that we're then obligated to support.  There are about a half dozen open bugs on red hat's bugzilla specifically related to TigerVNC + systemd, many of which have been open for many months (some more than a year).  If RH/Fedora can't resolve these issues then I just don't see how we have the bandwidth or expertise to do so.  And other distress are bound to have their own quirks too.

(2) The changes to vncserver introduce the ability for users to arbitrarily override settings on non-systemd platforms.  For example, on el6 vncservers might be configured in /etc/sysconfig/vncserver with specific -SecurityType parameters as defined by corporate IT policy.  It seems like those settings could now be superseded by anything set in ~/.vnc/config because it accepts any key=value pairs.  I definitely like the idea of allowing users to manage things like geometry, etc. but I think there needs to be some facility for making certain parameters immutable.

-brian

graysky2:
Good point about implicate support.  I will remove the service from the
install and put it back to contrib.

For the 2nd point, I do not fully understand the system level overrides you
are describing in /etc/says config/vncserver since my distro does not ship
that file.  If a user on that system you describe were to enter alternative
options for things that are disallowed directly from a shell, what would
happen?

On Wednesday, October 14, 2015, Brian Hinz <notifications@github.com> wrote:

> I have a couple of issues with this.
>
> (1) I'm pretty strongly opposed to having the systemd service files be
> installed as a component of the software. Putting it into the contrib
> directory is one thing, but installing it alongside Xvnc pretty much makes
> it a first class citizen that we're then obligated to support. There are
> about a half dozen open bugs on red hat's bugzilla specifically related to
> TigerVNC + systemd, many of which have been open for many months (some more
> than a year). If RH/Fedora can't resolve these issues then I just don't see
> how we have the bandwidth or expertise to do so. And other distress are
> bound to have their own quirks too.
>
> (2) The changes to vncserver introduce the ability for users to
> arbitrarily override settings on non-systemd platforms. For example, on el6
> vncservers might be configured in /etc/sysconfig/vncserver with specific
> -SecurityType parameters as defined by corporate IT policy. It seems like
> those settings could now be superseded by anything set in ~/.vnc/config
> because it accepts any key=value pairs. I definitely like the idea of
> allowing users to manage things like geometry, etc. but I think there needs
> to be some facility for making certain parameters immutable.
>
> -brian
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/TigerVNC/tigervnc/pull/221#issuecomment-148258669>.
>


-- 
Sent from Gmail Mobile

bphinz:
On Thu, Oct 15, 2015 at 6:15 AM, graysky <notifications@github.com> wrote:

> For the 2nd point, I do not fully understand the system level overrides you
> are describing in /etc/says config/vncserver since my distro does not ship
> that file. If a user on that system you describe were to enter alternative
> options for things that are disallowed directly from a shell, what would
> happen?
>

Right, so my point is that the changes to the vncserver script would change
how it behaves on those other platforms.  Currently, sessions are
auto-started on many distros by assigning a particular port to a particular
user and specifying the arguments that are to be passed to vncserver in the
/etc/sysconfig/vncservers file.  I've never cared for this approach because
(1) it starts however many sessions are defined in the file, regardless of
whether or not they're ever used, and (2) for anything more than a handful
of users/machines it's a sysadmin nightmare because anytime someone wants
to change their screen size it has to be changed by the sysadmin.

Your work begins to address point #2, which I think is good, but I hate to
pull the rug out from under people that may have come to depend on the
legacy behavior.  I think it would be fine if the scope of the parameters
that a user is allowed to override were limited to either (a) ones that
have no potential security implications, or (b) those that are deemed
allowable by system configuration (meaning another system config file?
ick...).

graysky2:
OK.  I have organized the PR into more logical parts removing the installation of the service just placing it in contrib (again, good point).

To your point 2: I am still confused.  What would happen on a machine that has `/etc/sysconfig/vncservers` setup as you described if a user would open a shell and simply run the following:
```
/usr/bin/Xvnc -alwaysshared -geometry 2000x1200 -localhost -auth /home/facade/.Xauthority -fp /usr/share/fonts/misc,/usr/share/fonts/Type1 -pn -rfbauth /home/facade/.vnc/passwd -rfbport 5901 -rfbwait 30000
```

Would it refuse to execute if some of those options were blacklisted in `/etc/sysconfig/vncservers` ?

bphinz:
On Thu, Oct 15, 2015 at 2:58 PM, graysky <notifications@github.com> wrote:

> OK. I have organized the PR into more logical parts removing the
> installation of the service just placing it in contrib (again, good point).
>
> To your point 2: I am still confused. What would happen on a machine that
> has /etc/sysconfig/vncservers setup as you described if a user would open
> a shell and simply run the following:
>
> /usr/bin/Xvnc -alwaysshared -geometry 2000x1200 -localhost -auth /home/facade/.Xauthority -fp /usr/share/fonts/misc,/usr/share/fonts/Type1 -pn -rfbauth /home/facade/.vnc/passwd -rfbport 5901 -rfbwait 30000
>
> Would it refuse to execute if some of those options were blacklisted in
> /etc/sysconfig/vncservers ?
>
As long as 5901 wasn't already in use, a user could start a vnc session
with whatever parameters they want (unless otherwise locked down).  What
I'm talking about is users being able to override the parameters that would
have to have been setup by a sysadmin.  For security irrelevant parameters
like -geometry, that's fine (a plus, really) but not necessarily for things
like -SecurityTypes.

graysky2:
@bphinz - OK, I understand now.  So a user can run Xvnc with any options including -SecurityTypes just on a free port on this sort of system.

I think to implement what you're talking about:
1) A "banned" list of switches would need to be defined.
2) A check would need to be there such that if that file is not present, the "banned" option could be allowed.

Can you confirm this logic?  Also, can you suggest a list of "banned" switches besides -SecurityTypes?  Finally, I do not know how to write this logic in perl (it is NOT my mother tongue)... is there a project member with this skill set?

Thanks!

CendioOssman:
I'm not sure I support Brian's concerns here. Are the Xvnc processes running as the user, or as root? If they're running as the user then reading a user configuration file only exposes a security issue that already existed (as @graysky2 pointed out). So I do not think we need to concern ourselves with that problem in this PR.

bphinz:
On Fri, Oct 16, 2015 at 4:29 AM, Pierre Ossman (Work account) <

> I'm not sure I support Brian's concerns here. Are the Xvnc processes
> running as the user, or as root? If they're running as the user then
> reading a user configuration file only exposes a security issue that
> already existed (as @graysky2 <https://github.com/graysky2> pointed out).
> So I do not think we need to concern ourselves with that problem in this PR.
>
If you're OK with the rest then go ahead and merge it.  The value that the
change adds outweighs my concern.

-brian

graysky2:
I guess the only other question is: do we want the vncserver perl script to create a commented ~/.vnc/config as it does for ~/.vnc/xstartup or ~/.vnc/passwd?  The advantage of that is (if properly commented), the file would be self exploratory.  We could include example usage based on the defaults to illustrate passing options that do and do not expect an argument.  Overkill?

Something like:

```
# enter server options to pass to vncserver upon invocation
#
#name=desktop
#geometry=2000x1200
#depth=24
#localhost
#alwaysshared
```

bphinz:
On Fri, Oct 16, 2015 at 3:24 PM, graysky <notifications@github.com> wrote:

> I guess the only other question is: do we want the vncserver perl script
> to create a commented ~/.vnc/config as it does for ~/.vnc/xstartup or
> ~/.vnc/passwd? The advantage of that is (if properly commented), the file
> would be
>
I think it should

-brian

graysky2:
Hows 8fe3b85?

graysky2:
@CendioOssman - Is it line 89 to which you refer?  I didn't actually use `vencrypt` but perhaps I am misunderstanding your comment. https://github.com/TigerVNC/tigervnc/commit/8fe3b853ec9345af9ac0cc3d07fada2f3e4d6e6e#diff-29085adc2224c2aa522d7e2fcf6e7c86R89

PR #223, opened 2015-10-17
Allow static linking against only the GCC libs

dcommander:
BUILD_STATIC was originally intended as a way of preventing dependencies
on libgcc and libstdc++ in the TigerVNC binaries, thus allowing the
binaries to be built on an older Linux distro and then run on any newer
Linux distro.  In practice, the X11 libraries and most system libraries
tend to be forward compatible, so really the only libraries that need to
be statically linked are libjpeg-turbo (easy, since libjpeg-turbo
supplies static libs in its packages) and GnuTLS (requires building
GnuTLS from source.)

Over the years, BUILD_STATIC has been changed such that it attempts to
link TigerVNC statically with everything, but this is a very difficult
(and usually unnecessary) build environment to set up.  This patch adds
a new CMake variable (BUILD_STATIC_GCC) that behaves like BUILD_STATIC
was originally intended to behave.  BUILD_STATIC now implies
BUILD_STATIC_GCC.

### DIFF
--- a/cmake/StaticBuild.cmake
+++ b/cmake/StaticBuild.cmake
@@ -10,8 +10,13 @@
 option(BUILD_STATIC
     "Link statically against most libraries, if possible" OFF)
 
+option(BUILD_STATIC_GCC
+    "Link statically against only libgcc and libstdc++" OFF)
+
 if(BUILD_STATIC)
   message(STATUS "Attempting to link static binaries...")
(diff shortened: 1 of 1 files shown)


CendioOssman:
Merged as 2a2e2c3611e660da3f30744cb720bf363a4d3a01.

PR #224, opened 2015-10-17
Allow BUILD_TIMESTAMP to be set statically

dcommander:
Otherwise, every time CMakeLists.txt changes, the entire source tree
must be rebuilt, which is a productivity killer for people who are
trying to work with this code.  I'm assuming that there is a good reason
why BUILD_TIMESTAMP is being set dynamically every time CMakeLists.txt
changes, so this patch retains that behavior as the default while
allowing developers to override it.


CendioOssman:
Merged as 872e5dd859074567c3befbda9225e08710d13669.

PR #231, opened 2015-10-30
Added option -xstartup.

Siot:
A different xstartup file, from anywhere, could be loaded for each server.


Siot:
Thank you Pierre.

PR #236, opened 2015-11-12
Fix inverted logic in ProcVncExtSetParam.

michalsrb:
Using vncconfig to set the desktop parameter doesn't propagate the change because the check for "desktop" is inverted.


CendioOssman:
Oops. My bad. That got broken in f8e3b34c.

PR #247, opened 2015-12-06
Update Turkish Translation

ercanal:


CendioOssman:
Thank you for your contribution. Unfortunately we cannot accept it this way. Our translations are handled by the translation project and all updates need to go via them. You can submit your updated file to their Turkish team here:

http://translationproject.org/team/tr.html

PR #256, opened 2015-12-31
Split out acceptPrimaryCB option in UNIX vncconfig

### LINKED ISSUE #240: add option to not overwrite primary selection
I connect to a Linux VNC server with a Windows viewer. When I copy something in Windows it gets sent to both the X primary selection buffer and the X clipboard on Linux. I don't want it to overwrite the primary selection buffer.

I also don't like that the Linux X primary selection gets sent to Windows as the clipboard.

I have tried all combinations in the VNC config on the server and the Preference in the viewer and can't get the selection to not be sent or modified when the clipboard is.

I am running the Windows 64-bit 1.4.3 viewer on Windows 7, because all the more recent versions crash. I have seen this problem on most versions of the server and can confirm it with the 1.6beta (1.5.90).

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/28662242-add-option-to-not-overwrite-primary-selection?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github).
</bountysource-plugin>

hackonteur:
Resolves #240. This is the simplest approach, with acceptPrimaryCB
having the same relationship to acceptClipboard as sendPrimaryCB has to
sendClipboard. This does not change vncviewer, which should probably
have something similar on UNIX platforms only.

### DIFF
--- a/unix/vncconfig/vncconfig.cxx
+++ b/unix/vncconfig/vncconfig.cxx
@@ -51,6 +51,8 @@ LogWriter vlog("vncconfig");
 StringParameter displayname("display", "The X display", "");
 BoolParameter noWindow("nowin", "Don't display a window", 0);
 BoolParameter iconic("iconic", "Start with window iconified", 0);
+BoolParameter acceptPrimary("AcceptPrimary", "Accept the PRIMARY as well "
+                            "as the CLIPBOARD selection", true);
 BoolParameter sendPrimary("SendPrimary", "Send the PRIMARY as well as the "
                           "CLIPBOARD selection", true);
 IntParameter pollTime("poll",
@@ -87,6 +89,8 @@ public:
   VncConfigWindow(Display* dpy)
     : TXWindow(dpy, 300, 100), cutText(0), cutTextLen(0),
       acceptClipboard(dpy, "Accept clipboard from viewers", this, false, this),
+      acceptPrimaryCB(dpy, "Accept primary selection from viewers",
+                      this, false, this),
       sendClipboard(dpy, "Send clipboard to viewers", this, false, this),
       sendPrimaryCB(dpy, "Send primary selection to viewers", this,false,this),
       pollTimer(this),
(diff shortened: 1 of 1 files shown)


CendioOssman:
Patch looks good, but I'm concerned about the naming and description. It could easily be misinterpreted as PRIMARY and CLIPBOARD being handled independently. Perhaps just swap out "accept" for "set"? What do you think?

hackonteur:
I intentionally mimicked the naming from the existing SendPrimary setting. I think that the wording of the two should be as similar as possible.
I do not object to changing the wording of both to be clearer, but I would propose that be a change after this one which initiates the new setting to match the existing ones.

My long-term goal would be to add a new command to the protocol for sending the primary as separate from the clipboard, such that a UNIX viewer connected to a UNIX server would copy the two things independently. There would be no dependency on the settings (as there currently is):
SendClipboard: Send clipboard to viewer
SendPrimary: Send primary selection to viewer
AcceptClipboard: Accept clipboard from server
AcceptPrimary: Accept primary selection from server

If either side were non-UNIX, it would be handled accordingly:
SendClipboard: Send clipboard to viewer as clipboard
SendPrimary: Send clipboard to viewer as primary selection
AcceptClipboard: Accept clipboard from server as clipboard
AcceptPrimary: Accept primary selection from server as clipboard

CendioOssman:
> I intentionally mimicked the naming from the existing SendPrimary setting. I think that the wording of the two should be as similar as possible.

I understand, but they are not quite the same so I'm not sure that's a good idea. I'd like to avoid people getting upset that "AcceptPrimary" still puts PRIMARY from the client into CLIPBOARD on the server (or vice versa). The name and description makes it sound like it is an independent thing.

> I do not object to changing the wording of both to be clearer, but I would propose that be a change after this one which initiates the new setting to match the existing ones.
> 

Not sure I understand why. If we're going to change it, why not do it before it is committed?

> My long-term goal would be to add a new command to the protocol for sending the primary as separate from the clipboard, such that a UNIX viewer connected to a UNIX server would copy the two things independently.

Sounds good, although the jury is still out if PRIMARY will survive the transition to Wayland. At that point it makes perfect sense to have the same naming. But even then there will be years of legacy servers and clients that need to be handled so the original issue will remain.

CendioOssman:
Merged with naming change. Thanks for your contribution.

PR #272, opened 2016-02-22
Build libXvnc as separate library.

michalsrb:
So it can be used by others, not only vncconfig.

### DIFF
--- a/unix/vncconfig/CMakeLists.txt
+++ b/unix/vncconfig/CMakeLists.txt
@@ -3,13 +3,25 @@ include_directories(${X11_INCLUDE_DIR})
 include_directories(${CMAKE_SOURCE_DIR}/common)
 include_directories(${CMAKE_SOURCE_DIR}/unix/tx)
 
+include(GNUInstallDirs)
+
+add_library(Xvnc SHARED
+  vncExt.c)
+
+set_target_properties(Xvnc
+  PROPERTIES
+    VERSION 1.0.0
+    SOVERSION 1
+)
+
 add_executable(vncconfig 
   buildtime.c
-  vncExt.c
   vncconfig.cxx
   QueryConnectDialog.cxx)
 
-target_link_libraries(vncconfig tx rfb network rdr ${X11_LIBRARIES})
+target_link_libraries(vncconfig tx rfb network rdr Xvnc ${X11_LIBRARIES})
 
 install(TARGETS vncconfig DESTINATION ${BIN_DIR})
+install(TARGETS Xvnc LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} RENAME libXvnc.so)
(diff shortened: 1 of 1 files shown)


CendioOssman:
The VNC extension is rather horrible, so I'm not really keen on officially supporting it through something like this. At least not without a better API/protocol review first.

Did you have something specific in mind here?

michalsrb:
I am working on a VNC proxy & session manager that would allow users to create, share and reconnect their sessions on single VNC port. It is just a thin proxy, so authentication of users, encoding of the framebuffer etc. is responsibility of the VNC server behind. It is currently meant to be used on top of Tigervnc.

Part of it is a graphical tool where the users can configure security and sharing and make the session accessible to other clients from inside the session. This can be done nicely with the libXvnc extension. Alternative would be to execute the vncconfig tool, but that is less nice solution.

The proxy is here: [michalsrb/vncmanager](https://github.com/michalsrb/vncmanager). The configuration tool and a greeter app are not yet public. It's still work in progress.

CendioOssman:
Have you considered putting it under the TigerVNC umbrella and having it in the same tree? Then the protocol would remain an internal affair.

hildred:
I had a backburner project to port vncconfig to xcb, Which has a reasonably stable api and uses xml protocol descriptions to build the libraries, but if the extension is not standard perhaps a standard should be discussed first. What issues with the current protocol exist  that we may want to resolve first?

michalsrb:
> Have you considered putting it under the TigerVNC umbrella and having it in the same tree? Then the protocol would remain an internal affair.

If there is interest, that would be great. I am going to introduce the project to the Tigervnc community once all parts are public and I have proper explanation how it works written. It should be ready in few days.

CendioOssman:
> I had a backburner project to port vncconfig to xcb, Which has a reasonably stable api and uses xml protocol descriptions to build the libraries, but if the extension is not standard perhaps a standard should be discussed first. What issues with the current protocol exist that we may want to resolve first?

First off, if it gets to be a more official protocol then we need to think about how it can be extended in a backwards and forwards compatible manner. Ideally I'd like it to be general enough that other VNC projects (or even other remote desktop servers) can use it. That way people could develop generic tools for it. E.g. Gnome could handle connection requests no matter what server is used.

Which brings me to the second point, having it as a X11 extension might not be the best transport. Only things linked in to the X server can use those. So e.g. x0vncserver or vino cannot use such a mechanism. Something based on e.g. properties on the root window might be better. d-bus might also be an option, but that doesn't play well with X11 forwarding.

The clipboard bit is a problem, but I'm planning on nuking that so it's not of much concern. Some stronger type checks for the parameter handling would also be nice.

CendioOssman:
I'll close this PR as we're not planning on going this route (right now at least). @michalsrb, please open a new PR or send something on tigervnc-devel when you have more to show or discuss with regard to the "manager".

PR #279, opened 2016-03-18
Various fixes for Java viewer UI

bphinz:
Improves layout manager specifications rather than relying on any specific hard-coded values to achieve proper size and layout of UI.  Also adds command line arguments for specifying X.509 CA and CRL files.  Fixes a problem on OS X where the labels menu key field of the options dialog was displaying glyphs rather than text.


CendioOssman:
Please try to avoid merging `master` back to your topic branches. It just makes things confusing, as you can see from the supposed list of commits on this PR.

PR #280, opened 2016-03-18
change the Copyright year

BojanCendio:


CendioOssman:
If we updated the Java README, then shouldn't the top README also be updated? There's also a copyright year in the FLTK viewer somewhere.

PR #282, opened 2016-03-21
Add AllowOverride parameter.

michalsrb:
Allows to specify which configuration parameters can be modified on runtime. `ProcVncExtSetParam` now calls `vncOverrideParam` from `vncExtInit.cc` instead of directly calling `vncSetParamSimple`. `vncOverrideParam` verifies the parameter name against list of allowed parameters.

This is alternative for pull request #195.


CendioOssman:
Looks good. Could you update the man page as well? Might be good to note how `AllowOverride` and `NoClipboard` interact in there as well.

CendioOssman:
Sorry for the delay. It seems like github doesn't notify you of added commits to a PR. :/

PR #288, opened 2016-04-01
Update ru.po

Pro-pra:
Full rework russian translation.


CendioOssman:
Thank you for your contribution. Unfortunately our translations aren't handled by us directly, but rather by the translation project. So please submit your additions here instead:

http://translationproject.org/team/ru.html

PR #290, opened 2016-04-07
Add ubuntu xenial debian package in contrib.

not-a-user:
Ubuntu 16.04 LTS (xenial) is approaching, so there should be a source package to support tigervnc on the new release.

The package is simplified compared to trusty, because xenial brings both xorg server sources and fltk 1.3.3. However the former is version 1.18 and not yet supported by tigervnc, which is still worked around for the time being.


CendioOssman:
Thanks. Git has copy tracking in many of its tools, so two commits should not be necessary. It's also generally bad form to commit something that doesn't work, even if you clean it up in a following commit.

And I'd rather see we do this properly by adding a patch for 1.18. Have you tried applying the 1.17 patch and fixing the conflicts? You might also want to check Fedora's packaging. They usually fix up the patches without telling us. :/

not-a-user:
Added support for 1.18. Updated xenial package to finally use xorg-server-source. Added build-from-git.sh, which generates an up to date package (which is now easy, given that we can use git tar to generate the upstream tarball).

bphinz:
Looks great except that if the patch to unix/xserver/hw/vnc/Input.c is specifically for Xorg 1.18 then it should be merged into the xserver118.patch.  If it addresses some other issue and truly applies to all versions of Xorg, then I think it should be split off onto it's own commit.  I will start standing up a pair of xenial build hosts for the nightly builds this weekend.

CendioOssman:
> Looks great except that if the patch to unix/xserver/hw/vnc/Input.c is specifically for Xorg 1.18 then it should be merged into the xserver118.patch.

More likely that it needs some new `#ifdef`, like all other code there handling multiple Xorg versions.

contrib/packages/deb/ubuntu-xenial/debian/rules
```diff
@@ -180,6 +166,9 @@ install: build
 	# tigervncserver
 	make install DESTDIR=$(CURDIR)/debian/tigervncserver
 	(cd unix/xserver/hw/vnc; make install DESTDIR=$(CURDIR)/debian/tigervncserver)
+	# TODO: why does viewer stuff end up in server installation?
```
CendioOssman: Because there is no such thing as a server installation. It looks like the rules file need some general fixing. The proper approach is to do what the RPM build does, i.e. run `make install` in the top directory and separate things after that. Anything else is likely to break.

This should probably be fixed in a separate commit for 12.04 and 14.04 before the 16.04 packaging.

CendioOssman:
build-from-git.sh is a separate thing that should at the very least have its own commit (maybe its own PR. I haven't looked at it closely).

CendioOssman:
(And of course most of these commits should be merged in to one. Just to a --force push to this branch and this PR will update itself.)

not-a-user:
I see your points. See new PR.

PR #303, opened 2016-05-06
Add Catalan translation

rbuj:


CendioOssman:
Thank you for your contribution. Unfortunately all our translations are handled by the translation project. You can submit your translation to the Catalan team here and it will get to us the proper way:

http://translationproject.org/team/ca.html

rbuj:
Thanks for the info, but I don't have enough time for an application in translationproject.org, it's a pity.

CendioOssman:
I believe they will in most cases accept one-off submissions as well.

rbuj:
I just sent an email to Catalan team asking for application, and I completed also the FSF disclaimer. I'll send the translation to robot as soon as I'm accepted. Finally, I read the submission procedure and It doesn't seem very difficult. It seems to take a while for being accepted, despite this I think that you were right and I have to send it to TP.

PR #316, opened 2016-05-24
Documents $HOME/.vnc/default.tigervnc format

### LINKED ISSUE #270: Documentation for ~/.vnc/default.tigervnc
Is there documentation somewhere about `~/.vnc/default.tigervnc`?

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30829674-documentation-for-vnc-default-tigervnc?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jblaine:
Fixes #270

### DIFF
--- a/vncviewer/vncviewer.man
+++ b/vncviewer/vncviewer.man
@@ -280,7 +280,10 @@ respectively.
 .SH FILES
 .TP
 $HOME/.vnc/default.tigervnc
-Default configuration options.
+Default configuration options. This file must have a "magic" first line of
+"TigerVNC Configuration file Version 1.0" (without quotes), followed by simple
+<setting>=<value> pairs of your choosing. The available settings are those
+shown in this man page.
 .TP
 $HOME/.vnc/x509_ca.pem
 Default CA certificate for authenticating servers.


CendioOssman:
I'm a bit cautious about promising too much about this file as I do consider it somewhat internal. The fact that is has the same format as the "regular" .tigervnc files is because it was easy to implement it that way, rather than some stable design.

But we lack documentation for the format when used in other places, so I guess this is an improvement over the current state of things.

Corax26:
@CendioOssman why do you consider it internal? Is there any other way to configure vncviewer ? Anyhow thanks @jblaine for documenting it!

CendioOssman:
That file is the backend storage for the GUI. I.e. the GUI is the canonical interface for it. Hence it might also be different when appropriate (e.g.  on Windows we store stuff in the registry rather than to such a file) and extended or changed.

This however leaves gaps in automating things. Right now I'd say using the command line arguments, or separate .tigervnc files is the way to go. A global configuration would perhaps also be desirable, which would most likely have to be officially documented and may influence the stance on the local configuration storage.

Corax26:
Not sure I understand, is there a way to configure things persistently through the GUI? And how would you use a config file other than the default one for vncviewer?

CendioOssman:
> Not sure I understand, is there a way to configure things persistently through the GUI?

Any changes you do in the options dialog are saved to default.tigervnc.

> And how would you use a config file other than the default one for vncviewer?

By using the "Load" and "Save as" buttons in the viewer.

Corax26:
Oh sweet, thank you!

PR #322, opened 2016-06-11
Check CMake version for MinGW RC file handling

klusark:
CMake now does work with MinGW to build RC files

I've only tested this with CMake 3.5. For CMake 3.4 an up <INCLUDES> needs to be added to the RC compile line, however that would break CMake before 3.4. This fixes that as it looks to me like MinGW RC files with CMake was fixed in CMake 2.8.7

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -104,8 +104,9 @@ else()
   message(STATUS "32-bit build")
 endif()
 
-# CMake doesn't properly support resource compilation with MinGW.  Boo!
-if(MINGW)
+# Versions of CMake before 2.8.7 do not properly support resource compilation
+# with MinGW.  Boo!
+if(MINGW AND "${CMAKE_VERSION}" VERSION_LESS "2.8.7")
   if(NOT DEFINED RC)
     set(CMAKE_RC_COMPILER_INIT windres)
   else()


CendioOssman:
Great. The patch would need to update the minimum required version of CMake as well though.

bphinz:
@klusark - I just happened to try building with CMake 3.5.2 the other day and found that it failed because windres doesn't seem to be able to find certain header files from #include statements.  I don't believe that it's anything specific to my environment because the only thing that changed was upgrading cmake from 3.0.2 to 3.5.2.  Did you observe anything similar?

klusark:
@bphinz that's the exact issue I'm solving here. In more recent versions of CMake they changed how include flags worked and the way we call winres doesn't account for that. The only way I could think to fix it is to do this.

bphinz:
Great, then I can stop debugging that!  Rather than just removing that whole section, how about wrapping it in a conditional so that the older versions can still be used?  ie:

if("${CMAKE_VERSION}" VERSION_LESS "3.5.0")
  if(MINGW)
  ...
  endif()
endif()

Where 3.5.0 is replaced by the appropriate version.

klusark:
@bphinz I updated the commit. If you still have an environment with cmake 3.0.2, could you try it there?

CendioOssman:
Seems to work fine with cmake 2.8.12 at least.

klusark:
Is there anything blocking this from the repository? I'd really like to have this in so build on modern mingw work.

bphinz:
Sorry.  I will try to get this verified and committed tonight.

PR #341, opened 2016-08-06
Support /etc/tigervnc/vncserver-config. Overrides $HOME/.vnc/config

### LINKED ISSUE #317: Support system-wide UNIX/Linux Xvnc forced options
I'm proposing pretty simply edits to the `vncserver` perl script which will allow administrators a sane way to configure system-wide Xvnc options which override user defaults found in `$HOME/.vnc/config`

After the block https://github.com/TigerVNC/tigervnc/blob/master/unix/vncserver#L276-290 , system-wide options will be read from < taking-suggestions-for-filename-and-path > which then override only those options that were already set in `$config`

If the system-wide options file (yet to be named) does not exist, `vncserver` will function exactly as it always has.

Likewise, I will update the `vncserver` man page.

This will provide a way to at least **attempt** to enforce some default policy, such as TLS, through the most commonly used means of starting UNIX/Linux Xvnc sessions. It is fully understood that this system-wide preferred policy could be subverted by someone constructing their own `vncserver` script, running `Xvnc` by hand, etc.

Before I start on the code, are there objections? I intend to start work on the code + doc June 2.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/34512726-support-system-wide-unix-linux-xvnc-forced-options?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jblaine:
Fixes #317 

This commit introduces support for optional basic system-wide
configuration of vncserver.

The format and syntax of /etc/tigervnc/vncserver-config match
that of $HOME/.vnc/config exactly.

Any setting in /etc/tigervnc/vncserver-config (optional, used
if found) will override the same setting if found in
$HOME/.vnc/config.

For example, if you want to make a best effort to force everyone
to use TLS and some form of authentication, you could define
/etc/tigervnc/vncserver-config as *something like*:

securitytypes=tlsplain,tlsvnc,X509Vnc,X509Plain
plainusers='*'

### DIFF
--- a/unix/vncserver
+++ b/unix/vncserver
@@ -26,10 +26,7 @@
 # vncserver - wrapper script to start an X VNC server.
 #
 
-#
 # First make sure we're operating in a sane environment.
-#
-
 $exedir = "";
 $slashndx = rindex($0, "/");
 if($slashndx>=0) {
@@ -41,7 +38,8 @@ $vncClasses = "";
 &SanityCheck();
 
 #
-# Global variables.  You may want to configure some of these for your site.
+# Global variables.  You may want to configure some of these for
+# your site
 #
 
 $geometry = "1024x768";
@@ -51,10 +49,16 @@ $vncJavaFiles = (((-d "$vncClasses") && "$vncClasses") ||
                  ((-d "/usr/local/vnc/classes") && "/usr/local/vnc/classes"));
                  
 $vncUserDir = "$ENV{HOME}/.vnc";
+$vncUserConfig = "$vncUserDir/config";
+
+$vncSystemConfigDir = "/etc/tigervnc";
+$vncSystemConfigDefaultsFile = "$vncSystemConfigDir/vncserver-config-defaults";
+$vncSystemConfigMandatoryFile = "$vncSystemConfigDir/vncserver-config-mandatory";
+
 $skipxstartup = 0;
 $xauthorityFile = "$ENV{XAUTHORITY}" || "$ENV{HOME}/.Xauthority";
 
-$xstartup = $vncUserDir . "/xstartup";
+$xstartupFile = $vncUserDir . "/xstartup";
 $defaultXStartup
     = ("#!/bin/sh\n\n".
        "unset SESSION_MANAGER\n".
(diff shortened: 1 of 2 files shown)


CendioOssman:
Looks good. Nice work. Just two notes really:

1. It would be nice if the naming was clear that these settings are enforced, rather than default. Normally user settings are given higher priority than things in `/etc`. E.g. look at gconf which has `gconf.xml.defaults` and `gconf.xml.mandatory`. Given your `LoadConfig` I guess it would also be trivial to support both default and mandatory options?

2. TigerVNC might be installed with a different sysconf path, so we may need to have `vncserver` be generated at build time. This is something we can leave for later if you want though.

Oh, and merge the commits. No point in having needless noise like that in the history.

jblaine:
Yes, item 1 is easy. I'm passing on item 2.

How do you feel about warning the user for each setting from $HOME/.vnc/config that was overridden by a found /etc/tigervnc/vncserver-config-mandatory setting?

CendioOssman:
No strong feelings either way, so do what you feel works best.

jblaine:
Settting of $displayNumer, $desktopName, %default_opts, and loading all 3 config files had to be moved to above the code related to making $HOME/.vnc/passwd. This way no passwd file gets made if %config ends up having securitytypes defined with no VNC-auth types.

This fixes a bug that has existed for quite some time where $HOME/.vnc/config could say 'securitytypes=tlsplain' (or similar non-VNC-auth setting), yet running vncserver would still ask the user to define a password in $HOME/.vnc/passwd. This was confusing to new users who were handed a $HOME/.vnc/config so that their PAM/enterprise password could be used, but then they'd get asked to make a VNC password file.

jblaine:
See new single commit above which covers everything.

CendioOssman:
Some of the cleanups could probably have been separate commits to make the diff more readable. But other than that it looks great. Nice work. :)

PR #366, opened 2016-10-04
Do not die on no mcookie

metalefty:
for operating systems other than GNU/Linux.

mcookie is a part of util-linux. Usually only GNU/Linux systems have it.
Do not die even if mcookie is not found. Try to generate cookie from
/dev/random.

I thought generating cookie with pure perl. For example, using
Crypt::Random::Source but it requires additional dependency on another
perl module.  In order to keep it simple, I chose shell one-liner.

### DIFF
--- a/unix/vncserver
+++ b/unix/vncserver
@@ -293,8 +293,24 @@ $desktopLog = "$vncUserDir/$host:$displayNumber.log";
 unlink($desktopLog);
 
 # Make an X server cookie and set up the Xauthority file
-
+# mcookie is a part of util-linux, usually only GNU/Linux systems have it.
 $cookie = `mcookie`;
+# Fallback for non GNU/Linux OS - use /dev/urandom on systems that have it,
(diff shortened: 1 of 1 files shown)


CendioOssman:
The `mcookie` dependency was added in 5fe6070187d9f5ad212b9118502d972cba211dfa. It's probably best to use the old code in that commit as the fallback rather than something new. That code is pure perl as well.

metalefty:
I think simply using the old code is the best.  Need not use `mcookie`. Pure perl code is portable.

CendioOssman:
The distributions prefer `mcookie` so that commit was made to reduce their patching.

metalefty:
OK, use `mcookie` first and use the old pure perl code as fallback as you said.  Non GNU/Linux distributions should be cared.  The commit increases patching rather than reduce for non GNU/Linux operating systems.

metalefty:
I've updated the pull request.

CendioOssman:
Great. Thanks. :)

PR #371, opened 2016-10-25
MacOS package: Added NSHighResolutionCapable tag to Info.plist

fredowski:
With the flag the application will start enable the high resolution
display. I tested this on a Macbook Pro with a High Resolution Display.
The fonts are better to read.

Here you can see the difference in font rendering on my Macbook:

highres-display


CendioOssman:
Nice. But how does this affect the session? Is everything scaled properly?

fredowski:
Hi Pierre,

the session is identical. As far as I see it it only affects the gui elements like fonts.

Friedrich

PR #382, opened 2016-11-16
Separate 'Grab Keys' from 'Fullscreen'

jennamagius:
Currently, vncviewer will intercept special key combinations such as "Alt+Tab" and send them to the remote host if and only if vncviewer is in "Fullscreen" mode. This patch adds an option to the F8 menu that allows users to select "Grab Keys" mode separately from "Fullscreen" mode.

### DIFF
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -472,11 +472,10 @@ void DesktopWindow::fullscreen_on()
 }
 
 void DesktopWindow::toggleGrabKeys() {
-  this->grabActive = !(this->grabActive);
   if(this->grabActive)
-    grabKeyboard();
-  else
     ungrabKeyboard();
+  else
+    grabKeyboard();
 }
 
 void DesktopWindow::grabKeyboard()
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -59,7 +59,7 @@ DesktopWindow::DesktopWindow(int w, int h, const char *name,
                              const rfb::PixelFormat& serverPF,
                              CConn* cc_)
   : Fl_Window(w, h), cc(cc_), firstUpdate(true),
-    delayedFullscreen(false), delayedDesktopSize(false)
+    delayedFullscreen(false), delayedDesktopSize(false), grab_keys(false)
 {
   scroll = new Fl_Scroll(0, 0, w, h);
   scroll->color(FL_BLACK);
@@ -387,6 +387,10 @@ int DesktopWindow::handle(int event)
 }
 
 
+bool DesktopWindow::grab_active() {
+  return grab_keys;
+}
+
 int DesktopWindow::fltkHandle(int event, Fl_Window *win)
 {
   int ret;
@@ -469,8 +473,14 @@ void DesktopWindow::fullscreen_on()
   fullscreen();
 }
 
+void DesktopWindow::toggle_grab_keys() {
+  this->grab_keys = !(this->grab_keys);
+  grabKeyboard();
+}
+
 void DesktopWindow::grabKeyboard()
 {
+  if(!(this->grab_keys)) { ungrabKeyboard(); return; }
   // Grabbing the keyboard is fairly safe as FLTK reroutes events to the
   // correct widget regardless of which low level window got the system
   // event.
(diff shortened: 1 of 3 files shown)


vncviewer/DesktopWindow.cxx
```diff
@@ -469,8 +473,14 @@ void DesktopWindow::fullscreen_on()
...
 }
 
+void DesktopWindow::toggle_grab_keys() {
+  this->grab_keys = !(this->grab_keys);
+  grabKeyboard();
+}
+
 void DesktopWindow::grabKeyboard()
 {
+  if(!(this->grab_keys)) { ungrabKeyboard(); return; }
```
CendioOssman: Put this in `toggle_grab_keys()` instead so that the method names make sense.

vncviewer/DesktopWindow.cxx
```diff
@@ -469,8 +473,14 @@ void DesktopWindow::fullscreen_on()
   fullscreen();
 }
 
+void DesktopWindow::toggle_grab_keys() {
+  this->grab_keys = !(this->grab_keys);
```
CendioOssman: This should be updated when the state changes, even if calling `grabKeyboard()` directly.

vncviewer/DesktopWindow.cxx
```diff
@@ -59,7 +59,7 @@ DesktopWindow::DesktopWindow(int w, int h, const char *name,
                              const rfb::PixelFormat& serverPF,
                              CConn* cc_)
   : Fl_Window(w, h), cc(cc_), firstUpdate(true),
-    delayedFullscreen(false), delayedDesktopSize(false)
+    delayedFullscreen(false), delayedDesktopSize(false), grab_keys(false)
```
CendioOssman: Follow the naming conventions please. This should be `grabActive` or something like that and the functions should be `toggleGrabKeyboard()` and `isGrabActive()`.

jennamagius:
Thanks for reviewing. I have updated accordingly.

CendioOssman:
I just need to test this on the different platforms, but otherwise it looks good to go.

CendioOssman:
Hmm... This doesn't interact well with the existing behaviour. Entering full screen no longer does an automatic grab.

We may need to rethink the settings a bit here, given the interactions of these. Ideas?

jennamagius:
The reason I make this change in my personal client is: I am frequently working on three or more hosts that all have the same resolution (1920x1080). Fullscreen mode is very useful in that case because it lets you access the entire screen of remote hosts without using scrollbars (scrollbars are bad). But I also want to be switching between multiple remote hosts rapidly.

In my case, I even have a different "switch window" gesture on my local host than I do on the remote hosts, so I can use Super+Up & Super+Down to switch between client instances and Alt+Tab still gets passed through the guest because my window manager isn't eating it. But: when FLTK is grabKeyboard'd, I still have to un-full-screen to switch clients & re-full-screen every time I return to a client.

There's a couple of ways to support this use case:

1. The obvious way to keep this same patch is to make Grab Keys a tristate that defaults to "Follow fullscreen". It's easy & compatible, but not terribly elegant.

2. Another patch that I have made to my own client has been to add an option to the F8 menu for "temporary ungrab". Basically: the vncviewer window ungrabs stays ungrabbed until it gets a Focus-In event. This makes the "switch client" gesture "F8 u Super+Up" in my case. That's kinda OK, but it means I can't just thumb through my windows anymore, 'cause every time I hit a vncviewer window I get hung up there until I ungrab again. This also works pretty similarly to just using the existing minimize button, but the problem with the minimize button is: it moves vncviewer to the bottom of my window list. When I want to focus away, I just want it to go to second-place in my window list, not the bottom.

3. There could maybe be a config option where specific key-combos get passed through somehow? If I could just tell vncviewer to never eat Super+Up and Super+Down, that'd serve my needs.

4. We don't actually merge anything into core at all, and I just keep patching mine locally. I have no problem with that, I just wanted to offer this patch up 'cause if I'm having to make this change someone else probably is too.

CendioOssman:
Perhaps something like VMware where the grab is released by pressing a few modifiers?

jennamagius:
Sure, that would do the trick. Ideally it would be a configurable combo, because Ctrl+Alt to ungrab from VMWare is already something I use through vncviewer. Functionally, that's pretty similar to number 2 in my list, just in my case the gesture is "F8 u". I'll look at putting together a good version of that patch.

PR #397, opened 2017-01-10
Updated Polish translation

m4sk1n:


CendioOssman:
Thank you for your contribution. Unfortunately our translations are handled by [the translation project](http://translationproject.org). Please send you updated .po file to them and it will get included in TigerVNC.

PR #399, opened 2017-01-17
Fix buffer overflow in ModifiablePixelBuffer::fillRect.

michalsrb:
LibVNC had security bug and the reproducer can trigger similar kind of security bug in vncviewer as well:
https://github.com/LibVNC/libvncserver/pull/137
(Using the `send_copyrect_crash` function, which has bit misleading name because the issue is in RRE and filling rectangles, not copying.)

Malicious VNC server can send RRE message with subrectangle which is out of the framebuffer rectangle. Vncviewer then fills this rectangle and writes into random memory.

I wasn't sure at what level should the test whether the subrectangle is inside framebuffer be. I've added it to `PixelBuffer::fillRect` inspired by the checks inside `PixelBuffer::copyRect`.

### DIFF
--- a/common/rfb/PixelBuffer.cxx
+++ b/common/rfb/PixelBuffer.cxx
@@ -101,15 +101,26 @@ void ModifiablePixelBuffer::fillRect(const Rect& r, const void* pix)
   int stride;
   U8 *buf;
   int w, h, b;
+  Rect drect;
 
-  w = r.width();
-  h = r.height();
+  drect = r;
+  if (!drect.enclosed_by(getRect())) {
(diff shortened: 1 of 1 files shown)


CendioOssman:
A good start. We should probably add more checks at other layers as well though. If nothing else to get better error messages.

And I don't see why we don't just completely fail things at this point, as these are invalid requests. I'll do some follow up commits.

Thanks for the report and patch.

carnil:
This has been assigned CVE-2017-5581[](http://www.openwall.com/lists/oss-security/2017/01/25/6)

PR #402, opened 2017-01-19
Workaround for shared memory leakage.

ken3:
Description of problem:
   vncviewer creates a shared memory segment for MIT-SHM. It is not removed when vncviewer is killed
   by signal, because PlatformPixelBuffer destructor is never called after CleanupSignalHandler().
Workaround:
   Mark the shared memory as "destroyed" as soon as possible.
   It will be invisible soon after the vncviewer is killed.

See https://bugzilla.redhat.com/show_bug.cgi?id=1358090

### DIFF
--- a/vncviewer/X11PixelBuffer.cxx
+++ b/vncviewer/X11PixelBuffer.cxx
@@ -187,8 +187,9 @@ int X11PixelBuffer::setupShm()
     goto free_xim;
 
   shminfo->shmaddr = xim->data = (char*)shmat(shminfo->shmid, 0, 0);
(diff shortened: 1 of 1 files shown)


CendioOssman:
Great. This doesn't really sound like a workaround though, but rather a proper fix. Could you get rid of the old code and goto target instead of just commenting it out?

ken3:
Fixed. How do you like it?

CendioOssman:
Looks good. Could you squash them in to a single commit?

ken3:
Done.

PR #426, opened 2017-03-03
Fullscreen on selected monitors

ercanal:
Hi 

I added fullscreen on selected monitors option to vncviewer.


CendioOssman:
Nice. Could you clean up the branch though so it's just the relevant change?

PR #427, opened 2017-03-03
Fullscreen on selected monitors

ercanal:


bphinz:
Is this limited to just two monitors?

ercanal:
Yes

Also selected monitors must create a rectangular area otherwise due to current fltk fullscreen implementation may overlap with other screens.

________________________________
Kimden: Brian Hinz [notifications@github.com]
Gönderildi: 03 Mart 2017 Cuma 19:29
Kime: TigerVNC/tigervnc
Bilgi: Ercan Altuntas; Author
Konu: Re: [TigerVNC/tigervnc] Fullscreen on selected monitors (#427)

Is this limited to just two monitors?

CendioOssman:
Hmm... That's a shame. How do other systems handle this? Do they even support this scenario?

ercanal:
What are referring with other systems ?? Other VNC software or other GUI toolkits??

RealVNC 5 doesn't have selectable monitors for fullscreen. 

If polygon border support is added to FLTK, non rectangular screen selections can be supported for monitors have touching edges or corner. This should be true for almost all GUI toolkits.

CendioOssman:
> What are referring with other systems ?? Other VNC software or other GUI toolkits??
> 
> RealVNC 5 doesn't have selectable monitors for fullscreen.
> 

Other applications that work in similar ways. Primarily other remote access clients, not just VNC.

> If polygon border support is added to FLTK, non rectangular screen selections can be supported for monitors have touching edges or corner. This should be true for almost all GUI toolkits.

Right. I was a bit unclear. It was the fact that we limit it to an arbitrary number, not that we will require a square area.

CendioOssman:
Closing for lack of activity.

ic3dr460n:
Does this actually work? Sorry to revive an old topic but I was looking for this feature and can't find it. I have 3 monitors and I only want 2 with my TigerVNC.

samhed:
This is available in 1.12.0 as of #1282

PR #430, opened 2017-03-09
minor changes to 'tigervncserver' as provided by the debian package.

abelew:
I edited vncserver in order to make it work with use strict and with Getopt::Long.  Upon completion I decided to compare its output
against that of the vncserver on my debian host. I realized immediately
that it starts with a different script 'tigervncserver', and that the
author already did 99% of what I wanted.  I therefore added the few
lines of extra help information which has helped folks here at school
when connecting and added it to this repository.  I assume but do not
know for certain that this is acceptable.

### DIFF
--- a/unix/tigervncserver
+++ b/unix/tigervncserver
@@ -1,0 +1,1482 @@
+#! /usr/bin/env perl
+# vim: set sw=2 sts=2 ts=8 syn=perl expandtab:
+#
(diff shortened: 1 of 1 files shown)


CendioOssman:
This is an entirely new file rather than a change to `vncserver`. Please change it so that it modifies the existing script.

CendioOssman:
No response. Closing.

PR #438, opened 2017-03-27
Prevent double free by crafted fences.

### LINKED ISSUE #437: Double free in server caused by fences
If client sends fence with some data, followed by a fence with no data (length 0), the `VNCSConnectionST::fence` will delete `VNCSConnectionST::fenceData`, but keep it pointing at the same memory:

      delete [] fenceData;
      if (len > 0) {
        fenceData = new char[len];
        memcpy(fenceData, data, len);
      }

      return;

If the client sends another fence or if it disconnects the `VNCSConnectionST::fenceData` is deleted for second time. It can crash the whole server or possibly cause some security issues.

It is not clear to me from the RFB protocol documentation if fences with zero length are ok. If not, then `SMsgReader` should reject it. If they are ok, then `VNCSConnectionST` needs to set the `fenceData` to null and `SMsgWriter::writeFence` must be able to write that kind of fence back to client.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43489777-double-free-in-server-caused-by-fences?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github).
</bountysource-plugin>

michalsrb:
If client sent fence with some data, followed by fence with no data (length 0), the original fence data were freed, but the pointer kept pointing at them. Sending one more fence would attempt to free them again.

fixes #437

### DIFF
--- a/common/rfb/SMsgWriter.cxx
+++ b/common/rfb/SMsgWriter.cxx
@@ -101,7 +101,9 @@ void SMsgWriter::writeFence(rdr::U32 flags, unsigned len, const char data[])
   os->writeU32(flags);
 
   os->writeU8(len);
-  os->writeBytes(data, len);
+
+  if (len > 0)
+    os->writeBytes(data, len);
 
   endMsg();
 }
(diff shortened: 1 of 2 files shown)


carnil:
This has been assigned CVE-2017-7393

PR #440, opened 2017-03-29
Fixes ssecurityplain

michalsrb:
Two security fixes for SSecurityPlain.

### DIFF
--- a/common/rfb/SSecurityPlain.cxx
+++ b/common/rfb/SSecurityPlain.cxx
@@ -86,8 +86,15 @@ bool SSecurityPlain::processMsg(SConnection* sc)
   if (state == 0) {
     if (!is->checkNoWait(8))
       return false;
+
     ulen = is->readU32();
+    if (ulen > MaxSaneUsernameLength)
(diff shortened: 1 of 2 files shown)


CendioOssman:
I don't see where those two extra bytes are read. I see one `readBytes()` for `ulen`, and one for `plen`. So the existing code looks correct?

michalsrb:
The original code was:

    if (is->checkNoWait(ulen + plen + 2))

It was missing the negation, so I corrected it to:

    if (!is->checkNoWait(ulen + plen + 2))

And I created the pull request. But then I noticed that the formula is also incorrect so I removed the + 2:

    if (!is->checkNoWait(ulen + plen))

Now I believe it is correct. I have modified the commit while it was waiting for pull, which is probably what caused the confusion, sorry about that.

CendioOssman:
Ah. I managed to look at the diff backwards somehow. I'll blame it on it being late here. ;)

carnil:
This has been assigned CVE-2017-7394

PR #441, opened 2017-03-29
Delete underlying ssecurity in SSecurityVeNCrypt.

michalsrb:
Otherwise it gets leaked which would allow even not authenticated clients to exhaust server memory.


carnil:
This has been assigned CVE-2017-7392

PR #442, opened 2017-03-30
Improve congestion control

CendioOssman:
This patch series makes the congestion control a bit more responsive and it should be able to make better decisions.

If you use TigerVNC over a high latency connection then please test and report back. Note that these changes are in the server, not the client.


CendioOssman:
Did a whole bunch of test runs over various network conditions (using simulator):

[compare.xlsx](https://github.com/TigerVNC/tigervnc/files/1759279/compare.xlsx)

All cases showed improvement or similar behaviour to the old system (generally latency > 100ms showed improvement).

PR #444, opened 2017-04-06
Limit size of cursor accepted by client.

michalsrb:
Width and height of a cursor are received as U16 from network. Accepting full range of U16 values can cause integer overflows in multiple places.

The worst is probably VLA in CMsgReader::readSetXCursor:

    rdr::U8 buf[width*height*4];

The `width*height*4` can be too big to fit on stack or it can overflow into negative numbers. Both cases are undefined behaviour. Following writes to buf can overwrite other data on stack.


CendioOssman:
Thanks. You doing an audit of the entire network code? :)

michalsrb:
Kinda, poking around. :-)

PR #449, opened 2017-04-28
macOS performance fix fixes

CendioOssman:
There was some fallout from the macOS performance improvement. Primarily the screen could stop updating if you moved the window to a secondary monitor.

This PR fixes that, along with some other tweaks.


CendioOssman:
@bphinz, this exposes a bug in FLTK 1.3.3. Do you have 1.3.4 in your build system?

PR #450, opened 2017-04-29
Mark strings for translation

alshopov:


alshopov:
Even though it is impossible to translate all of the strings in these dialogs as some of them come from FLTK - it is still better to have some of them localized.
Usually translations are not 100% for most of the languages most of the time so it is a normal and expected situation.

Issue #448, opened 2017-04-28
XServer crash with dummy driver and vnc module


grulja:
Using xorg 1.19 and dummy driver + vnc module makes Xserver to crash due to latest change in vncHooks.c. It's specifically this commit [1] as this issue is not reproducible with Tigervnc 1.7.1 or with 1.7.90 and reverted mentioned commit. I'm also attaching the backtrace, let me please now whether you need more information.

[1] - https://github.com/TigerVNC/tigervnc/commit/84c72feb3999339fc

[xorg_crash.txt](https://github.com/TigerVNC/tigervnc/files/963833/xorg_crash.txt)
6529b4b0abae9b50fee4245

grulja:
Ok, I did some investigation and I think I spotted potential problem:
In previous code I see there was call:
**rp->rrCrtcSet = vncHooksScreen->RandRCrtcSet;**
which you replaced by:
**unwrap(vncHooksScreen, rp, rr##field);**
but if I'm not mistaken this will produce:
**rp->rrCrtSet = vncHooksScreen->rrCrtSet**
which is different from previous call

grulja:
If that's true, then it applies also to the function above, but I was looking into the part mentioned in backtrace.

CendioOssman:
That member was renamed in order to make the macros possible. It would not have compiled otherwise.

The backtrace shows that it hit an assert. Do you have the logs to see which assert?

And why are you using the dummy driver instead of Xvnc?

grulja:
Aha, I see so I was wrong. The reason why we use the dummy driver is that some machines are without a graphic card, also the dummy driver works with GDM service.

Attached Xorg.log if you meant that.But I don't see anything relevant there. Also interesting is that this is reproducible only when you use "startx", running Xserver using "X", or "xinit" or "Xorg" seems to avoid the crash.
[xorg_log.txt](https://github.com/TigerVNC/tigervnc/files/964300/xorg_log.txt)

CendioOssman:
> Aha, I see so I was wrong. The reason why we use the dummy driver is that some machines are without a graphic card, also the dummy driver works with GDM service.

Neither of these should be a problem with Xvnc.

> Attached Xorg.log

This is not the same crash as the first report, so please try again so that we get something consistent. It also looks like you have some problem with missing debug symbols.

grulja:
[...]

PR #452, opened 2017-05-02
Make RandR callbacks optional as they were before

grulja:
Previously tigervnc had RandR callbacks optional, after it was rewritten in https://github.com/TigerVNC/tigervnc/commit/84c72feb3999339fc to use wrap/unwrap macros this has been removed, which causes a crash https://github.com/TigerVNC/tigervnc/issues/448. This commit makes them optional again and I cannot reproduce the mentioned crash.

### DIFF
--- a/unix/xserver/hw/vnc/vncHooks.c
+++ b/unix/xserver/hw/vnc/vncHooks.c
@@ -312,9 +312,21 @@ int vncHooksInit(int scrIdx)
 #ifdef RANDR
   rp = rrGetScrPriv(pScreen);
   if (rp) {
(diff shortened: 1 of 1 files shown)


CendioOssman:
Ah. Good catch. But I'd like the comment back as well then. And can we avoid the `else` clause to avoid clutter? Probably requires `if` clauses in the cleanup as well though. Thanks.

CendioOssman:
Oh right. Please also make the PR against `master`. We can cherry-pick it for the 1.8 release as a separate step.

PR #453, opened 2017-05-03
Make RandR calls optional as they were before

grulja:
Previously tigervnc had RandR callbacks optional, after it was rewritten in https://github.com/TigerVNC/tigervnc/commit/84c72feb3999339fc to use wrap/unwrap macros this has been removed, which causes a crash https://github.com/TigerVNC/tigervnc/issues/448. This commit makes them optional again and I cannot reproduce the mentioned crash.

Sorry for creating a new PR, didn't know how to change my previous one to master.

### DIFF
--- a/unix/xserver/hw/vnc/vncHooks.c
+++ b/unix/xserver/hw/vnc/vncHooks.c
@@ -312,9 +312,13 @@ int vncHooksInit(int scrIdx)
 #ifdef RANDR
   rp = rrGetScrPriv(pScreen);
   if (rp) {
-    wrap(vncHooksScreen, rp, rrSetConfig, vncHooksRandRSetConfig);
-    wrap(vncHooksScreen, rp, rrScreenSetSize, vncHooksRandRScreenSetSize);
-    wrap(vncHooksScreen, rp, rrCrtcSet, vncHooksRandRCrtcSet);
+    /* Some RandR callbacks are optional */
+    if (rp->rrSetConfig)
+      wrap(vncHooksScreen, rp, rrSetConfig, vncHooksRandRSetConfig);
+    if (rp->rrScreenSetSize)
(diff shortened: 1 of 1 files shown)


CendioOssman:
This looks odd. Why start with a bunch of `unwrap()`? Wouldn't something like this work:

```C
if (rp) {
  if (rp->rrSetConfig)
    wrap(vncHooksScreen, rp, rrSetConfig, vncHooksRandRSetConfig);
...
```

> Sorry for creating a new PR, didn't know how to change my previous one to master.

If you click `Edit` at the top you should get a drop down allowing you to change the target branch.

grulja:
Because you still want to set something like:
`vncHooksScreen->rrSetConfig = rp->rrSetConfig`
regardless rp->rrSetConfig is not available, but you want to have 
`rp->rrSetConfig = vncHooksRandRSetConfig;`
only when it is actually available, at least that is how it was done before.

grulja:
To be honest I'm not really sure how this works internally, I just spotted this difference in the mentioned commit and tried to change it back to avoid the crash. If setting
`vncHooksScreen->rrSetConfig = rp->rrSetConfig`
doesn't make sense without rp->rrSetConfig being available, then we do it the way you suggested.

CendioOssman:
Using `unwrap()` for that side effect isn't very clear in that case.

Is the `vncHooksScreen` structure cleared to zero when we get it? Seems so from a quick glance of the Xorg code. If so then `vncHooksScreen->rrSetConfig` is already set to NULL and we don't need to change it. We could also be paranoid and do a `memset()` to be sure.

CendioOssman:
Great. Please merge these in to one commit and it'll get everything sorted. You can do `git push --force` to update this PR without having to create a new one.

PR #464, opened 2017-05-23
vncviewer: Fix fullscreen scrolling

### LINKED ISSUE #463: Fullscreen scrolling no longer works
The vncviewer(1) manual says about fullscreen mode:

> If the remote screen is bigger, you can scroll by bumping the mouse against the edge of the screen.

This worked in 1.7.1, but no longer works after upgrading to 1.8.0 (and I have confirmed that downgrading to 1.7.1 "fixes" the issue).

This is on X11 with FLTK 1.3.4.

LukeShu:
I've gone ahead and fixed #463 .

It's a fairly simple fix.


CendioOssman:
Thanks. Please change the commit and PR to target `master` though.

Issue #465, opened 2017-05-23
Doesn't show scrollbar when other scrollbar reduces visible area.


LukeShu:
Have a window that is sized to the remote screen. Now shrink the window vertically, making it shorter than the remote screen in one axis. The vertical scrollbar appears.  However, the horizontal scrollbar does not appear, despite the rightmost ~24 pixels (`Fl::scrollbar_size()`) being hidden by the vertical scroll bar.

PR #466, opened 2017-05-23
vncviewer: Fix scrollbar visibility

LukeShu:
Have a window that is sized to the remote screen.  Now shrink the window
vertically, making it shorter than the remote screen in one axis. The
vertical scrollbar appears. However, the horizontal scrollbar does not
appear, despite the rightmost ~24 pixels (Fl::scrollbar_size()) being
hidden by the vertical scroll bar.

Fix that. (#465)

While the adjusted check is fairly simple, I think that it might not be
entirely obvious until you get it; so for posterity:

  The check for the X-axis is

      if ((w() - (h() < viewport->h() ? Fl::scrollbar_size() : 0) < viewport->w())

  But to reason about it a little more explicitly, let's split that
  ternary in to two separate checks, and add some comments:

      if (
             (w() -  < viewport->w()) // X needs a scrollbar
          ||
             ( (w() - Fl::scrollbar_size() < viewport->w()) && (h() < viewport->h()) )
             //( X doesn't need a scrollbar unless Y does ) && (  Y does need one  ) )
         )

  Within the "Y does need one" check, we don't need to worry about the
  case where `h() - Fl::scrollbar_size() < viewport-h()` is true,
  because if both sides are saying "I don't need a scrollbar unless
  you do", then neither needs one.

### DIFF
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -1046,15 +1046,35 @@ void DesktopWindow::repositionWidgets()
 
   // Scrollbars visbility
 
-  if (!fullscreen_active() && (w() < viewport->w()))
-    hscroll->show();
-  else
+  if (fullscreen_active()) {
     hscroll->hide();
-
-  if (!fullscreen_active() && (h() < viewport->h()))
-    vscroll->show();
-  else
     vscroll->hide();
+  } else {
+    // Decide whether to show a scrollbar by checking if the window
+    // size (possibly minus scrollbar_size) is less than the viewport
+    // (remote framebuffer) size.
+    //
+    // We decide whether to subtract scrollbar_size on an axis by
+    // checking if the other axis *definitely* needs a scrollbar.  You
+    // might be tempted to think that this becomes a weird recursive
+    // problem, but it isn't: If the window size is less than the
+    // viewport size (without subtracting the scrollbar_size), then
+    // that axis *definitely* needs a scrollbar; if the check changes
+    // when we subtract scrollbar_size, then that axis only *maybe*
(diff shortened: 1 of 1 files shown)


CendioOssman:
Great catch. And as you write, the logic isn't clear at first glance.

Perhaps we don't need to solve it in a single if statement? Might be clearer if we do a first pass just checking width. After that a second pass to check if the other also needs to be enabled. We'll duplicate the calls to `show()`, but hopefully the logic will be easier to follow?

CendioOssman:
Oh, and like the other PR please change this to target `master`.

LukeShu:
I had at one point tried writing it as

```
  if (fullscreen_active()) {
    hscroll->hide();
    vscroll->hide();
  } else {
    if (w() < viewport->w())
      hscroll->show();
    else if ((w() - Fl::scrollbar_size() < viewport->w()) && (h() < viewport->h()))
      hscroll->show();
    else
      hscroll->hide();

    if (h() < viewport->h())
      vscroll->show();
    else if ((h() - Fl::scrollbar_size() < viewport->h()) && (w() < viewport->w()))
      vscroll->show();
    else
      vscroll->hide();
  }
```

But I ended up deciding that that was less clear than what I ended up submitting.

I think the code itself is clear enough if you read it carefully "if the window width (possibly minus scrollbar_size) is less than the viewport width ...".  I think the trickier part is convincing the reader that the check to decide if to subtract the scrollbar_size is correct.  I guess I should add a comment then...

I've rebased onto `master` and updated the PR.

LukeShu:
I've just now updated the PR; I pulled the `fullscreen_active()` part of the check out, and added a comment. I think it's now sufficiently clear, but I wrote it, so of course I understand it.

PR #472, opened 2017-06-10
Override connection to an untrusted remote host

jankratochvil:
Sorry but there should be an option to override a certificate.  I do not have the public CA file or public key from a remote server.  I have only its fingerprint (but I do not need to even have that).  I know it is insecure but I just have no other option.
A missing feature is also to verify the remote server only by its fingerprint.
Being tested on: 15191.vm91.wedos.net:25191
I get there:
```
 TLS:         GNUTLS status of certificate verification: 322
 CConn:       Invalid status of server certificate verification
322 = 0x142
        GNUTLS_CERT_INVALID = 1 << 1,
        GNUTLS_CERT_SIGNER_NOT_FOUND = 1 << 6,
        GNUTLS_CERT_INSECURE_ALGORITHM = 1 << 8,
```

### DIFF
--- a/common/rfb/CSecurityTLS.cxx
+++ b/common/rfb/CSecurityTLS.cxx
@@ -309,13 +309,14 @@ void CSecurityTLS::checkSession()
   if (gnutls_x509_crt_import(crt, &cert_list[0], GNUTLS_X509_FMT_DER) < 0)
     throw AuthFailureException("decoding of certificate failed");
 
+char *const override(getenv("VNCVIEWER_JANKRATOCHVIL_OVERRIDE"));
   if (gnutls_x509_crt_check_hostname(crt, client->getServerName()) == 0) {
     char buf[255];
     vlog.debug("hostname mismatch");
     snprintf(buf, sizeof(buf), "Hostname (%s) does not match any certificate, "
 			       "do you want to continue?", client->getServerName());
     buf[sizeof(buf) - 1] = '\0';
-    if (!msg->showMsgBox(UserMsgBox::M_YESNO, "hostname mismatch", buf))
+    if (!override&&!msg->showMsgBox(UserMsgBox::M_YESNO, "hostname mismatch", buf))
       throw AuthFailureException("hostname mismatch");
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm sorry, but this does not look like something we are interested in. If you wish to use TLS without server authentication then enable the `TLS` auth types rather than the `X509` ones.

dahaic:
@CendioOssman As I got to the same situation, I want to point out that in this case, you can't just enable `TLS` auth types. Server configuration is given - they only accept `X509` auth, but their certificate is in a bad shape. So it would be useful to really have a way to tell "Hey, I know it's a mess, but please proceed".

bphinz:
> Sorry but there should be an option to override a certificate. I do not have the public CA file or public key from a remote server. I have only its fingerprint (but I do not need to even have that).

I don't think the error you are seeing is due to either of those reasons.  If you look at line 263 in common/rfb/CSecurityTLS.cxx you'll see that GNUTLS_CERT_INVALID and GNUTLS_CERT_SIGNER_NOT_FOUND are handled as allowed in certain situations.  The problem is most likely being caused by the GNUTLS_CERT_INSECURE_ALGORITHM, which means the server certificate is signed using MD2 or MD5.  A better way to handle this would be to check for this condition and raise a dialog similar to how expired certs are handled (see line 291).

jankratochvil:
I would like to note that after using my patched vncviewer and clicking there to save some insecure certificate I can use now standard unpatched VNC.  I just have to click again "_Yes_" for "_Hostname (*) does not match any certificate, do you want to continue?_". The important file is: `~/.vnc/x509_savedcerts.pem`
[x509_savedcerts.pem.zip](https://github.com/TigerVNC/tigervnc/files/1202655/x509_savedcerts.pem.zip)

bphinz:
Hmm.  If that's the cert from 15191.vm91.wedos.net:25191 then it's SHA1 so I don't know why you're getting the GNUTLS_CERT_INSECURE_ALGORITHM.

PR #481, opened 2017-06-28
Support for building with AOSP

cyanogen:
First patch adds support for compilation of a static library for Android. The second patch cleans up most of the warnings emitted by Clang on Android Nougat. I've also written an server implementation which uses the virtual display feature of Surfaceflinger. Android doesn't permit input injection from normal apps, so it only works if bundled with the system image (or as part of a custom rom / rooted device). If it's something that makes sense as part of the TigerVNC codebase, I'd be happy to clean it up and submit it. The server code is currently here: https://github.com/cyanogen/vncflinger/tree/tigervnc

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -1,0 +1,3 @@
+LOCAL_PATH := $(call my-dir)
+
+include $(call first-makefiles-under,$(LOCAL_PATH))
--- a/common/Android.mk
+++ b/common/Android.mk
@@ -90,11 +90,8 @@ LOCAL_C_INCLUDES := \
     external/libjpeg-turbo \
 	external/zlib
 
-LOCAL_CFLAGS := -Ofast -Wall -Wformat=2 -DNDEBUG -UNDEBUG
-
-# !! FIXME FIXME FIXME !!
-LOCAL_CFLAGS += -Wno-non-virtual-dtor -Wno-overloaded-virtual -Wno-mismatched-tags \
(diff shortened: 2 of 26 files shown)


CendioOssman:
In general, an Android port would be welcome. However requiring rooted devices is a bit too large of a hurdle to motivate the extra code though. We would also want to avoid the duplication of the build system. :/

The cleanups shouldn't be a problem though. But could you split them up according to what kind of problem they fix? "Fix warnings" is a bit vague. :)

cyanogen:
I suppose the Android.mk could be generated but to build with AOSP a trivial makefile is required.

cyanogen:
Also, the use case for the server is geared toward system builders on odd form factor devices which might not have attached input devices.

CendioOssman:
> I suppose the Android.mk could be generated but to build with AOSP a trivial makefile is required.

That would be fine. It's mostly the duplication of source file lists I'm concerned about.

> Also, the use case for the server is geared toward system builders on odd form factor devices which might not have attached input devices.

A reasonable use case, but it does sound very niche. We are rather understaffed here, so I'd rather not extend things to new platforms without a decent user (and hopefully developer) base.

Could the server work on non-root devices in a view only mode?

cyanogen:
Yes, view-only would work without root.

cyanogen:
Closing this out, will submit new requests.

PR #484, opened 2017-07-08
Fix all warnings emitted by Clang in common code

cyanogen:
Android uses Clang with relatively strict options as it's default compiler. This series of patches eliminates all warnings from the common core and allows building with Werror. Patches are split up by error/issue type.

### DIFF
--- a/common/rdr/FileInStream.cxx
+++ b/common/rdr/FileInStream.cxx
@@ -70,12 +70,12 @@ int FileInStream::overrun(int itemSize, int nItems, bool wait)
 
   while (end < b + itemSize) {
     size_t n = fread((U8 *)end, b + sizeof(b) - end, 1, file);
-    if (n < 1) {
-      if (n < 0 || ferror(file))
+    if (n == 0) {
+      if (ferror(file))
         throw SystemException("fread", errno);
       if (feof(file))
         throw EndOfStream();
-      if (n == 0) { return 0; }
+      return 0;
     }
     end += b + sizeof(b) - end;
   }
--- a/common/rfb/util.h
+++ b/common/rfb/util.h
@@ -38,6 +38,10 @@ struct timeval;
 #  define __printf_attr(a, b)
 #endif // __GNUC__
 
+#ifndef __unused
+#  define __unused __attribute((__unused__))
+#endif
+
 namespace rfb {
 
   // -=- Class to handle cleanup of arrays of characters
--- a/common/network/Socket.h
+++ b/common/network/Socket.h
@@ -82,6 +82,7 @@ namespace network {
   class ConnectionFilter {
   public:
     virtual bool verifyConnection(Socket* s) = 0;
+    virtual ~ConnectionFilter() {}
   };
 
   class SocketListener {
--- a/common/rdr/FdInStream.h
+++ b/common/rdr/FdInStream.h
@@ -30,6 +30,7 @@ namespace rdr {
   class FdInStreamBlockCallback {
   public:
     virtual void blockCallback() = 0;
+    virtual ~FdInStreamBlockCallback() {}
   };
 
   class FdInStream : public InStream {
--- a/common/rfb/CMsgWriter.cxx
+++ b/common/rfb/CMsgWriter.cxx
@@ -239,7 +239,7 @@ void CMsgWriter::pointerEvent(const Point& pos, int buttonMask)
 }
 
 
-void CMsgWriter::clientCutText(const char* str, rdr::U32 len)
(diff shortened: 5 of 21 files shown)


cyanogen:
Also updated my android branch to generate the Android.mk from a template. Will keep that separate for now.

common/rfb/Logger.cxx
```diff
@@ -37,6 +37,9 @@ Logger::~Logger() {
   // *** Should remove this logger here!
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wformat-nonliteral"
```
CendioOssman: Rather than simply silencing this, I think we should try to help clang out. Apparently it wants a `__printf__` attribute for these cases:

https://clang.llvm.org/docs/AttributeReference.html#id149

See `LogWriter.h` for something similar.

common/rdr/FileInStream.cxx
```diff
@@ -56,7 +56,7 @@ int FileInStream::pos()
   return ftell(file) + ptr - b;
 }
 
-int FileInStream::overrun(int itemSize, int nItems, bool wait)
+int FileInStream::overrun(int itemSize, int nItems, bool /* wait */)
```
CendioOssman: As you've noticed there will just be lots and lots of these, and they will continue to pop up. Can you handle this via a compiler flag instead?

common/rfb/CMsgReader.cxx
```diff
@@ -205,8 +205,8 @@ void CMsgReader::readSetXCursor(int width, int height, const Point& hotspot)
   if (width > maxCursorSize || height > maxCursorSize)
     throw Exception("Too big cursor");
 
-  rdr::U8 pr, pg, pb;
-  rdr::U8 sr, sg, sb;
+  rdr::U8 pr = 0, pg = 0, pb = 0;
+  rdr::U8 sr = 0, sg = 0, sb = 0;
```
CendioOssman: I'm not a huge fan of this as it basically just turns off the warning and might hide future problems. Better to restructure things. You should be able to move the following section in to the if clause. That would also allow you to rescope those variables.

common/rdr/FileInStream.cxx
cyanogen: Yeah, I was on the fence about this. It's not necessary. However there are a bunch of no-op/stub methods in various header files which cause warnings to spill over into users of these headers.  If I want to include PixelBuffer.h (and many others), I have to add my own flags to hush warnings which I don't like. It's probably better to move some of these to the implementation files, would you agree?

common/rfb/CMsgReader.cxx
cyanogen: Gonna drop this one, Clang isn't actually emitting anything here only GCC.

common/rfb/Logger.cxx
cyanogen: Cool, that works.

common/rdr/FileInStream.cxx
cyanogen: Updated patchset only touches PixelBuffer.h, InputHandler.h, and SDesktop.h in a much nicer way.

common/rfb/util.h
```diff
@@ -40,6 +40,8 @@ struct timeval;
 
 namespace rfb {
 
+  template <typename T> void UNUSED(const T&) {}
```
CendioOssman: Does LLVM support the `unused` attribute:

https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes

That would be a lot clearer than trying to fool the compiler like this.

cyanogen: Seems to work now, let's do that.

common/rfb/util.h
```diff
@@ -38,6 +38,10 @@ struct timeval;
 #  define __printf_attr(a, b)
 #endif // __GNUC__
 
+#ifndef __unused
+#  define __unused __attribute((__unused__))
```
CendioOssman: Unfortunately the name `__unused` causes breakage:

```
In file included from /usr/include/sys/stat.h:105:0,
                 from /usr/include/FL/fl_utf8.h:61,
                 from /usr/include/FL/Fl.H:30,
                 from /home/ossman/devel/tigervnc/vncviewer/CConn.h:23,
                 from /home/ossman/devel/tigervnc/vncviewer/DesktopWindow.cxx:37:
/usr/include/bits/stat.h:103:22: error: expected unqualified-id before '[' token
     long int __unused[3];
                      ^
```

CendioOssman:
Fixed and merged.

Issue #361, opened 2016-09-28
No cursor with x0vncserver


pjlobo:
Executive summary: There is no cursor in viewer using TurboVNC viewer 2.1 running on a Debian 8 machine to connect to an Ubuntu 12.04 (yes, I know it's old, but can't upgrade now) server running TigerVNC x0vncserver 1.7.0.

I've come to this combination because TurboVNC does not have a x0vncserver (which I need) and TigerVNC viewer does not support scaling (which is nice to have). Using TigerVNC at both sides works fine, but there's no scaling. Both TurboVNC and TigerVNC viewers with x11vnc server work fine, but this is waaay slower.

I've tried all viewer and server options (i.e. local/server cursor, scaling on/off, etc.) to no avail. Can't say whether the problem is at the viewer or server side, so I'm opening an issue on both in the hope that this otherwise fantastic combination becomes useable.

Best regards.
##

CendioOssman:
I'm afraid x0vncserver doesn't support a mouse cursor in any way. So you need to use a client that supports a fallback cursor display. There is commonly an option to show a dot when there is no cursor.

mshaddock:
The option to display a dot when there is no cursor is better than nothing, but barely.  That dot is so small that it is hard to follow.  An option to specify the cursor shape, or even an option to specify an arrow cursor would help greatly.

Thanks.

romario74:
A workaround is to use the vnc.so module. I just tested it under RHEL7.3 (tigervnc-server-module-1.3.1-9.el7.x86_64), it shows me the server's mouse cursor.

alanc:
https://github.com/alanc/tigervnc/commit/9a739d4bc716d08fd6a24cb50ad1d6a8367dfbd3 is a quick & simple implementation using the XFixes extension to get the cursor image from the X server and display it.  (It doesn't cache cursor images - that might be more efficient, but seems pretty quick as is.)

PR #493, opened 2017-08-11
x0vncserver: Use Xfixes to display cursors if available

alanc:
https://github.com/TigerVNC/tigervnc/issues/361

This is a simple implementation that refetches and transforms the cursor
image every time it changes, and doesn't use the cursor naming functions
of the XFixes extension to save & cache cursor images.

Signed-off-by: Alan Coopersmith <alan.coopersmith@oracle.com>

### DIFF
--- a/unix/x0vncserver/x0vncserver.cxx
+++ b/unix/x0vncserver/x0vncserver.cxx
@@ -45,6 +45,9 @@
 #ifdef HAVE_XDAMAGE
 #include <X11/extensions/Xdamage.h>
 #endif
+#ifdef HAVE_XFIXES
+#include <X11/extensions/Xfixes.h>
+#endif
 
 #include <x0vncserver/Geometry.h>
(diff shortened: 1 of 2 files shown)


unix/x0vncserver/x0vncserver.cxx
```diff
@@ -174,6 +177,21 @@ class XDesktop : public SDesktop, public TXGlobalEventHandler
 #ifdef HAVE_XDAMAGE
     }
 #endif
+
+#ifdef HAVE_XFIXES
+    int xfixesErrorBase;
+
+    if (XFixesQueryExtension(dpy, &xfixesEventBase, &xfixesErrorBase)) {
+      if (not haveDamage)
+        TXWindow::setGlobalEventHandler(this);
```
CendioOssman: It's probably cleaner to always set the event handler and handle it there. Avoids interdependencies like this.

PR #497, opened 2017-08-16
Add support for additional mouse buttons

grulja:
This patch adds support for up to 9 mouse buttons. I tested it with xev and buttons 8 and 9 are send and recognized properly.

### DIFF
--- a/common/rfb/CMsgWriter.cxx
+++ b/common/rfb/CMsgWriter.cxx
@@ -232,7 +232,7 @@ void CMsgWriter::pointerEvent(const Point& pos, int buttonMask)
   if (p.y >= cp->height) p.y = cp->height - 1;
 
   startMsg(msgTypePointerEvent);
-  os->writeU8(buttonMask);
+  os->writeU16(buttonMask);
   os->writeU16(p.x);
   os->writeU16(p.y);
   endMsg();
--- a/common/rfb/SMsgReader.cxx
+++ b/common/rfb/SMsgReader.cxx
@@ -189,7 +189,7 @@ void SMsgReader::readKeyEvent()
 
 void SMsgReader::readPointerEvent()
 {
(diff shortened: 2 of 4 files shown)


CendioOssman:
This breaks the protocol so I'm afraid you cannot do it this way. You would need to do a VNC extension in order to be able to send a more advanced pointer event.

grulja:
Oh, I see, didn't realize that this is standardized.

grulja:
Can you be more specific about VNC extension? How does it work? Is that something like the client sends to the server a list of extensions it supports and if the server responds that some of them are supported, like the one for 9 mouse buttons (if such extension exists), then the client can use that?

CendioOssman:
Extensions are unfortunately a bit difficult in VNC as there is no really good handshake. Check the protocol specification here for details of how things work and for a bunch of extensions:

https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst

You can then look at something like the QEMU extension for an idea how to do a proper handshake for something like this.

I should also add that it might be worthwhile to think about what else might be added to an "advanced mouse event" since it is rather annoying to add new extensions.

hildred:
On Thu, Aug 24, 2017 at 3:33 AM, Pierre Ossman (Work account) <

> I should also add that it might be worthwhile to think about what else
> might be added to an "advanced mouse event" since it is rather annoying to
> add new extensions.
>
FYI: I am in the process of writing a proposal for an extension to X11
providing industrial button support  and bar-code events. If we were to
work together  a unified solution may be possible.


-- 
--
Ben Hildred
Automation Support Services

CendioOssman:
You might want to have a look at Xinput and libinput. Those provide a rather generic way of exposing input devices. Converting that to a VNC extension should be able to cover most things.

PR #499, opened 2017-08-21
Enable X509 configuration in winvnc

bphinz:


CendioOssman:
Seems you got a few extra commits there. ;)

bphinz:
Replaced with #500

PR #513, opened 2017-09-14
Support QEMU LED states and keyboard events

CendioOssman:
Adds support for QEMU's keyboard extensions to all servers and the native client. This means synchronisation of CapsLock/NumLock/ScrollLock between client and server, and "raw keyboard" mode where the server side keyboard layout is fully respected.

The LED state extension is always enabled, but the raw keyboard mode is turned off by default in the servers.


CendioOssman:
@bphinz, any possibility of supporting these extensions in Java?

PR #514, opened 2017-09-19
modernized to c++11 using clang-tidy

wutzi15:
Signed-off-by: Benedikt Bergenthal <benedikt@kdrennert.de>


CendioOssman:
Can these be split up to one type of change per commit? Makes it easier to see exactly what it is being changed.

I'm also concerned if the relevant compilers all support this. @bphinz, can you do a test build on EL6?

wutzi15:
I've added all the different modernizations as separate commits now. 

As for the relevant compilers: C++11 is now 6 years old and I hope it is well enough supported.

MarkMielke:
On Mon, Sep 25, 2017 at 4:38 AM, wutzi15 <notifications@github.com> wrote:

> I've added all the different modernizations as separate commits now.
>
> As for the relevant compilers: C++11 is now 6 years old and I hope it is
> well enough supported.
>

RHEL 6 was released November 10, 2010.

We still use RHEL 6 for many use cases. I compile TigerVNC 1.8.0 for it,
and use this instead of the version of TigerVNC that comes with RHEL 6.

I am also concerned that your suggested change may break this. While
staying up-to-date with C++ standard makes general sense - what is the real
value here, and how does it compare with the cost of dropping support for
older operating system releases?

-- 
Mark Mielke <mark.mielke@gmail.com>

MarkMielke:
On Mon, Sep 25, 2017 at 4:43 AM, Mark Mielke <mark.mielke@gmail.com> wrote:

>
> On Mon, Sep 25, 2017 at 4:38 AM, wutzi15 <notifications@github.com> wrote:
>
>> I've added all the different modernizations as separate commits now.
>>
>> As for the relevant compilers: C++11 is now 6 years old and I hope it is
>> well enough supported.
>>
>
> RHEL 6 was released November 10, 2010.
>
> We still use RHEL 6 for many use cases. I compile TigerVNC 1.8.0 for it,
> and use this instead of the version of TigerVNC that comes with RHEL 6.
>
> I am also concerned that your suggested change may break this. While
> staying up-to-date with C++ standard makes general sense - what is the real
> value here, and how does it compare with the cost of dropping support for
> older operating system releases?
>

I should have just checked. I checked now... RHEL 6 has GCC 4.4.7. C++11
was introduced in GCC 4.8.

You may be able to apply some of the features, as long as the feature is
listed as being provided by GCC 4.4, I think it should be ok:

https://gcc.gnu.org/projects/cxx-status.html

If you want to propose which features you want to leverage that should work
in GCC 4.4, I'm willing to try it out and help you test. But, for any
features not supported by GCC 4.4, I think they should not be accepted at
this time.

-- 
Mark Mielke <mark.mielke@gmail.com>

CendioOssman:
Right, so I did some testing and had a look here:

https://gcc.gnu.org/projects/cxx-status.html#cxx11

Which gives these results:

-std=c++11: NAK¹
C headers: OK
range: NAK (4.6)
raw string: NAK (4.5)
redundant void: OK
auto: OK²
booleans: OK
emplace: OK
nullptr: NAK³ (4.6)
override: NAK (4.7)
using: NAK (4.7)
braced init: OK
default: OK
deleted: OK

¹ Needs to be -std=c++0x on older g++
² Should have been done before the "range" commit
³ Might be possible to have a fallback

CendioOssman:
Ping @wutzi15

wutzi15:
I'm sorry for not answering, but I've been ill lately.

CendioOssman:
Any progress here? Otherwise I'm afraid we'll have to close this on account of inactivity.

Issue #160, opened 2015-03-31
support UNIX sockets for server and viewer


calestyo:
Hi.

It would be nice if the connection would work via UNIX sockets, for both the server and the client.

This would be a great security boost, since on a local system, connection to the server socket can be controlled via normal file permissions (which isn't possible for inet sockets), and to the remote the socket can be forwarded via SSH.

Cheers,
Chris.
##

CendioOssman:
Just a FYI, file permissions aren't respected on Unix sockets on all operating systems, and OpenSSH can forward Unix sockets. So security should be enforced using authentication.

But this might be a useful feature anyway. If nothing else it makes it easier to deal with port allocation.

calestyo:
Well than that's something the respective OS need to fix. And then of course SSH can also just forward those sockets, which it can access.

And getting other ways of proper authentication into VNC is probably a lost cause - actually I'd say that using the user permissions of the local system *is* the proper way, which is also what many other daemons (e.g. DBMS) offer and do.

hifi:
I've done something like this with no "upstream" support. The point is to run Xvnc on the remote server in classic inetd mode (stdin/stdout) and then forward that to local UNIX socket. Then, attach stdin/stdout of a ssh session to that socket. This works with any OpenSSH version and current TigerVNC Xvnc server. You also need some sort of session in $HOME/.xinitrc for it to do anything useful.

https://gist.github.com/hifi/4321ffe9d56612b2449f (vncpipe.pl in $HOME of target system)

Client side, I've used netcat with OpenSSH to convert the ssh stdin and stdout into a localhost bound TCP socket:

> nc -l 1234 -e "ssh user@host ./vncpipe.pl"

Finally connect TigerVNC without any authentication to localhost port 1234. This does work on Windows as well if you have OpenSSH ssh.exe and nc.exe.

Pros:
- Server side is secure with UNIX sockets
- Connection encryption and authentication done with SSH
- Multiple users can run their own Xvnc sessions simultaneously
- If you don't log out of the X session it will continue running in the background and reconnecting will resume it
- You can resize the client and server resizes with it (TigerVNC client/server feature)

Cons:
- Only one connection to the same Xvnc session because -inetd mode with vncpipe.pl
- Need to install vncpipe.pl and Perl on the target system
- Odd client side connection without support in TigerVNC
- Because of above, client depends on netcat (does work on Windows though!)

To fix all the cons:
[...]

PR #530, opened 2017-10-25
x0vncserver MIT-SHM, Xdesktop restructure, initial cursor

astrand:
Some improvements mostly related to the x0vncserver.


CMakeLists.txt
```diff
@@ -272,6 +272,11 @@ if(ENABLE_PAM)
 endif()
 set(HAVE_PAM ${ENABLE_PAM})
 
+# Check for MIT-SHM
+if(X11_XShm_FOUND)
+    add_definitions("-DHAVE_MITSHM")
```
CendioOssman: We already assume MITSHM in vncviewer, so I suggest simply removing the `#ifdef`s.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -0,0 +1,398 @@
...
+ * USA.
+ */
+
+#include <x0vncserver/XDesktop.h>
+
+using namespace rfb;
+
+extern rfb::BoolParameter useShm;
+extern rfb::BoolParameter useOverlay;
+extern rfb::BoolParameter rawKeyboard;
```
CendioOssman: These are only used from `XDesktop`, so let them live `XDesktop.cxx` as well.

unix/x0vncserver/XDesktop.h
```diff
@@ -0,0 +1,91 @@
...
+#endif
+#ifdef HAVE_XDAMAGE
+#include <X11/extensions/Xdamage.h>
+#endif
+#ifdef HAVE_XFIXES
+#include <X11/extensions/Xfixes.h>
+#endif
+
+#include <x0vncserver/Geometry.h>
+#include <x0vncserver/XPixelBuffer.h>
```
CendioOssman: Do all of these headers really need to be included in `XDesktop.h`? Isn't it sufficient to have them in `XDesktop.cxx`?

unix/x0vncserver/XDesktop.h
```diff
@@ -0,0 +1,91 @@
...
+
+// number of XKb indicator leds to handle
+#define XDESKTOP_N_LEDS 3
+
+extern const unsigned short code_map_qnum_to_xorgevdev[];
+extern const unsigned int code_map_qnum_to_xorgevdev_len;
+
+extern const unsigned short code_map_qnum_to_xorgkbd[];
+extern const unsigned int code_map_qnum_to_xorgkbd_len;
```
CendioOssman: These are only used by `XDesktop` so they don't need to be in the header file.

unix/x0vncserver/x0vncserver.cxx
```diff
@@ -39,27 +39,12 @@
...
-#include <X11/extensions/Xdamage.h>
-#endif
-#ifdef HAVE_XFIXES
-#include <X11/extensions/Xfixes.h>
-#endif
 
 #include <x0vncserver/Geometry.h>
 #include <x0vncserver/Image.h>
 #include <x0vncserver/XPixelBuffer.h>
 #include <x0vncserver/PollingScheduler.h>
```
CendioOssman: I suspect more of these can be removed?

common/rdr/FdOutStream.cxx
```diff
@@ -193,7 +194,14 @@ int FdOutStream::writeWithTimeout(const void* data, int length, int timeoutms)
...
 
   do {
+    // select only guarantees that you can write SO_SNDLOWAT without
+    // blocking, which is normally 1. Use MSG_DONTWAIT to avoid
+    // blocking, when possible.
+#ifdef WIN32
     n = ::write(fd, data, length);
+#else
+    n = ::send(fd, data, length, MSG_DONTWAIT);
+#endif
```
CendioOssman: Can't we just get rid of the WIN32 hack instead and call `send()` everywhere?

common/network/UnixSocket.cxx
```diff
@@ -0,0 +1,76 @@
...
+ * USA.
+ */
+
+#ifndef WIN32
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#define closesocket close
```
CendioOssman: Why? :)

common/network/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 include_directories(${CMAKE_SOURCE_DIR}/common)
 
 add_library(network STATIC
+  UnixSocket.cxx
```
CendioOssman: Does any patch also make use of this? Having a bunch of dead code isn't really desirable.

CendioOssman: And you can include `UnixSocket.cxx` condtionally. Avoids having a file that is completely `#ifdef`ed out.

unix/x0vncserver/x0vncserver.cxx
```diff
@@ -260,6 +260,10 @@ int main(int argc, char** argv)
 
   Configuration::enableServerParams();
 
+  // Disable configuration parameters which we do not support
+  Configuration::setParam("AcceptSetDesktopSize", "0");
```
CendioOssman: This is unnecessary as the default implementation will deny everything anyway.

common/network/CMakeLists.txt
astrand: Not yeat. We have https://github.com/TigerVNC/tigervnc/issues/160 though, but I can exclude UnixSocket.cxx if preferred?

common/network/UnixSocket.cxx
astrand: Because TcpSocket does it...

common/rdr/FdOutStream.cxx
astrand: WIN32 does not support MSG_DONTWAIT, but if you prefer, I can change to:
```
#ifdef WIN32
    n = ::send(fd, data, length, 0);
#else
    n = ::send(fd, data, length, MSG_DONTWAIT);
#endif
```

unix/x0vncserver/XDesktop.h
astrand: No, that's not sufficient:
```
  XDesktop(Display* dpy_, Geometry *geometry);
...
  XPixelBuffer* pb;
```

unix/x0vncserver/x0vncserver.cxx
astrand: Yes, but without this code, AcceptSetDesktopSize will be listed in --help output, which is confusing.

unix/x0vncserver/x0vncserver.cxx
astrand: Yes. In the updated request, I've moved 3 more, which are present in XDesktop.h. There are probably others as well (strings.h etc), but that's not related to moving XDesktop to a separate file.

common/network/UnixSocket.cxx
CendioOssman: But it does it conditionally. And it's only for Windows compatibility, which isn't needed here.

common/rdr/FdOutStream.cxx
CendioOssman: Well that's annoying. But yes, the alternative would be better. But I think you should have `#ifndef MSG_DONTWAIT` rather than assuming a direct correlation with `WIN32`.

unix/x0vncserver/XDesktop.h
CendioOssman: Those are just pointers so they don't need a full definition. Just add `class Geometry;` etc. beforehand. This is commonly done in lots of other headers in TigerVNC.

unix/x0vncserver/x0vncserver.cxx
CendioOssman: I meant specifically the `setParam()` call. The `removeParam()` call does indeed do useful things.

common/network/CMakeLists.txt
CendioOssman: I think so. It still has a lot of work left before it could be used in either the client or server.

common/network/UnixSocket.cxx
astrand: Correct, but it still makes the destructor of UnixSocket look like TcpSocket, which is nice I think. But in any case, I'll exclude UnixSocket for now.

common/rdr/FdOutStream.cxx
astrand: Can do, but also note that send() on Windows takes a "const char*" rather than a "void char*", so in that case, it will be:
```
#ifndef MSG_DONTWAIT
    n = ::send(fd, (const char*)data, length, 0);
#else
    n = ::send(fd, (const char*)data, length, MSG_DONTWAIT);
#endif
```
Do you like that?

PR #534, opened 2017-11-04
Fix build failure other than utf-8

otakuto:
This PR avoid build failures in environments other than utf-8.

This bug reported on gentoo bugzilla.
https://bugs.gentoo.org/610738
https://bugs.gentoo.org/634918


bphinz:
Thanks!  I was using JAVA_TOOL_OPTIONS="-Dfile.encoding=UTF8" for the other packages but this seems much cleaner.  Let me just remove that from the other package build scripts and do a test build tonight and then I'll merge this in.

bphinz:
Resolved in 335446d

PR #542, opened 2017-11-20
Remove " på"

scootergrisen:


CendioOssman:
Thank you for your contribution! Unfortunately we cannot do direct modifications of the translation files as they are handled by the [translation project](https://translationproject.org/html/welcome.html). The contact information for reporting bugs for Danish translations can be found here:

https://translationproject.org/team/da.html

PR #551, opened 2017-12-06
Add gitignore

tequillaz:
Add stadard ignores for C, C++, Java, CMake.

Based on [github gitignore](https://github.com/github/gitignore) project

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,69 +1,14 @@
 
 
-#######
-## C ##
-#######
-
-# Prerequisites
-*.d
-
-# Object files
-*.o
-*.ko
-*.obj
-*.elf
-
-# Linker output
-*.ilk
-*.map
-*.exp
-
-# Precompiled Headers
-*.gch
-*.pch
-
-# Libraries
-*.lib
-*.a
-*.la
-*.lo
-
-# Shared objects (inc. Windows DLLs)
-*.dll
-*.so
-*.so.*
-*.dylib
-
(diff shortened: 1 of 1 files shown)


CendioOssman:
Thank you, but personally I think this thing is massively excessive and difficult to work with. Could you reduce it to things that are actually relevant for TigerVNC?

tequillaz:
TigerVNC doesn't use c-lang, it is possible to remove c-lang part. All
other parts, yes, they are huge but it is community experience in
c++/java/cmake usage. I think we must save c++/java/cmake parts.

7 дек. 2017 г. 12:43 ПП пользователь "Pierre Ossman (Work account)" <
notifications@github.com> написал:

Thank you, but personally I think this thing is massively excessive and
difficult to work with. Could you reduce it to things that are actually
relevant for TigerVNC?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/TigerVNC/tigervnc/pull/551#issuecomment-349915313>, or mute
the thread

CendioOssman:
There's a bunch of crap in the other ones as well. E.g. `*.d` and `*.mod`. So I'm still :-1: for just putting this massive template in.

CendioOssman:
Thank you, a little bit better, but still lots of stuff left. I'd like to see this reduces to just the things we need. So please check the rest and remove the sections that are for things that we don't have in our project.

PR #552, opened 2017-12-06
Replacing some platform independant tools

tequillaz:
Replacing os/thread, os/mutex, os/condition by modern (hah!) analogs from std library


CendioOssman:
Cleanups are always nice, but unfortunately we want to support RHEL 6 for a while longer. And the gcc there doesn't support all of C++11. Please see #514 and see what changes you are proposing that might cause issues.

tequillaz:
Some time before try to compile this with gcc 4.4.7. With c++0x flag std::thread and std::mutex is supported but lamdas not. I can make patch If someone can test it on RHEL 6.

bphinz:
> I can make patch If someone can test it on RHEL 6.

I can run a test build for el6.  Just provide me with a branch name in your repo (if other than feature/replace_os_tools).

tequillaz:
@bphinz look at [feature/replace_os_tools__rhel6](/tequillaz/tigervnc/tree/feature/replace_os_tools__rhel6)

bphinz:
> @bphinz look at feature/replace_os_tools__rhel6

That failed to build on all platforms.  

Linux (it looks like at a minimum c++0x flag is needed):

`#error This file requires compiler and library support for the upcoming ISO C++ standard, C++0x. This support is currently experimental, and must be enabled with the -std=c++0x or -std=gnu++0x compiler options.`

OS X:

`/Users/bphinz/workspace/osx-binaries/build/tigervnc-1.8.80-20171224git8287affb/common/rfb/Configuration.h:46:10: fatal error: 'mutex' file not found
#include <mutex>`

Windows (MXE):

`Scanning dependencies of target os
[  1%] Building C object common/os/CMakeFiles/os.dir/w32tiger.c.obj
[  2%] Building CXX object common/os/CMakeFiles/os.dir/os.cxx.obj
/home/bphinz/workspace/win64-binaries/tigervnc-1.8.80-20171224git8287affb/common/os/os.cxx:28:17: fatal error: pwd.h: No such file or directory
 #include <pwd.h>`

CendioOssman:
Any progress on this?

PR #588, opened 2018-01-29
Rename README.txt to README.rst and convert to reStructuredText

astrand:


CendioOssman:
Seems okay, but please don't modify the `.po` files (or `.pot` file ideally) as those things are handled separately. It should sync up the next time the translations are done.

PR #597, opened 2018-02-08
Make sure auth argument is escaped

olivergondza:
Some options are already escaped, some are validated to be sure they do not have to but these are  still not.

I discovered this while running with `XAUTHORITY` file in directory containing `&&` that caused broken xvnc was launched as only the pre-`&&` portion was passed through `sh -c`.

### DIFF
--- a/unix/vncserver
+++ b/unix/vncserver
@@ -207,7 +207,7 @@ my %config;
 # override these where present.
 $default_opts{desktop} = &quotedString($desktopName);
 $default_opts{httpd} = $vncJavaFiles if ($vncJavaFiles);
-$default_opts{auth} = $xauthorityFile;
+$default_opts{auth} = &quotedString($xauthorityFile);
 $default_opts{geometry} = $geometry if ($geometry);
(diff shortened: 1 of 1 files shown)


CendioOssman:
The auth bit looks obvious, but why also pixelformat? It shouldn't have any need for special characters. Is it the only other variable not otherwise validated?

olivergondza:
I have tried my best (which is somewhat limited in PERL) to identify other variables that might be impacted and that's how pixelformat got into the fix. You are likely right there is no usecase for using strange characters there, though I prefer to quote such args anyway as the error will be issued instead of executions of unpredictable command. 

It took me quite a while to figure out why seemingly healthy xvnc process cannot be connected to before I realized half of the arguments are missing...

CendioOssman:
Well, all of these end up as arguments. So I think we should either go all out and quote everything, or be minimal and just quote the stuff that realistically might need quoting. Doing something in between is just confusing.

olivergondza:
Agreed, restricting escaping to what I know is broken. Please re-review.

PR #617, opened 2018-03-22
change parameter name

lean0708:
the "server" parameter should be named "Host" because the most vnc viewers recieves that parameter instead off "Server"


bphinz:
This change would not be backwards compatible.  I'd much rather see something that allows for Host or Server.

Issue #38, opened 2014-09-24
TigerVNC viewer should be able to open *.tigervnc files as an argument [$50]


simula67:
# Symptom

TigerVNC viewer says 

```
"unable to resolve host by name: No such host is known. (11001)" 
```

while trying to open a *.tigervnc file with ServerName given in <hostname>:<port> format.

Specifying this directly in TigerVNC viewer works as expected.
# How to reproduce

Save a *.tigervnc file with port name. Example with specifics removed follows :

```
TigerVNC Configuration file Version 1.0



ServerName=example.com:5901
x509ca=
x509crl=
SecurityTypes=X509Plain,TLSPlain,X509Vnc,TLSVnc,X509None,TLSNone,VncAuth,None
DotWhenNoCursor=0
AutoSelect=1
FullColor=1
LowColorLevel=2
PreferredEncoding=Tight
CustomCompressLevel=0
CompressLevel=2
NoJPEG=0
QualityLevel=8
FullScreen=0
FullScreenAllMonitors=1
DesktopSize=
geometry=
RemoteResize=1
ViewOnly=0
Shared=0
AcceptClipboard=1
SendClipboard=1
SendPrimary=1
MenuKey=F8
FullscreenSystemKeys=1
```

Now, use Windows to associate *.tigervnc files to be opened by vncviewer executable.
Try opening the file, it shows the error : 

```
"unable to resolve host by name: No such host is known. (11001)"
```
# OS

Windows
# Expected behaviour

When *.tigervnc file contains ServerName key in the format <hostname>:<port>, vncviewer should attempt to connect to <hostname> at <port>.
##

gqgunhed:
Same behavior here when using IP-addresses and port like "ServerName=192.168.0.10:1".
As far as I looked through the code it seems to be located within tigervnc/vncviewer/parameters.cxx, here lines 574 and following.
I cannot see any code splitting the string after "ServerName=" at the ":" if present into hostname/IP-address and port.

CendioOssman:
It supports this just fine. What it doesn't support is being given configuration files as a command line argument. It will then use the filename as the host to connect to.

So this is really a feature request for opening configuration files on the command line.

gqgunhed:
Thanks for clarification. So: When windows-double-clicking on (the associated) .tigervnc file the _filename_ gets handed over to TigerVNC, not the content of the config file itself. Your change of the subject is correct then.

CendioOssman:
[...]

PR #632, opened 2018-04-22
Accept a cfg as an argument as an alt way to start the viewer

Kypert:
The user can specify a tigervnc configuration file as an argument to the
viewer. Previously the viewer assumed this to be a server, but now we
will first check if there is any file matching the given argument. If
so, try to load the content of that file, like we normally do.

Fixes issue #38.

### DIFF
--- a/vncviewer/vncviewer.cxx
+++ b/vncviewer/vncviewer.cxx
@@ -349,8 +349,9 @@ static void usage(const char *programName)
 
   fprintf(stderr,
           "\nusage: %s [parameters] [host:displayNum] [parameters]\n"
-          "       %s [parameters] -listen [port] [parameters]\n",
-          programName, programName);
+          "       %s [parameters] -listen [port] [parameters]\n"
+          "       %s [parameters] [.tigervnc file]\n",
+          programName, programName, programName);
   fprintf(stderr,"\n"
           "Parameters can be turned on with -<param> or off with -<param>=0\n"
           "Parameters which take a value can be specified as "
@@ -366,6 +367,31 @@ static void usage(const char *programName)
 #endif
 
   exit(1);
+}
+
+static void
+potentiallyLoadConfigurationFile(char *vncServerName)
+{
+  const bool hasPathSeparator = (strchr(vncServerName, '/') != NULL ||
+                                 (strchr(vncServerName, '\\')) != NULL);
+
+  if (hasPathSeparator) {
+    try {
+      strncpy(vncServerName, loadViewerParameters(vncServerName),
+              VNCSERVERNAMELEN);
+      if (vncServerName[0] == '\0') {
+        vlog.error("Unable to load the server name from given file");
+        if (alertOnFatalError)
+          fl_alert("Unable to load the server name from given file");
+        exit(EXIT_FAILURE);
+      }
+    } catch (rfb::Exception& e) {
+      vlog.error("%s", e.str());
+      if (alertOnFatalError)
+        fl_alert("%s", e.str());
+      exit(EXIT_FAILURE);
+    }
+  }
 }
 
 #ifndef WIN32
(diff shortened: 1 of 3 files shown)


CendioOssman:
Thanks. This looks nice. Just two things:

1. Ignoring errors from `fopen()` is probably confusing. It's better if the user gets a proper permission denied error or such in those cases. What we should do is check if the file exists using `stat()`.
2. I'm wondering if there is a possible security issue here if we prefer files over hostnames. E.g. someone manages to sneak in a file called `myserver.example.com` in the working directory of vncviewer. Should we perhaps require a path separator in the filename? I.e. you would have to specify `./myconfig.tigervnc` or an absolute path.

CendioOssman:
> 1. Reason for ignoring the errors is that I wanted to keep the implementation simple. I have started on an implementation that uses stat(), which is easy for the Linux target, but since I do not have the mingw environment setup, I am not sure if I need to take special consideration for that target (will work on installing the cross-compiler correctly). Also breaking it out to a separate helper function.
> 

I think mingw makes sure that `stat()` behaves sanely even when compiling for Windows.

> 2. I can see your point, but for the Windows users, I am not sure the filename would always come as an absolute or path-separated relative path (will investigate, unless you know for sure?). I suppose it is fine to add the constraint for the console users but I am thinking about the file associated .tigervnc case:

I'm fairly sure you always get an absolute path in those cases. A relative path is too risky in those cases.

Kypert:
> I'm fairly sure you always get an absolute path in those cases. 

Seems like you are indeed right, from my own manual testing.

Working on setting up the mingw environment, unfortunately long since I last used it so I need to read up a bit, sorry for the delays!

Kypert:
@CendioOssman: Managed to setup mingw, attempted to address the issues you pointed out, please see the amended commit.

vncviewer/vncviewer.cxx
```diff
@@ -368,6 +369,56 @@ static void usage(const char *programName)
...
+    }
+  }
+  else {
+    if (S_ISREG(st.st_mode)) {
+#ifndef WIN32
+      const bool absolute = vncServerName[0] == '/';
+#else
+      const bool absolute = vncServerName[0] == '\\' || vncServerName[1] == ':';
+#endif
+      if (!absolute && vncServerName[0] != '.') {
```
CendioOssman: You can have normal files that start with `.` (and maybe even a hostname?).
I suggest you simply look for a path separator (`\` on Windows and `/` on Unix). No hostname can contain those. So you can do that check first and avoid the `stat()` call completely.

Kypert: Thank you for the comment! I like the comment about `/` and `\`, but `stat()` also serves the purpose to make sure the file exists and that the user has the access, like you proposed in previous attempt. Need to establish that knowledge before calling `loadViewerParameters()`. I can return back to `fopen()` and make some special handling for the permission denied case, if you like?

CendioOssman: What extra error handling do we actually need? Just call `loadViewerParameters()` and handle any exception. It will tell you why the file couldn't be opened.

Kypert: Originally I wanted to dodge any failures in `fopen()` in the event that the given argument in fact was a host. However, with your latest comment about checking for the path separator, we can distinguish if we have a file or hostname directly (or at least supposed to be). I understand now and will change accordingly, thank you for the persistently good code review comments! Please bear with me :)

Kypert:
@CendioOssman, please let me know if any other issues remain in the latest amendment.

CendioOssman:
The C++ code looks good now. :)

But the Java code should be changed to match so we don't have different behaviours.

(And we don't need the amendments in the commit message. Your original comment works fine. :) )

Kypert:
Thanks again for your patience! I have aligned the two clients (added checking for empty server name in the C++ version too) and removed the amendment comments in the commit message. I am still learning best practices when working in GitHub, so I appreciate all the pointers you can give me :)

@CendioOssman, notice anything else that should be done here?

gitcnd:
Was there some reason why this was not done as a switch?
Having a parameter which means 2 different things depending on presence of files seems less-than-standard

CendioOssman:
A switch often makes it more difficult to integrate with "Open with" features in other programs.

Issue #91, opened 2014-12-12
full screen over multiple, but not all monitors


ercanal:
It will be great which monitors  can be used by vncviewer. 
Ex:
2 Monitors -> Vnc Viewer Full Screen
1 Monitor -> Other programs
##

jrwagz:
I too would like to be able to do this.  I tried the following to see if it would work, and had no luck :(

- enable full screen over all monitors in TigerVNC client
- while in full screen over all monitors, go to the GNOME display settings, and disable one of the monitors

While this does affect the overall display, it doesn't do what you would expect, or anything useful.  I have 4 monitors arranged horizontally, and when I disabled one of them I ended up getting 3 "monitors" centered across the 4 monitors, with empty black space on the far left and right, and the monitor transitions occurred in the middle of the actual monitors.

dezza:
How do you go fullscreen over a single monitor?

When I try "vncviewer host:displaynum" I think it interprets it as a port.

CendioOssman:
> How do you go fullscreen over a single monitor?

It's in the viewer options dialog. It will go fullscreen over the same monitor the window is currently in.

> When I try "vncviewer host:displaynum" I think it interprets it as a port. It won't work.

`displaynum` here means display in the X11 sense of the word. It does not have anything to do with monitors.

luisgutz:
Would be great to do this.
I have the same configuration as ercanal: 3 monitors, and I would like to use 1 for windows/email.

fbordeu:
Same case for me, I have 3 screens (I'm in windows connectig to a remot linux). I will line 1 monitor for windows and 2 monitors for tigervnc in fuul screen mode.

wickedviking:
I can tell you that the first VNC client to support this feature will become the go to VNC client for windows users. I"ve been wanting this for years. I don't want to mess with ugly geometry hacks because I have different resolutions in the office than I do at home. I just want to open my VNC client, tell it which monitors to use, and connect to the remote system...

3 monitors in both home and office, home is 4k, office is 1080p. This all or one thing is for the birds...

microwaveguy:
I've been following this issue for a while. Until a recent search and found the following link:
https://superuser.com/questions/827969/how-to-configure-realvnc-viewer-to-display-on-only-specific-monitors-instead-of
I've tested today, it worked pretty well with my 3-monitor setting. Still in 30-day free trial mode, not sure what will happen in free mode, though. But could worth a purchase.

[...]

PR #642, opened 2018-05-07
Recreate fullscreen window if screens are removed or added

astrand:
As the code comment says, this functionality is not available on macOS, but I don't think that should prevent us from adding this functionality on the other platforms.


jcpunk:
I was sent here by https://www.fltk.org/str.php?L3462.  Any chance to get this merged?

CendioOssman:
I think this is good to go (a bit of a shame that macOS doesn't work though). But I'm concerned that there are a number of users relying on this not working as a workaround for #91. :/

CendioOssman:
So #91 is now fixed, which means this can be merged.

PR #656, opened 2018-06-07
Feature/recent connection list

oscarfmg:
As a vnc user I usually have to connect to several servers. As a matter of fact it's somehow hard to remember multiple vnc server names or load configuration form multiple locations.

I've implemented a list with the most recent servers used through the vncviewer in order to make my daily job easier.

It maybe would be helpful for anyone.

The trick here is that I've used c++11 features in order to make it easier. Then you must have to compile it with gcc 7 or greather. Or if you want to use gcc 4.7 you must append the CXX_FLAG -std=c++11 to the project.

I can rewrite to use only cxx03 specs but It will demand more time in order to get it ready.

### DIFF
--- a/vncviewer/CMakeLists.txt
+++ b/vncviewer/CMakeLists.txt
@@ -5,6 +5,7 @@ include_directories(${CMAKE_SOURCE_DIR}/common)
 set(VNCVIEWER_SOURCES
   menukey.cxx
   CConn.cxx
+  ConnectionsTable.cxx
   DesktopWindow.cxx
   UserDialog.cxx
   ServerDialog.cxx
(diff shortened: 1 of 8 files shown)


CendioOssman:
I'm afraid we want to support RHEL 6 for a while longer, so gcc 4.4.7 is the baseline. That still has some C++11 features, but it does require more care.

oscarfmg:
Hello again.
Finally I've refactorized code and now it doesn't require the C++11 features.

CendioOssman:
This would be a very nice feature. :)

However it is pretty intrusive in the UI for just a server name cache. Have you looked at doing an auto-complete/drop down in the server field? That would be similar to many other applications, most notably browsers.

oscarfmg:
Sorry I was out of time those days.
Ok. Let me check how to add the history as a dropdown rather than a table.

CendioOssman:
Any luck? :)

PR #669, opened 2018-06-27
Avoid error/warning about implicit function declaration

grulja:
Function xorgGlxCreateVendor() is defined in glx_extinit.h, if this header is not included, we might get either error or warning. This header also needs to be included after scrninststr.h header as it defines some structures used in glx_extinit.h


CendioOssman:
Thank you for your contribution.

This header is only present in newer versions of Xorg, so there needs to be a check. I've committed things as 8c6c5843 and 423e04fc.

PR #674, opened 2018-07-07
viewer: reset ctrl / alt to menu state on focus

martinetd:
Setting Ctrl or Alt key on menu only sends the key press, and the
state is lost when focus is lost and recovered.
This checks the menu variable and sends the keys again if needed.

PS: thanks for tigervnc! Feel free to ask me to change anything, this is a very small patch and I tried to respect style but I'll admit to not having read any contributing rules.


CendioOssman:
Nice. Thanks.

PR #676, opened 2018-07-10
Add a basic gitignore

clbr:
I looked at #551, and made this by hand.


clbr:
The ignore-everything-except-this-subdir logic doesn't work in the same
level, so that will have go to one level above (unix/.gitignore instead
of unix/xserver/.gitignore). Moving the other things.

CendioOssman:
You sure? This seems to work for me:
```
$ cat unix/xserver/.gitignore 
/*
!/hw
/hw/*
!/hw/vnc
```

clbr:
Oh, I tried with just "*", without the leading /. Trying now.

clbr:
Yes, works, pushed.

CendioOssman:
Great. Thanks.

Issue #672, opened 2018-07-06
Variable JPEG Quality


mmcclaskey:
I know it is outside the specifications, however, I would like to know if it is possible to have a variable JPEG quality based on the rate of change. When change rate slows back down the a high quality block will be sent. It would have to be on a block by block basis, otherwise an ad on a web page would result in the entire screen looking bad. I'm looking for ways to increase efficiency in order to provide scalability in an a large deployment. Any ideas would be greatly appreciated. I'm willing to pay a bounty for ideas that can work towards more scalability from a network load point of view (bandwidth usage). I know a lot of work in the community has already gone into this issue, however, I believe more efficiency may be possible.

clbr:
I'm currently investigating, but I believe dynamic jpeg quality is relatively easy to implement, and would work in all existing VNC viewers that support jpeg (tight). It should have nice b/w savings indeed.

If targeting various existing viewers, it would probably be best as a server option, something like "-dynamicJpeg min,max", which would enable it for all clients that request the tight encoding.

If you have cpu to burn, and can also modify the viewers, there's a few things possible (in order of efficiency and cpu usage):
- dropbox lepton can recompress jpeg losslessly, saving about 22%
- webp about halves the size at the same quality
- av1's image format beats webp at about 20%, meaning 60% over jpeg

These all tend to be cpu-heavy though, especially av1 as it hasn't been optimized yet. You should try each of them on your target server, encode a full-screen image, measure how much cpu it took, and multiply by the number of clients you expect. That'd be a reasonable approximation of the maximum cpu load.

mmcclaskey:
Personally, I am interested in noVNC for the client. Unfortunately that limits what can be done on the client side. I have tested JPEG2000 using a javascript decoder and writing to a canvas, the performance was horrible. JPEG2000 is natively supported in Safari only. Webp is supported natively in Chrome base browsers, so performance might be good there. JPEG XR is supported in IE and Edge and is similar to JPEG2000. AV1 might be an option, I haven't previously considered that, however, I did read this article which seems to indicate that using video encoding for vnc has its downfalls as well...

https://turbovnc.org/About/H264

All that being said, probably the easy kill would be dynamic jpeg quality for round 1. I'm not sure if the tigerVNC community would desire supporting JPEG2000, Webp, and JPEG-XR to support all the native browser implementations. However, if the desire was there I post a separate bounty for that.

mmcclaskey:
[...]

PR #677, opened 2018-07-10
Add support for dynamic JPEG quality adjustment

clbr:
These commits add support for dynamic JPEG quality adjustment, requested in #672.

It's added as a server-side option, where if enabled, all clients supporting the Tight
encoding with jpeg images get it. It can be configured with the min and max parameters,
giving a range of quality values, using the same 0-9 scale as the client-side quality
option.

I ran some tests compared to the defaults (jpeg at constant level 8). First was a web
page with text and two animated banners. Using the full dynamic range, 0-9, the numbers
reported by Xvnc for Tight (JPEG) improved by 12%. Range 2-8 improved by 21%.

Then I tried playing video, a 10-sec clip from a movie. Range 0-9 improved over the
baseline by 25%, and range 2-8 by 29%.

I couldn't notice visual degradation, so setting the maximum even lower may work, depending
on workload. Fully compatible with existing VNC viewers, I tested two old binaries from
2010 and 2008.

### DIFF
--- a/common/rfb/EncodeManager.cxx
+++ b/common/rfb/EncodeManager.cxx
@@ -550,7 +550,7 @@ Encoder *EncodeManager::startRect(const Rect& rect, int type)
   encoder = encoders[klass];
   conn->writer()->startRect(rect, encoder->encoding);
 
-  if (encoder->flags & EncoderLossy)
+  if (encoder->flags & EncoderLossy && !encoder->treatLossless())
     lossyRegion.assign_union(Region(rect));
   else
     lossyRegion.assign_subtract(Region(rect));
--- a/common/rfb/EncodeManager.cxx
+++ b/common/rfb/EncodeManager.cxx
@@ -1,6 +1,7 @@
 /* Copyright (C) 2000-2003 Constantin Kaplinsky.  All Rights Reserved.
  * Copyright (C) 2011 D. R. Commander.  All Rights Reserved.
  * Copyright 2014-2018 Pierre Ossman for Cendio AB
+ * Copyright (C) 2018 Lauri Kasanen
  * 
  * This is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -24,6 +25,7 @@
 #include <rfb/Encoder.h>
 #include <rfb/Palette.h>
 #include <rfb/SConnection.h>
+#include <rfb/ServerCore.h>
 #include <rfb/SMsgWriter.h>
 #include <rfb/UpdateTracker.h>
 #include <rfb/LogWriter.h>
@@ -38,6 +40,9 @@
 using namespace rfb;
 
 static LogWriter vlog("EncodeManager");
+
+// If this rect was touched this update, add this to its quality score
+#define SCORE_INCREMENT 8
 
 // Split each rectangle into smaller ones no larger than this area,
 // and no wider than this width.
@@ -75,6 +80,12 @@ enum EncoderType {
 struct RectInfo {
   int rleRuns;
   Palette palette;
+};
+
+struct QualityInfo {
+  struct timeval lastUpdate;
+  Rect rect;
+  unsigned score;
 };
 
 };
@@ -123,7 +134,8 @@ static const char *encoderTypeName(EncoderType type)
   return "Unknown Encoder Type";
 }
 
-EncodeManager::EncodeManager(SConnection* conn_) : conn(conn_)
+EncodeManager::EncodeManager(SConnection* conn_) : conn(conn_),
+  dynamicQualityMin(-1), dynamicQualityOff(-1)
 {
   StatsVector::iterator iter;
 
@@ -146,6 +158,12 @@ EncodeManager::EncodeManager(SConnection* conn_) : conn(conn_)
     for (iter2 = iter->begin();iter2 != iter->end();++iter2)
       memset(&*iter2, 0, sizeof(EncoderStats));
   }
+
+  if (Server::dynamicQualityMax && Server::dynamicQualityMax <= 9 &&
+      Server::dynamicQualityMax > Server::dynamicQualityMin) {
+    dynamicQualityMin = Server::dynamicQualityMin;
+    dynamicQualityOff = Server::dynamicQualityMax - Server::dynamicQualityMin;
+  }
 }
 
 EncodeManager::~EncodeManager()
@@ -156,6 +174,9 @@ EncodeManager::~EncodeManager()
 
   for (iter = encoders.begin();iter != encoders.end();iter++)
     delete *iter;
+
+  for (std::list<QualityInfo*>::iterator it = qualityList.begin(); it != qualityList.end(); it++)
+    delete *it;
 }
 
 void EncodeManager::logStats()
@@ -259,8 +280,10 @@ void EncodeManager::pruneLosslessRefresh(const Region& limits)
 }
 
 void EncodeManager::writeUpdate(const UpdateInfo& ui, const PixelBuffer* pb,
-                                const RenderedCursor* renderedCursor)
+                                const RenderedCursor* renderedCursor,
+                                size_t maxUpdateSize)
 {
+    curMaxUpdateSize = maxUpdateSize;
     doUpdate(true, ui.changed, ui.copied, ui.copy_delta, pb, renderedCursor);
 }
 
@@ -316,6 +339,8 @@ void EncodeManager::doUpdate(bool allowLossy, const Region& changed_,
 
     writeRects(changed, pb);
     writeRects(cursorRegion, renderedCursor);
+
+    updateQualities();
 
     conn->writer()->writeFramebufferUpdateEnd();
 }
@@ -804,6 +829,14 @@ void EncodeManager::writeSubRect(const Rect& rect, const PixelBuffer *pb)
   if (encoder->flags & EncoderUseNativePF)
     ppb = preparePixelBuffer(rect, pb, false);
 
+  if (type == encoderFullColour && dynamicQualityMin > -1) {
+    trackRectQuality(rect);
+
+    // Set the dynamic quality here. Unset fine quality, as it would overrule us
+    encoder->setQualityLevel(scaledQuality(rect));
+    encoder->setFineQualityLevel(-1, subsampleUndefined);
+  }
+
   encoder->writeRect(ppb, info.palette);
 
   endRect();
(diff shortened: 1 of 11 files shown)


clbr:
algo
Here's a graph showing how the algorithm currently picks qualities. It takes about one second for constant fps content to converge, and the same for the scoring to reset once it stops changing.

60fps content converges to the lower quality limit, 30fps to the middle, and slow changes barely drop in quality.

CendioOssman:
Thank you for looking at this. However there are two design things that would need to be adjusted:

a) This should be in `EncodeManager` so that the principle can be extended to other encodings in the future.

b) A fixed ratio between rate of change and quality is probably not a good idea. If there are enough resources (bandwidth, CPU) then giving a low quality image is not what we wont. The user will get a bad experience, and we might not really save that much as the lossless refresh will kick in and just resend the same image right away.

I think we should aim to do something that works well enough that we don't need a setting to disable it. Compression settings are black magic to most people.

clbr:
Thanks for the quick review.

> a) This should be in `EncodeManager` so that the principle can be extended to other encodings in the future.

What kind of interface do you prefer? Perhaps passing the rate of
change number to writeRect()?

> b) A fixed ratio between rate of change and quality is probably not a good idea. If there are enough resources (bandwidth, CPU) then giving a low quality image is not what we wont. The user will get a bad experience, and we might not really save that much as the lossless refresh will kick in and just resend the same image right away.
> 
> I think we should aim to do something that works well enough that we don't need a setting to disable it. Compression settings are black magic to most people.

How to measure the target bandwidth? A server param for global bw, or
per-client bw? CPU is easier, but bandwidth would need to be supplied
by the user somehow.

CendioOssman:
> What kind of interface do you prefer? Perhaps passing the rate of change number to writeRect()?

We already have `Encoder::setCompressLevel()` so I don't think any API changes are needed.

> How to measure the target bandwidth? A server param for global bw, or per-client bw? CPU is easier, but bandwidth would need to be supplied by the user somehow.

We already have bandwidth estimation from the congestion control, which is currently used for lossless refresh (look in `VNCSConnectionST::writeDataUpdate()`). However we need a fallback for clients that don't support the extensions needed for that though.

I don't think we can pass it off to the user. This is the kind of thing that needs to be self adjusting. Network conditions aren't static after all. A crude algorithm is fine if it works well enough in practice though.

clbr:
The original request left the impression he wanted to leave bandwidth
for other things too, in which case automatic measurement by Xvnc
wouldn't be suitable. @mmcclaskey can you comment?

I wonder about setCompressLevel, seeing it's currently used to set the
zlib and tight encoder levels by the client. The clients have a default
for that, so it's always sent, no? In that case, the new logic would
always automatically override the client, not sure if that's desired
without an explicit server param.

Do you have any ideas on the bw measurement fallback? There don't seem
to be timestamps or such that could be used to see if a packet was
delayed.

CendioOssman:
> The original request left the impression he wanted to leave bandwidth for other things too, in which case automatic measurement by Xvnc wouldn't be suitable. @mmcclaskey can you comment?

I'd say that is an entirely different can of worms. But if we were to add artificial bandwidth restrictions, then automatic encoding settings like this would still be beneficial.



> I wonder about setCompressLevel, seeing it's currently used to set the zlib and tight encoder levels by the client. The clients have a default for that, so it's always sent, no? In that case, the new logic would always automatically override the client, not sure if that's desired without an explicit server param.

Sorry. I meant `Encoder::setQualityLevel()`. Although at some point automatic `setCompressLevel()` would also be nice to avoid CPU churn. :)

I don't see a problem with overriding clients as long as we can make sure the code makes good decisions. Manually configuring this magic is a cop-out and I would prefer it to go away long term.

> Do you have any ideas on the bw measurement fallback? There don't seem to be timestamps or such that could be used to see if a packet was delayed.

Not really, no. I decided to avoid that mess since it wasn't strictly need to get decent behaviour from lossless refresh. The lack of feedback is indeed the main obstacle for the fallback. I assume we'll have to do something creative based on measure how quickly we are able to send things.

One thing to note is that we refuse to use continuous updates if we cannot measure things properly. So any client needing the fallback method will be forced to keep sending update requests. Perhaps that can allow us to do something?

mmcclaskey:
>The original request left the impression he wanted to leave bandwidth
for other things too, in which case automatic measurement by Xvnc
wouldn't be suitable. @mmcclaskey can you comment?

You are correct, I am looking for scalability through the network. So we don't want clients to use all the available bandwidth as they will collectively consume everything that is available. I tried setting the -FrameRate option in Xvnc to 24fps. I found this in the man documentation, is this value respected? If so, I would want the algorithm to use the max frame rate when deciding when to start degrading the jpeg quality. If this threshold was set to 30fps, it would never kick in. Ideally, it would be nice if that threshold to degrade was automatically selected but overridable. 

Great work by the way, this is looking very promising. One other thing I did was set the max resolution 1280 width then on the client side scale the image to full screen. The jpeg quality has to be set to 8 to avoid blowing up jpeg compression artifacts. I had to adjust the display DPI and other display settings so things look appropriate. This seems to make a good difference, especially for users with high resolutions.

clbr:
> I tried setting the -FrameRate option in Xvnc to 24fps. I found this in the man documentation, is this value respected? If so, I would want the algorithm to use the max frame rate when deciding when to start degrading the jpeg quality. If this threshold was set to 30fps, it would never kick in. Ideally, it would be nice if that threshold to degrade was automatically selected but overridable. 

It is respected by the server, limiting the updates to that fps.
However it also limits the change tracking to that fps, making the
server unable to detect higher fps changes being over that fps (without
changes that would increase server cpu usage). How would this best
interact with the scaling?

Ie, for -FrameRate 24, should it drop to minimum for 24fps, half for
12fps, and so on? (assuming the minimum fps to degrade at, ie maximum
quality threshold, was set at zero)

mmcclaskey:
I think there are two things here. What is a good default behavior for tigervnc and what is a good configuration specific to my needs. I don't think I can answer the first question, but my guess is that you would want to select a default threshold fps (greater than this jpeg quality drops) that does not cause visually noticeable artifacts. For my specific use case, I would probably drop that threshold value a bit to help drop BW utilization at the cost of visual artifacts during high rates of change.

clbr:
Updated patches pushed. The code is moved to EncodeManager, enabled by default, and with a rudimentary bandwidth check. I added a preset parameter, PreferBandwidth, that addresses the "black magic" somewhat: it sets a few defaults such that Xvnc prefers to save bandwidth over quality.

I also noticed some low-hanging fruit with the lossless refresh option. The two highest JPEG levels can be treated as good as lossless when preferring bandwidth; with the defaults, quality is preferred and lossless refreshes happen as before.

clbr:
Oh, no attempt on the fallback b/w estimation, I have no good ideas for that.

mmcclaskey:
I'm trying to build this commit. I've got a docker container that sets up the build environment, however, it fails. The documentation I've found out there on building tigervnc doesn't cover all the dependencies needed, or perhaps it does but not explicitly. So through trial and error I've got the dockerfile with all the needed ubuntu packages. I get the following error when doing a "make servertarball"

-- Installing: /tmp/tigervnc-build.NCurOK/inst/usr/local/share/icons/hicolor/48x48/apps/tigervnc.png
-- Installing: /tmp/tigervnc-build.NCurOK/inst/usr/local/share/icons/hicolor/scalable/apps/tigervnc.svg
install: cannot stat './xorg.build/bin/Xvnc': No such file or directory
CMakeFiles/servertarball.dir/build.make:59: recipe for target 'CMakeFiles/servertarball' failed
make[3]: *** [CMakeFiles/servertarball] Error 1
CMakeFiles/Makefile2:103: recipe for target 'CMakeFiles/servertarball.dir/all' failed
make[2]: *** [CMakeFiles/servertarball.dir/all] Error 2
CMakeFiles/Makefile2:110: recipe for target 'CMakeFiles/servertarball.dir/rule' failed
make[1]: *** [CMakeFiles/servertarball.dir/rule] Error 2
Makefile:176: recipe for target 'servertarball' failed
make: *** [servertarball] Error 2

If I just try make tarball (just the viewer) I get the following error.

-- Installing: /tmp/tigervnc-build.64oip7/inst/usr/local/share/icons/hicolor/scalable/apps/tigervnc.svg
release/maketarball: 55: release/maketarball: pushd: not found
CMakeFiles/tarball.dir/build.make:59: recipe for target 'CMakeFiles/tarball' failed
make[3]: *** [CMakeFiles/tarball] Error 127

------- Dockerfile --------

FROM ubuntu:16.04

RUN apt-get update
RUN apt-get install -y gettext libjpeg-turbo8 libjpeg-turbo8-dev libfltk1.3 openjdk-9-jre make gcc git
RUN apt-get install -y autoconf cmake build-essential libx11-dev zlib1g-dev libfltk1.3-dev libgnutls-dev
RUN apt-get install -y xorg-dev

clbr:
I believe the servertarball target doesn't actually build Xvnc, just packs it up if it exists. The BUILDING.txt file has examples on how to build Xvnc. As for the pushd error, that's a bash command, so I guess you need to install bash, I think Ubuntu defaults to dash nowadays?

common/rfb/EncodeManager.cxx
```diff
@@ -997,3 +1030,125 @@ void EncodeManager::OffsetPixelBuffer::update(const PixelFormat& pf,
...
+  // The tracker gives quality as 0-128. Convert to our desired range
+  dynamic *= dynamicQualityOff;
+  dynamic += 64; // Rounding
+  dynamic /= 128;
+  dynamic += dynamicQualityMin;
+
+  // Bandwidth adjustment
+  if (!Server::preferBandwidth) {
+    // Prefer quality, if there's bandwidth available, don't go below 7
+    if (curMaxUpdateSize > 2000 && dynamic < 7)
```
CendioOssman: These are still a bit arbitrary, so not quite what I was after when I said to look at the bandwidth. Basically the system should make a guess on what `dynamicQualifyOff` and `dynamicQualityMin` should be set to based on:

 * Available bandwidth
 * How large this update is

It knows how many pixels it needs to push, and it knows how many bytes it has for that. That gives the system a target compression ratio. What it doesn't know is the actual compression ratio it will get. For that we can start with some hard coded values and solve that issue later. Look in the JPEG encoder for some previous numbers.

common/rfb/EncodeManager.h
```diff
@@ -126,11 +138,16 @@ namespace rfb {
     };
     typedef std::vector< std::vector<struct EncoderStats> > StatsVector;
 
+    std::list<QualityInfo*> qualityList;
```
CendioOssman: Probably better to avoid pointers here. It's just a simple struct after all.

common/rfb/ServerCore.cxx
```diff
@@ -101,3 +101,21 @@ rfb::BoolParameter rfb::Server::queryConnect
...
+("DynamicQualityMax",
+ "The maximum dynamic JPEG quality, 0 = low, 9 = high",
+ 8, 0, 9);
+
+static void bandwidthPreset() {
+  rfb::Server::dynamicQualityMin.setParam(2);
+  rfb::Server::dynamicQualityMax.setParam(9);
+}
+
+rfb::PresetParameter rfb::Server::preferBandwidth
```
CendioOssman: I don't think we need a new parameter class for this. Just let it be a boolean and handle the logic in the encoder manager.

common/rfb/ServerCore.cxx
```diff
@@ -101,3 +101,21 @@ rfb::BoolParameter rfb::Server::queryConnect
 ("QueryConnect",
  "Prompt the local user to accept or reject incoming connections.",
  false);
+rfb::IntParameter rfb::Server::dynamicQualityMin
```
CendioOssman: You can put these in `EncodeManager.cxx`. The ServerCore stuff is more things that are used in various places. These new settings are very specific to the encoder.

common/rfb/EncodeManager.cxx
```diff
@@ -550,7 +550,7 @@ Encoder *EncodeManager::startRect(const Rect& rect, int type)
   encoder = encoders[klass];
   conn->writer()->startRect(rect, encoder->encoding);
 
-  if (encoder->flags & EncoderLossy)
+  if (encoder->flags & EncoderLossy && !encoder->treatLossless())
```
CendioOssman: This is an entirely different feature, so let's keep that separate. This PR will get complex enough as it is. :)

common/rfb/EncodeManager.h
```diff
@@ -57,7 +57,7 @@ namespace rfb {
 
     void writeUpdate(const UpdateInfo& ui, const PixelBuffer* pb,
                      const RenderedCursor* renderedCursor,
-                     size_t maxUpdateSize);
+                     size_t maxUpdateSize = 2000);
```
CendioOssman: Update the tests instead. We don't want to overlook this parameter as it will be crucial to get right.

common/rfb/EncodeManager.cxx
```diff
@@ -77,6 +82,12 @@ struct RectInfo {
   Palette palette;
 };
 
+struct QualityInfo {
+  struct timeval lastUpdate;
+  Rect rect;
```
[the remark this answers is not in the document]
mmcclaskey: I requested individual rect because I was afraid that a small GIF animation on a webpage would result in noticeably bad quality in static neighboring portions.

clbr:
> Maybe we should put that logic in the encoder in that case. It seems odd to add that here.

Then it needs to be able to tell who set it. Ok to change to 
setQualityLevel(int level, bool overridesClient = false)?

> Why not regions? Do we really need to track each individual rect?

It simplifies the logic to think of each component as a rect, instead
of a complex region. Typically animated banners don't change their
entire area in a frame, but in pieces. The logic tries to unify nearby
rects such that one ad banner would become one rect, even if it never
changes the entire area at once.

This then lets the entire ad get the quality it deserves, which is
lower than if a small part was considered on its own.

> It knows how many pixels it needs to push, and it knows how many bytes it has for that. That gives the system a target compression ratio. What it doesn't know is the actual compression ratio it will get. For that we can start with some hard coded values and solve that issue later. Look in the JPEG encoder for some previous numbers.

Yep the JPEG encoder has comments on the ratios each quality level gets.
Those could be put into a table in the encoder.

> Update the tests instead. We don't want to overlook this parameter as it will be crucial to get right.

Should they hardcode a high value, for high bandwidth, or?

CendioOssman:
Urgh. Mailing response to a review apparently did not integrate well. Your responses did not end up in the correct threads. :/

> > Maybe we should put that logic in the encoder in that case. It seems odd to add that here.
> Then it needs to be able to tell who set it. Ok to change to
> setQualityLevel(int level, bool overridesClient = false)?

Just let whatever was called last be the one that takes effect.

> > Why not regions? Do we really need to track each individual rect?
> It simplifies the logic to think of each component as a rect, instead
> of a complex region. Typically animated banners don't change their
> entire area in a frame, but in pieces. The logic tries to unify nearby
> rects such that one ad banner would become one rect, even if it never
> changes the entire area at once.
> 
> This then lets the entire ad get the quality it deserves, which is
> lower than if a small part was considered on its own.

I'd say it makes things more complex as all other change tracking is region based. :/

Is it an actual problem that we need all this detail though?

> > Update the tests instead. We don't want to overlook this parameter as it will be crucial to get right.
> Should they hardcode a high value, for high bandwidth, or?

Just pick something for now and add a "FIXME" comment there. We can deal with how the tests are affected by this later.

CendioOssman:
I'm thinking that we're probably trying to do too many things at once here. So let's take a step back and see if we can break up this problem a bit.

Let's start with trying to solve the issue of picking the right quality setting. Once we have that working we can start looking at the change rate tracking. That way we can focus on one thing at a time.

IOW, just let the decision be based on the current bandwidth and current update size. Ignore any history for now. We can add more complex logic later.

clbr:
> Let's start with trying to solve the issue of picking the right quality setting. Once we have that working we can start looking at the change rate tracking. That way we can focus on one thing at a time.
> 
> IOW, just let the decision be based on the current bandwidth and current update size. Ignore any history for now. We can add more complex logic later.

The thing is, neither I nor the original reporter are interested in
that case. None of the VNC clients I'm testing with even support the
continuous updates and fences, so none of them get the bandwidth
reporting. IOW I don't want to spend a lot of time on that, especially
inventing a fallback for those clients could take very long.

I don't mind going for a few iterations. The current change tracking
works decently in everything I tested. The changes requested are mostly
stylistic. If we move the "treat lossless" to a separate request, do
you see big issues in this design?

CendioOssman:
Unfortunately I will not accept something that requires users to tweak settings. Things need to get simpler for users, not more complex. Heuristics are fine, but they cannot make too large assumptions about available resources and use cases. The system needs to adapt or we're not making any real progress from where we currently are.

clbr:
That is the reason for the "PreferBandwidth" parameter. It's a single
knob, no magic numbers. The original request was about the use case
where there's plenty of bandwidth - no congestion is expected ever.
Supporting that use case better does not make things more complex for
the "prefer quality" use case.

CendioOssman:
If bandwidth wasn't an issue then we wouldn't need any changes to the quality at all. And the users without "lots" will still have very varying needs. So some kind of sliding scale is needed. And network conditions aren't constant so the setting needs to change as conditions change.

Which is why I'm saying this system needs to be adaptive.

I'm afraid we don't want features which we cannot provide to all users, including the inexperienced ones.

clbr:
In the use case, you want VNC to take less bandwidth even when there is
free bandwidth available. That is something that Xvnc cannot detect,
and so there must be a setting by the user. In this use case, the
bandwidth scaling code would always be a no-op.

I agree that having quality scaling by default, in the "prefer quality"
case would be nice, and I can add that for the recent clients; that's
not the issue here. It's the time required for the fallback, which
would make this no longer worth it for me. So to be clear, for quality
scaling to be merged for the original use case, is it enough to scale
automatically for new clients, or do you require automatic scaling for
old clients too?

CendioOssman:
Only enabling the feature when we have reliable bandwidth measurement would be alright, yes.

But if we don't then we should keep the existing behaviour (follow the client request). Which I got the impression that wasn't good enough for your usage?

clbr:
Only when the server is configured to do so, it should ignore the client's request for a constant JPEG quality, no matter if the client was old or new. This would be entirely separate from the default bw scaling that you requested.

clbr:
Are we speaking past each other? The use case I'm after cannot be detected automatically. It therefore always requires a config parameter. Therefore no automatic system is good enough.

CendioOssman:
Possibly, but I don't think so.

As far as I understand you, your use case is "I want TigerVNC to automatically pick the most appropriate JPEG quality for my conditions". That should be possible without any settings.

You also seem to have a second use case of "I want to restrict how much bandwidth TigerVNC consumes because I want to reserve some for other things". That would indeed require some feedback from the user. But that feature is entirely independent from the quality system and should be handled separately.

clbr:
It's not really independent, since it directly conflicts with it. Goal: "I want to restrict how much bandwidth TigerVNC consumes because I want to reserve some for other things". The bandwidth system you wanted by default: "Scale quality by bandwidth - it should act as both minimum and maximum. Don't go above this quality if we don't have the bw for it; don't go below this quality if we do have the bw".

We want it to go lower even though bw is available. So it conflicts, and so I want to disable the bw minimum under the option for my use case. But I understand you don't want it to ever be disable-able. If this is so, then the work cannot be merged in any case.

CendioOssman:
Overriding the quality selection logic is a very bad way of limiting bandwidth as it will not be reliable. If you want bandwidth restriction, then code specifically that and don't try to do it indirectly. With that in place the quality logic will adapt to the artificial restriction just as it would to a real bandwidth loss.

PR #681, opened 2018-07-18
Replace all README.txt with README.rst

metalefty:


CendioOssman:
Thank you. But these files are unfortunately handled by the translation project. This will be adjusted the next time the translations are updated by them.

PR #694, opened 2018-08-08
Check endianness when constructing PlatformPixelBuffer()

grulja:
This fixes wrong colors when using vncviewer on ppc64 or s390x (big endian). We shouldn't assume we use little-endian all the time.

### DIFF
--- a/vncviewer/PlatformPixelBuffer.cxx
+++ b/vncviewer/PlatformPixelBuffer.cxx
@@ -34,8 +34,8 @@
 static rfb::LogWriter vlog("PlatformPixelBuffer");
 
 PlatformPixelBuffer::PlatformPixelBuffer(int width, int height) :
-  FullFramePixelBuffer(rfb::PixelFormat(32, 24, false, true,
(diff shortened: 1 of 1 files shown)


wyatt8740:
After updating my packages in Debian, I am also experiencing this issue on PowerPC (32-bit, PowerBook G4).
The date of modification in the file index on this page indicates that the bug was introduced no later than July 30, when the last build was made for PPC 32-bit in Debian Ports:
http://ftp.ports.debian.org/debian-ports/pool-powerpc/main/t/tigervnc/

wyatt8740:
OK, reporting back:
It mostly works. There's still one problem spot, which is where the popup saying "Press %s to enter fullscreen mode" (in my case, %s being F8) is blue tinted on account of some unfixed endian problem which still exists with the overlay.

CendioOssman:
Thanks. Looks correct. This must have been introduced already in 1.8.

CendioOssman:
@wyatt8740, sounds like you have some other issue. Please file a new issue and include a screenshot.

PR #700, opened 2018-08-15
Increase maximum vnc server name length, fix #699

### LINKED ISSUE #699: Maximum vnc server name length in vncviewer is too small
Currently ``vncviewer`` limits maximum server name to 64 symbols which is too short for my use case and it is shorter than maximum fqdn length which is effectively 253 bytes as described in this stackoverflow answer: https://stackoverflow.com/questions/8724954/what-is-the-maximum-number-of-characters-for-a-host-name-in-unix.

daa:
Current maximum vnc server name length is too small as described in corresponding issue, thus increased it to 256 symbols.


daa:
Thank you!

PR #703, opened 2018-08-20
Release queueMutex using finally blocks

pepijnve:
Ensure queueMutex is always correctly released by using finally blocks. This is the closest approximation of AutoMutex style automatic release you can get in Java.


bphinz:
Looks good.  Thanks!

PR #728, opened 2018-09-26
Avoid potential crash when replacing buffer in PlainPasswd

grulja:
This is reproducible when using `export MALLOC_CHECK_=2 MALLOC_PERTURB_=204` and running vncpasswd `echo "123456" | vncpasswd -f >/tmp/out`. In vncpassd we use fgets() to set buffer in CharArray, but using vncpasswd as mentioned before, second run of fgets won't read anything, setting some unknown string to the buffer and then when PlainPasswd destructor is called, we attempt to replace the buffer with zeros, which might fail on getting buffer length. In this case we can use length we get from the contructor. Also add some additional checks to avoid deleting empty buffer.

### DIFF
--- a/common/rfb/util.h
+++ b/common/rfb/util.h
@@ -50,7 +50,7 @@ namespace rfb {
     CharArray() : buf(0) {}
     CharArray(char* str) : buf(str) {} // note: assumes ownership
     CharArray(int len) {
-      buf = new char[len];
+      buf = new char[len]();
     }
     ~CharArray() {
       delete [] buf;


CendioOssman:
I can't say I like the approach of adding that `length` attribute. It won't be properly updated when `replaceBuf()` is called. It is also not used in the class that defines it.

I'd say the bug here is that `PlainPasswd` expects C strings and giving it something else is a bug that will crash (much like using `fputs()` on something that isn't null terminated). We should fix the callers, and see if there are ways of improving the API so mistakes are more difficult to make.

grulja:
I now correctly read the documentation of fgets and spotted that the string remains unchanged if it fails to read anything so the easiest way how to avoid the crash is to fill the string with zeroes.

CendioOssman:
Great. However the `if` statements are not needed. It is perfectly safe to delete a null pointer.

PR #729, opened 2018-09-26
Fix coverity scan issues

grulja:


unix/tx/TXWindow.cxx
```diff
@@ -217,6 +217,10 @@ void TXWindow::getColours(Display* dpy, XColor* cols, int nCols)
       XFreeColors(dpy, cmap, &p, 1, 0);
     }
   }
+
+  delete [] got;
+  delete [] shared;
+  delete [] usedAsNearest;
```
CendioOssman: Compilers support dynamic arrays these days, so I would prefer a change to that. E.g. `bool got[nCols];`.

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -165,24 +165,40 @@ int main(int argc, char** argv)
     FILE* fp = fopen(fname,"w");
     if (!fp) {
       fprintf(stderr,"Couldn't open %s for writing\n",fname);
+      delete [] fname;
```
CendioOssman: `fname` might point to a part of `argv`.

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -165,24 +165,40 @@ int main(int argc, char** argv)
...
+      if (obfuscatedReadOnly)
+        delete obfuscatedReadOnly;
       exit(1);
     }
 
     if (obfuscatedReadOnly) {
       if (fwrite(obfuscatedReadOnly->buf, obfuscatedReadOnly->length, 1, fp) != 1) {
         fprintf(stderr,"Writing to %s failed\n",fname);
+        delete [] fname;
+        delete obfuscated;
```
CendioOssman: Delete these as soon as they are no longer need to avoid having to repeat this constantly.

unix/xserver/hw/vnc/xvnc.c
```diff
@@ -766,10 +766,11 @@ vfbUninstallColormap(ColormapPtr pmap)
 	    curpmap = (ColormapPtr) LookupIDByType(pmap->pScreen->defColormap,
 						   RT_COLORMAP);
 #else
-	    dixLookupResourceByType((void * *) &curpmap, pmap->pScreen->defColormap,
-				    RT_COLORMAP, serverClient, DixUnknownAccess);
+	    int rc =  dixLookupResourceByType((void * *) &curpmap, pmap->pScreen->defColormap,
+					      RT_COLORMAP, serverClient, DixUnknownAccess);
 #endif
-	    (*pmap->pScreen->InstallColormap)(curpmap);
+	    if (rc == Success)
```
CendioOssman: `rc` will be missing in the other `#if` code path. Better to check it early and return if it is a failure (and log).

vncviewer/ServerDialog.cxx
```diff
@@ -158,6 +158,7 @@ void ServerDialog::handleLoad(Fl_Widget *widget, void *data)
     fl_alert("%s", e.str());
   }
 
+  delete [] filename;
```
CendioOssman: Why do we even have a `strdup()` here?

unix/vncpasswd/vncpasswd.cxx
grulja: fname is used on the debug output, cannot be deleted earlier.

vncviewer/ServerDialog.cxx
```diff
@@ -213,6 +213,7 @@ void ServerDialog::handleSaveAs(Fl_Widget *widget, void *data)
     fl_alert("%s", e.str());
   }
   
+  delete [] servername;
```
CendioOssman: Can't this also be solved by just getting rid of the strdup()? And if not, `free()` needs to used, not `delete []`.

vncviewer/ServerDialog.cxx
```diff
@@ -244,4 +245,6 @@ void ServerDialog::handleConnect(Fl_Widget *widget, void *data)
   } catch (rfb::Exception& e) {
     fl_alert("%s", e.str());
   }
+
+  delete [] servername;
```
CendioOssman: Dito.

unix/xserver/hw/vnc/vncExtInit.cc
```diff
@@ -184,15 +184,15 @@ void vncExtensionInit(void)
             listeners.push_back(new network::TcpListener(vncInetdSock));
             vlog.info("inetd wait");
           }
-        } else if (rfbunixpath.getValueStr()[0] != '\0') {
+        } else if (rfbUnixPath[0] != '\0') {
```
CendioOssman: In these cases you can use `operator const char*()` which gives a reference, so no need to be freed.

vncviewer/DesktopWindow.cxx
```diff
@@ -103,12 +105,13 @@ DesktopWindow::DesktopWindow(int w, int h, const char *name,
...
     int matched;
-    matched = sscanf(geometry.getValueStr(), "+%d+%d", &geom_x, &geom_y);
+    char *geometryStr = geometry.getValueStr();
+    matched = sscanf(geometryStr, "+%d+%d", &geom_x, &geom_y);
     if (matched == 2) {
       force_position(1);
     } else {
       int geom_w, geom_h;
-      matched = sscanf(geometry.getValueStr(), "%dx%d+%d+%d", &geom_w, &geom_h, &geom_x, &geom_y);
+      matched = sscanf(geometryStr, "%dx%d+%d+%d", &geom_w, &geom_h, &geom_x, &geom_y);
```
CendioOssman: Same thing should be possible here.

vncviewer/OptionsDialog.cxx
```diff
@@ -282,7 +283,8 @@ void OptionsDialog::loadOptions(void)
   /* Screen */
   int width, height;
 
-  if (sscanf(desktopSize.getValueStr(), "%dx%d", &width, &height) != 2) {
+  char *desktopSizeStr = desktopSize.getValueStr();
+  if (sscanf(desktopSizeStr, "%dx%d", &width, &height) != 2) {
```
CendioOssman: And here.

unix/tx/TXWindow.cxx
```diff
@@ -132,7 +132,7 @@ TXGlobalEventHandler* TXWindow::setGlobalEventHandler(TXGlobalEventHandler* h)
 
 void TXWindow::getColours(Display* dpy, XColor* cols, int nCols)
 {
-  bool* got = new bool[nCols];
+  bool got[nCols];
```
danilaml: this is not valid C++

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -165,24 +165,40 @@ int main(int argc, char** argv)
     FILE* fp = fopen(fname,"w");
     if (!fp) {
       fprintf(stderr,"Couldn't open %s for writing\n",fname);
+      delete [] fname;
+      delete obfuscated;
+      if (obfuscatedReadOnly)
```
danilaml: delete on nullptr does nothing (same as free) so this check should be unnecessary.

unix/tx/TXWindow.cxx
grulja: Can you share what's wrong with it?

danilaml: VLAs (arrays with dynamic size allocated on stack) are not part of C++ standard and were even made optional in the recent C standard. They can also lead to issues if the size passed is not sanitized/can be big (you can easily overflow stack).

I'm not sure what specific requirements this project has (was just lurking), but if you don't want to deal with with manual new/delete (it looks atrocious here and already creates plenty of problem), the best solution is to just use `std::vector<...>`.

unix/vncpasswd/vncpasswd.cxx
danilaml: This is very wrong (UB). fname might be allocated by` strdup` so it mast be freed by `free` in that case, not `delete []`. But it looks like it's also possible to be allocated by `new[]`. This code is some ugly C/C++ mix ;P There are probably issues like this all over the place.

unix/tx/TXWindow.cxx
CendioOssman: Crap, forgot C++ doesn't have that. I agree that a `std::vector` is probably appropriate here.

CendioOssman:
@grulja , do you think you could have a look at these last few things and we can merge this?

unix/vncpasswd/vncpasswd.cxx
grulja: It uses new[] all the time, because it uses internally implemented strdup(), which uses new[] as well, so calling delete[] should be just fine.

grulja:
Should be fixed now.

unix/vncpasswd/vncpasswd.cxx
danilaml: Wow, that's totally not confusing at all. Are you sure about that? Can you point me to this internal implementation, just to be safe (I've found some other similar functions like `strDup`, but no `strdup`)?

grulja: You are right, I though both are strdup, in that case maybe strDup should be used instead?

danilaml: Looks like it would be the least invasive change (though it wouldn't make current code any better, just more correct).

grulja:
If you want, I can squash all the commits into one, if you don't want all the commits in history.

CendioOssman:
Looks good. And yes, please clean up the history so we avoid the back-and-forth.

grulja:
> Looks good. And yes, please clean up the history so we avoid the back-and-forth.

Done. Squashed into 4 unrelated commits.

PR #730, opened 2018-09-28
first speed bumps in porting to OpenBSD

hildred:
Since the only vnc server available for OpenBSD is x11vnc (tightvnc was replaced with ssvnc which is viewer only) and running it headless requires xvfb, I decided I could do better and started a port. I have not yet succeed, but here are some issues I found along the way.

contrib/xorg requires python two.
contrib/xorg tries to download a broken link.
contrib/xorg requires bash.

three piddley fixes. might help someone else.


CendioOssman:
Ping @hildred

PR #736, opened 2018-10-09
Remove Java web server

CendioOssman:
Applets don't work anymore so remove everything that has to do with serving them.


bphinz:
Applets are already unsupported in browsers and Oracle recently deprecated Java Web Start, so I'm fine with removing support for the embedded HTTP server and applet mode altogether from the Java viewer, but I'd rather not kill the Java viewer altogether at this point so I'd prefer to keep the client itself (which can be run standalone as well as deployed via JWS) as part of the RPMs.

CendioOssman:
Right, the suggestion is just to nuke the applet portions and keep it as standalone.

Got any suggestions for better packaging there though? We shouldn't hide it in `/usr/share` if it's supposed to be directly used. Can we put it in `/usr/bin` or is some form of wrapper needed?

And should it have an associated `.desktop` file?

bphinz:
Agreed.  Let me look and see what the RPM packaging guideline say about
java client apps.

bphinz:
From https://fedoraproject.org/wiki/Packaging:Java

🔗 <https://fedoraproject.org/wiki/Packaging:Java#Installation_directory>
Installation directory

   - All architecture-independent JAR files MUST go into %{_javadir} or its
   subdirectory.

%{__javadir} usually ends up being /usr/share/java

CendioOssman:
Can you work up a suggestion based on that? They seem to have a section about wrapper scripts as well.

bphinz:
I'm not sure what to do about the wrapper.  I don't want it to be ambiguous
such that the name confuses people about which viewer they are running.
Thoughts?

CendioOssman:
Perhaps just add "java" somewhere in the name? E.g. `vncviewer_java`?

CendioOssman:
Any luck? Perhaps I can merge this and then you can add the viewer as a normal application later?

PR #745, opened 2018-10-17
Define "TigerVNC Viewer" in header, to get a consistent spelling

astrand:
It turns out that we were using all four variants:

VNC Viewer
VNC viewer
TigerVNC viewer
TigerVNC Viewer

Also define "vncviewer" in the same way.


astrand:
Apparently, xgettext has some troubles with extracting strings from preprocessor defines. Needs to be reworked.

CendioOssman:
This looks like overkill. I don't think we mess this up that often. :)
I suggest just fixing the stuff we've overlooked and then be happy at that.

Issue #528, opened 2017-10-24
x0vncserver does not generate fake shifts etc


astrand:
Assuming classic keyboard handling (not Raw/Qemu):
The keyboard implementation in x0vncserver is very basic and does not handle all cases. Example: 

- Swedish keyboard on client
- English keyboard on x0vncserver Xserver
- User presses "Shift+7" = "/" on client

In this case, no symbol will be generated. This is because there's no shifted slash keysym in the server. x0vncserver should generate a fake un-shift; just like Xvnc does.

Also note that x0vncserver should have much better logging. Even with loglevel 100, the example above only gives:

 VNCSConnST:  Key pressed: 0xffe2 / 0x36
 VNCSConnST:  Key pressed: 0x2f / 0x8
 VNCSConnST:  Key released: 0x2f / 0x8
 VNCSConnST:  Key released: 0xffe2 / 0x36

...thus no information about the fact that XkbKeysymToKeycode() returned zero.

daimpi:
I have a similar problem, connecting from Windows 10 via vncviewer 1.8 to Ubuntu 17.10 running x0vncserver. I'm using a German keyboard on my Windows machine and even when setting the keyboard to German on the remote machine, i still cannot generate anything using the AltGr Key (e.g. not possible to generate an `@` via AltGr+q and even the alternative via ctrl+alt+q doesn't work). A friend of mine who is running ArchLinux on his machine (also German Keyboard) doesn't have this problem.

CendioOssman:
Similar, yes, but a big part of your problem is that AltGr is a bit broken on Windows and we have yet to find a good workaround. Hence why it works better on a Linux client.

daimpi:
Ah interesting… didn't know that there was this problem with Windows in general. Thanks for the info :-).

PR #749, opened 2018-10-23
Use RawKeyboard by default

astrand:
The keyboard implementation in x0vncserver is very basic and has many
issues, for example #528.


CendioOssman:
Closing until we have checked with the community if we should change the default.

PR #751, opened 2018-10-23
Add option -StartCommand, which can be used to start X server

astrand:
Corresponds to the x11vnc "-create" functionality.

### DIFF
--- a/unix/x0vncserver/x0vncserver.cxx
+++ b/unix/x0vncserver/x0vncserver.cxx
@@ -22,7 +22,9 @@
 
 #include <strings.h>
 #include <sys/types.h>
+#include <sys/wait.h>
 #include <sys/stat.h>
+#include <fcntl.h>
 #include <unistd.h>
 #include <errno.h>
 #include <rfb/Logger_stdio.h>
@@ -66,6 +68,8 @@ IntParameter queryConnectTimeout("QueryConnectTimeout",
                                  "rejecting the connection",
                                  10);
 StringParameter hostsFile("HostsFile", "File with IP access control rules", "");
+StringParameter startCommand("StartCommand", "If connection to X display fails, start a new X server "
+			     "using the specified command", "");
 
 //
 // Allow the main loop terminate itself gracefully on receiving a signal.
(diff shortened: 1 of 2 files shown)


CendioOssman:
I am sceptical about this. We're having a lot of issues regarding starting sessions already in `vncserver`, so I'm not thrilled about more on the same theme. And unlike `x11vnc` we have `Xvnc`, so we already handle the headless case in a much more sane way. Isn't most users of this already covered by simply putting `start-command; sleep 10; x0vncserver` in whatever startup procedures they have?

astrand:
> I am sceptical about this. We're having a lot of issues regarding starting sessions already in `vncserver`, so I'm not thrilled about more on the same theme. 

Me neither, but the difference is that "vncserver" actually contains a lot of commands to try & run. This obviously creates a maintenance burden. This patch, on the other hand, just provides a mechanism.  

>And unlike `x11vnc` we have `Xvnc`, so we already handle the headless case in a much more sane way. 

For some cases such as VDI with GPU acceleration, Xvnc is of no use. Users needs to run the real Xorg.

>Isn't most users of this already covered by simply putting `start-command; sleep 10; x0vncserver` in whatever startup procedures they have?

This is "possible", but with drawbacks:

- A Xserver will be started even if one is running. This causes ugly error messages at best.

- Users will always need to wait 10 seconds when connecting. Even if server is already running, fails to start etc. 10 seconds might also be bit low. In the pull request, the timeout is 20 seconds. Having to wait 10-20 seconds every time is likely not acceptable. 

- stdin/stdout/stderr will not be set to sane values and will not be line buffered. setsid() is not used, so in total, it will not possible to terminate x0vncserver and any SSH connections and let the Xserver continue to run.

CendioOssman:
If there are details with redirections and `setsid()` that need to be dealt with then it doesn't sound perfectly trivial this either. :)

I would very much prefer to see this implemented as something separate in the `contrib/` directory. I am not convinced we need to support yet another way to start sessions.

(We also have libvnc.so for when you have control of the startup of the X server, which also gives more functionality than x0vncserver)

PR #764, opened 2018-11-14
Native look and feel for the viewer

gui9394:
Modification in the setLookAndFeel method, so that it can use the native appearance, if the system does not have a native appearance configured it uses a multiplatform available, as this one described in the Swing documentation

https://docs.oracle.com/javase/7/docs/api/javax/swing/UIManager.html#getSystemLookAndFeelClassName()


bphinz:
The reason that I override the default appearance in favor of Nimbus is that there are so many bugs that are specific to each platform's native look and feel that it makes the entire UI-related codebase unmanageable.  If you want the native LAF you can set it with -Dswing.defaultlaf and bypass the application's preference for Nimbus.

PR #765, opened 2018-11-18
Use cmake TIMESTAMP function

bmwiedemann:
Use cmake TIMESTAMP function 
because it is not only platform independent
but also allows to override the build date
This helps to make the java timestamp file in tigervnc  builds reproducible.

See https://reproducible-builds.org/ for why this is good

Also adds UTC flag, to be independent of timezone.
Also changes JAVA_DATE format to ISO-8601 date format.

Requires cmake-2.8.11+ from 2013

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 # Setup
 #
 
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 2.8.11)
 if(POLICY CMP0022)
   cmake_policy(SET CMP0022 OLD)
 endif()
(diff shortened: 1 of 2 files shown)


bphinz:
My only complaint is having a different CMake minimum version than the top level build requirement.  Let me check the package builds for all of our supported distros and if they meet that already then I think you should adjust the PR to raise the overall minimum CMake version to 2.8.11.  I won't have access to the build system until I return from holiday at the end of the week.

bmwiedemann:
I had already updated the PR to update the main CMake version to 2.8.11 too. Thanks for checking in the build system.

bmwiedemann:
ping. did you have time to look into it yet?

bphinz:
Looks fine, all of the distros that we currently support supply cmake >= 2.8.12.  Thanks!

PR #774, opened 2018-12-09
Update RPM package dependencies to require version alignment between packages.

MarkMielke:
Without this - a broad mix of versions is possible, and can lead to broken deployments. With this, upgrading or downgrading one package will require the other packages to be upgraded or downgraded together.


CendioOssman:
Looks good to me, except for that release number going backwards.

@bphinz, what do you think?

bphinz:
Agreed.  It looks fine to me other than that

MarkMielke:
> Looks good to me, except for that release number going backwards.
> @bphinz, what do you think?

Check the changelog:

* Sun Dec 09 2018 Mark Mielke <mmielke@ciena.com> 1.9.80-3
- Update package dependencies to require version alignment between packages.
 * Sun Nov 26 2018 Brian P. Hinz <bphinz@users.sourceforge.net> 1.9.80-2
- Bumped Xorg version to 1.2.0

And the entry after in context:

* Mon Jun 20 2016 Brian P. Hinz <bphinz@users.sourceforge.net> 1.6.80-5
- Patch for Xorg 1.17 due to vendor bump of Xorg version

I believe Brian had reset the number in the changelog, but forgot to fix the Release at the top. I could break this correcting into a separate Pull Request, but I thought I'd slip it in and avoid making a big deal about it. :-)

bphinz:
Yes, please do.  Thanks!

CendioOssman:
> I believe Brian had reset the number in the changelog, but forgot to fix the Release at the top. I could break this correcting into a separate Pull Request, but I thought I'd slip it in and avoid making a big deal about it. :-)

That is probably how it happened, but the `5` release is already in the wild so we cannot roll that back now. We can reset it next time we change the version number, but unfortunately not now.

MarkMielke:
I don't mind changing it to 6... However, I'm not sure it's true that 5 is already in the wild. This was all done for "1.9.80", which is pre-release, correct? The prior release, which used 5, applied to 1.9.0 or earlier. I did review this before coming to the conclusion.

But, please let me know what you would like to do with this context and I'll do it.

PR #788, opened 2019-01-15
Build packages on Travis

CendioOssman:
Use docker containers on Travis in order to build the various packages we have.

@bphinz, does this look reasonable to you? It would give us quick feedback to problems. And if we want it should be possible to hook it up to do the nightly builds for us.


bphinz:
Looks awesome to me.  I've been thinking along the same lines (using docker) but the sticking point always seems to be OSX...

CendioOssman:
Travis claims to have macOS VMs now, with Brew support. So I figured I'd give it a go as well and see...

bphinz:
Interesting.  We'll have to figure out how to handle package signing before we can migrate the nightly & release builds, but hopefully it's as easy as just adding a signing step to the travis builds

CendioOssman:
I don't quite understand how versions are supposed to work with the Ubuntu builds. How do you normally build those, @bphinz?

bphinz:
I don't have access to the build server at the moment, but it's a two-step process where I create the src tarball first and then use that for the .debs.  It might well be that the version isn't templated well in the files found in the contrib dir...

CendioOssman:
Looking at your files you seem to have something that automatically adds an entry in the changelog. I've hacked together something similar in my scripts here now.

CendioOssman:
I'm seeing some odd problem with a patch failing to apply for the trusty build. @bphinz, could you have a look at your build and see if you can see where we differ?

CendioOssman:
Everything seems to work now, so please have a look @bphinz and then I'll merge it.

CendioOssman:
> Shouldn't fltk src be 1.3.4-2?

I just moved what was already there, but I suppose we can take the opportunity to upgrade.

> The only other thing I would add is that maybe you should explicitly require fltk-devel >= 1.3.4 in the el7 spec file since their fltk library wasn't bumped to 1.3.4 until 7.4.1708.
> 

Sure.

> Otherwise I think it looks fantastic. As usual you've taken something sloppy that I did and made it really professional.

I'd say the packages have been working very well so far. This is just a next step. :)

CendioOssman:
@bphinz, I'd like to also update to use Ubuntu 16.04 for the general builds, but there is some compatibility issue with newer Java:

https://travis-ci.org/TigerVNC/tigervnc/jobs/482794631

From what I can Google it seems you should avoid using that API. Is it something you can fix?

PR #795, opened 2019-02-13
Don't build xserver with -I$(includedir) as it breaks cross-compiling

chewi:
This is where to install headers to, not where to find headers to
build against. Toolchains should know where to locate their system
headers while non-system headers should be found using
pkg-config. Users with bizarre setups where the toolchain really
cannot find the system headers should set CPPFLAGS.


CendioOssman:
Thanks. Not sure why that got in originally.

Issue #93, opened 2014-12-18
Adding unknown keysym until keysym table is full and no further input possible


merkata:
I have a (Arch Linux) client box connecting via vncviewer to a (Arch Linux) server running (tigervnc) vncserver. On my client I use two languages (setxkbmap -layout us,bg (phonetic) -option grp:alt_shift_toggle) - switching and using both languages on the client is fine, both languages work on the server when in a terminal session, however when I connect to the VNC desktop, keys stop responding after a while. Relevant errors in ~/.vnc/hostname:1.log

```
...
Thu Dec 18 11:32:13 2014
Input:       Added unknown keysym 0x69 to keycode 221
Input:       Added unknown keysym 0x66 to keycode 219

Thu Dec 18 11:35:04 2014
Input:       Added unknown keysym 0x78 to keycode 217

Thu Dec 18 11:35:05 2014
Input:       Added unknown keysym 0x62 to keycode 202

Thu Dec 18 11:35:19 2014
Input:       Added unknown keysym 0x75 to keycode 197
...
```

Running xev shows some weird stuff - when pressing 'd' for instance

```
KeyRelease event, serial 38, synthetic NO, window 0x1200001,
    root 0x178, subw 0x1200002, time 175414596, (49,39), root:(50,64),
    state 0x2000, keycode 149 (keysym 0x64, d), same_screen YES,
    XKeysymToKeycode returns keycode: 40
    XLookupString gives 1 bytes: (64) "d"
    XFilterEvent returns: False
```

and xmodmap -pke would report that 'd' was added to 149

```
xmodmap -pke | grep 149
keycode 149 = d D d D
```

The table has room for 255 entries and they fill up really quickly - when it is full, I'm not able to use any 'unknown' keys that werent added to the table before it was full.

Locales on both machines

```
bg_BG.utf8
C
en_US.utf8
POSIX
```

tigervnc is at latest version (1.4.0-1 from official arch repository)

kozhin:
The same problem I have with CentOS 7. My locale is ru_RU.utf8.
Is there any way to reset the table when it's full?

CendioOssman:
Yes. Reloading a fresh keymap with setxkbmap will reset the table.

kozhin:
[...]

PR #804, opened 2019-03-11
xkb map exhaustion fix

ogurets:
Fix for issue #93. Tested a few weeks by me, no issues so far.
Tried testing Round Robin feature, turned out it wasn't such a good idea, as some of my local keys were conflicting with some existing exotic keys and therefore were not subjected to AddSym.

The simplest solution turned out to be the best one so far.

### DIFF
--- a/unix/xserver/hw/vnc/Input.c
+++ b/unix/xserver/hw/vnc/Input.c
@@ -533,14 +533,13 @@ static void vncKeysymKeyboardEvent(KeySym keysym, int down)
 
 	/* No matches. Will have to add a new entry... */
 	if (keycode == 0) {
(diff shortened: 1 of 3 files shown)


CendioOssman:
Thank you for working on this. Unfortunately this approach is not something we can accept. There are two problems here:

 * Blindly overwriting an existing key means we might mess up something that the user or system has set up.
 * By using one just key we'll have to spam constant keyboard updates as pretty much every key press results in a map change.

What we'd like to see if an improvement of the existing approach, i.e. only use "free" keys. The difficult problem to solve is to know if a used key is something we set up (and can reuse) or if it was something from the user or system.

PR #806, opened 2019-03-14
fix compiler warning FullFramePixelBuffer::stride’ is used uninitialized

ferkulat:
With commit 403ac27 parameter "int stride" got removed from constructor "PlatformPixelBuffer::PlatformPixelBuffer."
Before this parameter was passed to the constructor of "FullFramePixelBuffer".
"FullFramePixelBuffer" has also a member with the same name "stride". :(

Now this uninitialized FullFramePixelBuffer::stride gets passed to the
constructor FullFramePixelBuffer::FullFramePixelBuffer.

Now I understand the value of naming class members with a preceding "m_".


CendioOssman:
Sorry, I missed this whilst reading your comment. It got fixed on `master` in another way. :)

PR #807, opened 2019-03-15
fix valgrind memcheck warning about free/delete/delete[] mismatch

ferkulat:
new function std::string getvnchomedir() wraps around int getvnchomedir(char **dirp)
+ no output parameter -> result can be made const
+ no delete anymore (RAII)

### DIFF
--- a/common/os/os.cxx
+++ b/common/os/os.cxx
@@ -36,7 +36,7 @@
 #include <shlobj.h>
 #endif
 
-int getvnchomedir(char **dirp)
+static int getvnchomedir(char **dirp)
 {
 #ifndef WIN32
 	char *homedir, *dir;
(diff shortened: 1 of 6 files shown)


CendioOssman:
Thanks for finding this issue. However, we don't really use `std:string` in TigerVNC and it seems a bit overkill to start using it here just to fix this. Could you adjust this PR to just fix the two bad calls to `free()`? :)

ferkulat:
Yes I will change this.
But I would be really interested in finding a way to replace manual resource management by RAII.
Any other ideas you would accept?

CendioOssman:
The code uses very little RAII, so it would be a rather large change and I'm not sure it would be worth it. RAII can in many cases be difficult to achieve without making the code difficult to read. Feel free to play around with it, but I'm not sure what we'd accept. It would need to be an improvement beyond just for the sake of it.

CendioOssman:
If you want some nice coding challenges, then we have many bugs and feature requests that need work. Just check the issue list. :)

CendioOssman:
Thank you. Commited as fd3bfa6d9d1166b3fdb585c794abb4634d70409c.

PR #811, opened 2019-03-20
Fix issues caught by GCC 8

awilfox:
I believe these fixes are correct.  They fix errors found by GCC 8.  If they are not correct, I will be glad to amend them.


CendioOssman:
Thank you for reporting these issues. Unfortunately it wasn't quite the correct approach.

The allocation warning was because of a missing `throw`, where gcc correctly found a bad control flow.

The `strcpy()` fix was a good start, but we also need to explicitly terminate the strings.

I've pushed two commits to fix these issues to `master` now.

PR #813, opened 2019-04-01
Fix segmentation fault in SMsgWriter::writeNoDataUpdate()

amonott:
writeNoDataUpdate() calls writeNoDataRects(), which uses client->screenLayout()
in a call to writeExtendedDesktopSizeRect() without a check, if the layout
is empty. Unfortunately there is a race condition and there might not have
been a layout set before the first call, leading to a crash when blindly
using the layout. We have experienced several related crashes.

Fix: SMsgWriter::writeNoDataRects(): check that client layout has a screen before
trying to send encodings to it.

Signed-off-by: Amon Ott <a.ott@m-privacy.de>

### DIFF
--- a/common/rfb/SMsgWriter.cxx
+++ b/common/rfb/SMsgWriter.cxx
@@ -355,7 +355,7 @@ void SMsgWriter::writePseudoRects()
 
 void SMsgWriter::writeNoDataRects()
 {
-  if (!extendedDesktopSizeMsgs.empty()) {
+  if (!extendedDesktopSizeMsgs.empty() && client->screenLayout().num_screens() > 0) {
     if (client->supportsEncoding(pseudoEncodingExtendedDesktopSize)) {
       std::list<ExtendedDesktopSizeMsg>::const_iterator ri;
       for (ri = extendedDesktopSizeMsgs.begin();ri != extendedDesktopSizeMsgs.end();++ri) {


CendioOssman:
Thank you for your contribution. However I'm having some trouble understanding the scenario here.

First off, an empty layout should not be able to cause a crash. It should simply send an ExtendedDesktopSize rect without any monitors. Where did you see it crash? Do you have a backtrace of such a crash?

Secondly, there should not be a scenario where we try to send ExtendedDesktopSize without a layout. You mention a race, but have you figured out how this race occurs? We should not be sending any FramebufferUpdate without knowing the framebuffer dimensions. And if we know them then the layout should have been set.

amonott:
We repeatedly got such traces, if login was very fast, e.g. because of SSL certificate or Kerberos based login (our own extensions). The crash does not happen with password login. It is possible that some of our changes trigger an earlier sending of ExtendedDesktopSize. In any case, the potential crash should be fixed.

The for() loop in writeExtendedDesktopSizeRect() dereferences layout.begin() without check, which is NULL for an empty list.

/usr/bin/Xtightgatevnc(+0x27e465)[0x55e8efc89465]
/lib/x86_64-linux-gnu/libc.so.6(+0x37840)[0x7f7b92e28840]
/usr/bin/Xtightgatevnc(_ZN3rfb10SMsgWriter17writeNoDataUpdateEv+0x50)[0x55e8efc938c0]
/usr/bin/Xtightgatevnc(_ZN3rfb16VNCSConnectionST17writeNoDataUpdateEv+0x4c)[0x55e8efca18ac]
Trace:

/usr/bin/Xtightgatevnc(_ZN3rfb16VNCSConnectionST22writeFramebufferUpdateEv+0xf8)[0x55e8efca4168]
/usr/bin/Xtightgatevnc(_ZN3rfb16VNCSConnectionST15processMessagesEv+0xdf)[0x55e8efca42cf]
/usr/bin/Xtightgatevnc(_ZN14XserverDesktop17handleSocketEventEiPN7network12SocketServerEbb+0xdd)[0x55e8efc70f8d]
/usr/bin/Xtightgatevnc(_ZN14XserverDesktop17handleSocketEventEibb+0x6f)[0x55e8efc7105f]
/usr/bin/Xtightgatevnc(ospoll_wait+0x81)[0x55e8f0276481]
/usr/bin/Xtightgatevnc(WaitForSomething+0x22d)[0x55e8f026f02d]
/usr/bin/Xtightgatevnc(Dispatch+0xc6)[0x55e8f0208de6]
/usr/bin/Xtightgatevnc(dix_main+0x3ac)[0x55e8f020ee9c]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xeb)[0x7f7b92e1509b]
/usr/bin/Xtightgatevnc(_start+0x2a)[0x55e8efb6212a]

CendioOssman:
It is not a pointer, so it can never be `NULL`. Hence `begin()` cannot fail. Perhaps you have local modification to this area?

PS. Have you though about pushing your modifications upstream? They might have general interest.

amonott:
http://www.cplusplus.com/reference/list/list/begin/ says:
"If the container is empty, the returned iterator value shall not be dereferenced."
begin() does not fail, but you must not use the result, if the list is empty.

This matches our tests - after avoiding empty lists, the crash was no longer reproducable. Before, we could reliably reproduce.

We are planning to port some of our changes back to plain TigerVNC, but it takes a significant amount of work.

CendioOssman:
Right, but we do not dereference it, since http://www.cplusplus.com/reference/list/list/end/:
" If the container is empty, this function returns the same as list::begin."

There must be something else going on with your crash.

amonott:
I have not been able to find another reason and the fix works fine here. This means we will keep that change in our tree until we find something else.

CendioOssman:
Alright. I'll close this PR, but please let us know if you find out anything more.

PR #814, opened 2019-04-01
FdInStream::overrun(): use configurable read ahead size with sensible default

amonott:
perf has shown that select() got called way too often because of too small
buffer read sizes of 8 bytes in worst case, slowing down the client
significantly with lots of CPU overhead.
This change makes that minimum size configurable with a default of 2048.

Signed-off-by: Amon Ott <a.ott@m-privacy.de>

### DIFF
--- a/common/rdr/FdInStream.cxx
+++ b/common/rdr/FdInStream.cxx
@@ -1,4 +1,5 @@
 /* Copyright (C) 2002-2005 RealVNC Ltd.  All Rights Reserved.
+ * Copyright (C) 2019 m-privacy GmbH
  * 
  * This is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


CendioOssman:
Do you see that comment above the changed code? :)

The timing code is already a bit inaccurate, but I don't feel comfortable making it worse. I think the proper way forward here is to come up with a better line speed estimation algorithm. At that point we can (hopefully) start reading as much as the buffer has space for.

amonott:
I saw the comment, so I kept the code itself. However, with near fullscreen video, the Linux client spent about a third of its CPU time in select(), felt very sluggish and heavily triggered our automatic quality and rate reduction. After the change, select() cpu time was ignorable and quality stayed high.

amonott:
Extra note: Our auto quality adaption relies on round trip time through the client and ignores the timing values. So it was easier for us to make timing worse. Unfortunately, this feature needs our multiplexing extension, which is not easy to port back.

CendioOssman:
That is nothing we've seen, and we tend to push a lot of pixels through.

Can your multiplexing extension perhaps be the cause of all the small reads?

amonott:
It may be that the our reading pattern above FdInStream is different now, cannot really tell. The difference when tuning that value was enormous, especially unter Windows 10 - from almost impossible to watch a video to good quality.

PR #815, opened 2019-04-04
Print a call trace when segfaulting.

amonott:
On server side, we only print with non-Windows.

This call trace feature has helped us a lot in finding bugs in our many extensions.
ATM, I cannot test compile TigerVNC under Windows, but it works with our own builds.

### DIFF
--- a/common/rfb/SConnection.cxx
+++ b/common/rfb/SConnection.cxx
@@ -1,5 +1,6 @@
 /* Copyright (C) 2002-2005 RealVNC Ltd.  All Rights Reserved.
  * Copyright 2011 Pierre Ossman for Cendio AB
+ * Copyright 2019 m-privacy GmbH, Berlin
  * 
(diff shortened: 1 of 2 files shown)


CendioOssman:
Nice. This could indeed be helpful.

The server generally already has this though, as Xorg has the same functionality. Or are you not using Xvnc?

And do you think you could get it to use the logging system rather than a hard coded file?

amonott:
Since the segfault might also happen in the logging code, anything could happen. So I would rather keep it writing an independent file. Could still try the logging code afterwards though.

CendioOssman:
Extremely few calls are safe to use in a signal handler anyway, so this is largely a best effort kind of thing. Having output showing up in some other random file by default is a very weird behaviour IMO. So I really think this should go via the logging system. It's better we try to audit that for reliable behaviour if that is a concern.

Issue #652, opened 2018-05-28
Crash in Congestion.cxx after position wrap around


astrand:
Originally reported at: https://github.com/x11clone/x11clone/issues/10

The new congestion code per https://github.com/TigerVNC/tigervnc/pull/442 does not handle the case when the bytestream position wraps around. In some cases, this gives an assert of type:

Congestion.cxx:280: int rfb::Congestion::getUncongestedETA(): Assertion `iter != pings.end()' failed.
Aborted (core dumped)

lprylli:
I believe the fix in https://github.com/TigerVNC/tigervnc/commit/c05fba745ee42661b54bb4454b270e43232c1b is partly incorrect, I still get a crash like the original on tigervnc compiled on FreeBSD with clang.

Looking at coredump and disassembly of crash, it seems clang occasionally compiles the isAfter() statement:
 ` (int)a - (int)b > 0`
like if it was:
 ` (int)a > (int)b`

which means the wrapping case is still not handled correctly. I think clang is allowed to do that as per C standard, the result of overflow on signed integers is undefined and in the absence of overflow,` (int)a - (int)b > 0`  is equivalent to ` (int)a > (int)b `

So I would suggest isAfter to be changed instead to something with well-defined C behavior:
`(int)(a - b) > 0 `

Here is a snippet of code showing the problem with current definition for isAfter, on both freebsd/linux C or C++, the result of compiling/executing that snippet with "clang -O2 -fno-inline t.c":
```
0
1
1
```
(while gcc gives `1 1 1`).

```
// clang -O2 -fno-inline t.c
#include <stdbool.h>
#include <stdio.h>

bool
is_greater(unsigned x, unsigned y)
{
  return (int)x - (int)y > 0;
}

void
foo(unsigned x, unsigned y)
{
  printf("%d\n", is_greater(x,y));
  printf("%d\n", (int)x - (int)y > 0);
  printf("%d\n", (int)(x - y) > 0);
}

int
main()
{
  foo(0x80000001, 0x7fff0000);
}
```

PR #817, opened 2019-04-10
Fix wrap-aware isAfter function in Congestion.cxx

lprylli:
This is a follow-up on crash observed in https://github.com/TigerVNC/tigervnc/issues/652
the fix done there does not work in all environments.

Result of overflow on signed integer arithmetic is undefined in C/C++ standard.
So in previous version clang can compile the statement as (int)a > (int)b (i.e. assuming no overflow), which leads to incorrect result.

Correct deterministic behavior means doing overflow arithmetic as unsigned, using:
   (int)(a - b) > 0

### DIFF
--- a/common/rfb/Congestion.cxx
+++ b/common/rfb/Congestion.cxx
@@ -70,7 +70,7 @@ static const unsigned MAXIMUM_WINDOW = 4194304;
 
 // Compare position even when wrapped around
 static inline bool isAfter(unsigned a, unsigned b) {
-  return (int)a - (int)b > 0;
+  return a != b && a - b <= UINT_MAX / 2;
 }
 
 static LogWriter vlog("Congestion");


CendioOssman:
You're right, even though I'm not sure I agree with clang's decision to exploit that part of the specification.

However your new suggestion is also undefined (technically "implementation-defined"). This behaviour is even more de-facto than signed overflow, but it is not enough if we want to have a fully defined behaviour.

lprylli:
BTW, one option for clang to be conservative about any rearrangements of signed-integer operations is to add compiler option -fwrapv (and I agree with you I also wish clang would have -fwrapv as the default).

That said, I don't think my proposed-change is implementation-defined, can you clarify? The construction is a common one relying on specification assertions in the C standard (taking for reference http://www.open-std.org/JTC1/SC22/WG14/www/standards, C11 draft, but I think I remember original C standards always had same rules in that respect). There the semantics of unsigned "overflow" arithmetic is well-defined (see 6.2.5 alinea 9), and the conversion from unsigned to signed is also well-defined (see section 6.3.1.3).


The example code-snippet I put in https://github.com/TigerVNC/tigervnc/issues/652 (copied https://gist.github.com/lprylli/be21fa19fba3414ef40a7bb83c2aace5 ) can actually be examined by compiling it with `clang -fsanitized=undefined` (more details at https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html ) to at least check clang interpretation of defined vs undefined in C standard agrees with my above analysis.

CendioOssman:
See 6.3.1.3, point 3:
> Otherwise, the new type is signed and the value cannot be represented in it; either the
result is implementation-defined or an implementation-defined signal is raised.

If `a` is `0xf00000f` and `b` is `0x0000000f` then the subtracted value is `0xf0000000` which cannot be represented by an `int`.

lprylli:
Silly me, referencing the standard without reading it correctly.

I have updated the solution to be fully C-standard compliant.

I misremembered previous personal investigation where the conclusion was that implementation-defined "unsigned" to "int" conversion was actually well-defined for all systems I could research (both gcc/clang/microsoft couple others document it explicitly), so I know I am not afraid to use it, but that's definitely not the same as claiming standard provides any guarantees.

common/rfb/Congestion.cxx
```diff
@@ -70,7 +70,7 @@ static const unsigned MAXIMUM_WINDOW = 4194304;
 
 // Compare position even when wrapped around
 static inline bool isAfter(unsigned a, unsigned b) {
-  return (int)a - (int)b > 0;
+  return a - b > 0 && a - b <= INT_MAX;
```
CendioOssman: `a - b > 0` is not that clear as it is mostly always true for unsigned integers. Might I suggest `a != b`?

And I wonder if `UINT_MAX/2` is better as it avoids assuming any specific connection between `int` and `unsigned`?

lprylli: done

CendioOssman:
Great, thanks. :)

PR #829, opened 2019-04-25
Initial support for touch gesture recognition in Linux client

aeneby:
This PR implements left/middle/right click gestures using 1/3/2 finger taps respectively. These gestures are interpreted client-side and then translated into something the server understands and can act on. Still to come are scroll and zoom gesture support.


samhed:
Nice @aeneby! I'll try to do some testing of this tomorrow

vncviewer/GestureHandler.cxx
```diff
@@ -0,0 +1,180 @@
...
+        expireEvents();
+
+      // FIXME: By this point we should have a state, right?
+      //assert(hasState(this->state));
+
+      switch (this->state) {
+        case GH_LEFTBTN:
+	  processed.push_back({1,
+			       devev->event_x,
+			       devev->event_y,
```
CendioOssman: I think we might need some threshold here for clicks. I.e. you need to move your finger a certain distance, otherwise the press and release will both be at the same coordinate.

If we don't do this then you can get problems with clicks not registering as the system thinks it was a drag.

vncviewer/GestureHandler.cxx
```diff
@@ -0,0 +1,180 @@
...
+
+  return num_expired;
+}
+
+unsigned char GestureHandler::getState() {
+  return this->state;
+}
+
+bool GestureHandler::hasState(unsigned char state) {
+  return state && !(state & (state - 1));
```
CendioOssman: Huh?

vncviewer/GestureHandler.cxx
```diff
@@ -0,0 +1,180 @@
...
+			       devev->event_x,
+			       devev->event_y,
+			       GH_ButtonRelease});
+	  break;
+
+        case GH_RIGHTBTN:
+	  processed.push_back({4,
+			       devev->event_x,
+			       devev->event_y,
+			       GH_ButtonRelease});
```
CendioOssman: We probably need to be more careful about coordinates once multiple fingers are involved. Either the first finger, or an average, depending on what feels most natural.

vncviewer/GestureHandler.h
```diff
@@ -0,0 +1,74 @@
...
+
+#define GH_INITSTATE   (255 & ~GH_UNDEFINED)
+
+// Timeout when waiting for gesture
+#define GH_TOUCHDELAY  0.25 // s
+
+enum GHEventType {
+  GH_ButtonPress,
+  GH_ButtonRelease,
+  GH_TouchUpdate,
```
CendioOssman: Did you look at having a more gesture based abstraction here? I.e. events like `GH_SingleTap`, `GH_Drag`, H_Pinch`? I think it might make it clearer what it is the logic is trying to detect. Then we can let a different layer define what that means in terms of pointer events.

vncviewer/Viewport.cxx
```diff
@@ -1273,16 +1318,27 @@ int Viewport::handleSystemEvent(void *event, void *data)
               break;
           }
           break;
+
         case XI_TouchBegin:
-          buttonMask = 1;
+          if (gh->registerEvent(devev) == 1) {
+            // First touch. Give the user some time to complete
+	    // a gesture, then come back and process the events.
+            Fl::add_timeout(GH_TOUCHDELAY, touchTimeout, self);
```
CendioOssman: Can't we hide this inside `GestureHandler`? It seems odd that these details are forced upon `Viewport`.

vncviewer/GestureHandler.cxx
```diff
@@ -41,10 +42,11 @@ int GestureHandler::registerEvent(void *ev) {
...
       // For now, just pass the update on to be processed normally.
-      processed.push_back({0,
-		           devev->event_x,
-		           devev->event_y,
-			   GH_TouchUpdate});
+      ghev.detail = 0;
+      ghev.event_x = devev->event_x;
+      ghev.event_y = devev->event_y;
+      ghev.type = GH_TouchUpdate;
+      processed.push_back(ghev);
```
CendioOssman: You could always create a method that takes those for values as arguments. Then you'd get something very similar to your original `push_back()`.

vncviewer/GestureHandler.cxx
```diff
@@ -180,6 +171,15 @@ bool GestureHandler::hasState(unsigned char state) {
   return state && !(state & (state - 1));
 }
 
+int GestureHandler::isTracked(XIDeviceEvent ev) {
```
CendioOssman: Shouldn't this be `bool` given the usage?

vncviewer/Viewport.cxx
```diff
@@ -215,6 +218,27 @@ Viewport::~Viewport()
 }
 
 
+#if ! (defined(WIN32) || defined(__APPLE__))
+void Viewport::selectEvents() {
+  XIEventMask eventmask;
+  unsigned char flags[XIMaskLen(XI_LASTEVENT)] = { 0 };
+
+  eventmask.deviceid = XIAllMasterDevices;
```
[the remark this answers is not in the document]
aeneby: Yes, as far as I'm aware. I don't have a hotpluggable touch device to confirm though, perhaps @samhed could try this as part of his testing.

vncviewer/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ else()
   add_executable(vncviewer ${VNCVIEWER_SOURCES})
 endif()
 
-target_link_libraries(vncviewer rfb network rdr os ${FLTK_LIBRARIES} ${GETTEXT_LIBRARIES})
+target_link_libraries(vncviewer rfb network rdr os ${X11_Xi_LIB} ${FLTK_LIBRARIES} ${GETTEXT_LIBRARIES})
```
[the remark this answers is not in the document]
aeneby: You mean a separate directive for just `${X11_Xi_LIB}`?

vncviewer/GestureHandler.cxx
aeneby: I'd like it to return an index to the relevant event in the `tracked` vector if it exists, `-1` if it doesn't.

vncviewer/GestureHandler.cxx
aeneby: Yeah, this sort of thing is going to be happening all over the place, so I plan to provide some kind of short-hand for it in due course. Point taken.

vncviewer/Viewport.cxx
```diff
@@ -594,7 +626,7 @@ int Viewport::handle(int event)
     // We want a last move event to help trigger edge stuff
     handlePointerEvent(Point(Fl::event_x() - x(), Fl::event_y() - y()), 0);
     return 1;
-
+#if (defined(WIN32) || defined(__APPLE__))
```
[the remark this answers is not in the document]
aeneby: 96a60a0

vncviewer/Viewport.cxx
```diff
@@ -1174,8 +1207,87 @@ int Viewport::handleSystemEvent(void *event, void *data)
     self->handleKeyRelease(keycode);
     return 1;
   }
-#endif
```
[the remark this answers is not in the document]
aeneby: Probably, I think it's fixed with 96a60a0. I'll squash these commits at some point and it'll be easier to see.

vncviewer/Viewport.cxx
```diff
@@ -551,10 +575,18 @@ int Viewport::handle(int event)
...
   int buttonMask, wheelMask;
+#endif // !LINUX
   DownMap::const_iterator iter;
 
   switch (event) {
+#if ! (defined(WIN32) || defined(__APPLE__))
+  case FL_SHOW:
+    selectEvents();
+    break;
+#endif // LINUX
```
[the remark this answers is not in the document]
aeneby: Not sure I understand...

vncviewer/GestureHandler.cxx
aeneby: I'll close this comment since it references old code, but yes, agreed. Testing should give us an idea of what works best here.

vncviewer/Viewport.h
```diff
@@ -96,6 +96,10 @@ class Viewport : public Fl_Widget {
 
   static void handleOptions(void *data);
 
+#if ! (defined(WIN32) || defined(__APPLE__))
```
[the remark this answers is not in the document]
aeneby: Yeah wasn't sure what to do here. Is empty functions preferred?

vncviewer/GestureHandler.cxx
aeneby: Old code but I think I know what you mean and agree. I'm already working on this type of threshold for the scroll gesture (i.e. how far do fingers have to move before a right-click becomes a scroll?).

vncviewer/GestureHandler.cxx
aeneby: 25cb75b

vncviewer/GestureHandler.h
aeneby: I'm still thinking about this. I'm leaning towards just having a `GH_GestureBegin` event, with a member which describes which gesture has begun. The corresponding `GH_GestureEnd` wouldn't need this information since only one gesture is active at a time. Note that things like a single tap or click are essentially a gestures now since they're dependent on a timeout.

vncviewer/Viewport.cxx
aeneby: Maybe. There needs to be a way to make `Viewport` process the event queue immediately after timeout, without having to wait for another touch event to trigger the processing. Originally, I was playing around with `PropertyNotify` events, which are triggered whenever a window property changes. `GestureHandler` could modify a dummy window property to trigger a `PropertyNotify` event, and we could use this event to trigger processing of the queue (you can distinguish 'trigger' events based on the property which was changed). One potential benefit of doing this is that you get a current "X11" timestamp, which you can use to determine how old your tracked touches are. But this seems majorly convoluted. Open to ideas if you have any.

vncviewer/CMakeLists.txt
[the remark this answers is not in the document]
CendioOssman: Yup. Like there is for special Windows and macOS libraries just below that line.

vncviewer/Viewport.cxx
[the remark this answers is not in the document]
CendioOssman: You should be able to test using just a regular mouse, since you handle those events as well.

vncviewer/Viewport.cxx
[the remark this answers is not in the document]
CendioOssman: The code you are writing is equally valid on Solaris or AIX. It's not Linux specific.

vncviewer/Viewport.h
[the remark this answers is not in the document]
CendioOssman: No clear guidelines. Whatever you think is more readable. :)

vncviewer/GestureHandler.h
CendioOssman: That sounds like a very nice approach.

vncviewer/Viewport.cxx
CendioOssman: Right now you have a "pull" approach where `Viewport` pulls events from `GestureHandler`. You could switch to a "push" approach instead, like most of the existing callbacks you are using. I.e. define an interface class (e.g. `GestureEventHandler`), have `Viewport` subclass it and register itself with the `GestureHandler` object. `GestureHandler` can then inject events whenever it damn well pleases.

Have a look at `rfb::Timer` and you'll see how this pattern works in practice.

vncviewer/GestureHandler.cxx
CendioOssman: The caller of `isTracker()` doesn't seem to care about such details. :)

If you want to keep that return value, then I would suggest a different naming. "is" heavily implies a boolean return value.

vncviewer/Viewport.cxx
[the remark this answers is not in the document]
samhed: I'll test if hotplugging the touch screen works here.

vncviewer/CMakeLists.txt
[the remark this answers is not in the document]
aeneby: 556c85f

vncviewer/Viewport.cxx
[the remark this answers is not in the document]
aeneby: Still confused :)

"Copy the `#if`"? Do you mean because there's two different variations, `!(WIN32 || APPLE)` and `!WIN32 && !APPLE`? Is it just the closing comment you're referring to when you say "X11", or do you mean that there's a check for X11?

samhed:
I finally got some time over to build vncviewer and test using a touchscreen here on a Fedora 29 machine. Not really related, but the first thing to note is that Gnome's touch support sucks when using multiple screens:
 - the screen with touch had to be set as the primary screen
 - the two screens were one 1080p (touch) and one 1440p (not touch), it seemed to use the resolution of the larger 1440p screen for the coordinates for the touch input on the other screen.
 - it did not care if the screen was rotated, the touch input always assumed it was in standard landscape mode

As far as this PR goes:
 * input using a mouse seems to work flawlessly after the move to XInput 2, nice work!
 * 2 fingers to right-click felt very natural, nice!
 * I could not get middle-click gesture to work, is there any range threshold that I need to be aware of between my 3 fingers?
 * The delay added for detecting gestures seemed to be a bit long, but I guess this is easily tweaked?

aeneby:
Thanks @samhed,

>     * input using a mouse seems to work flawlessly after the move to XInput 2, nice work!
Excellent!
 
>     * 2 fingers to right-click felt very natural, nice!
I've noticed that a *fast* two-finger tap results in XInput2 registering `Down-Up-Down-Up` (two single-finger taps) rather than `Down-Down-Up-Up` (double-finger tap). Not sure if this is a peculiarity with XInput2, my particular touch screen/drivers, etc, or what. The consequence of this is that multi-finger taps need to be done quite "deliberately" in order to work, but even so, I found it felt quite natural to perform them, so glad you thought so too.

>     * I could not get middle-click gesture to work, is there any range threshold that I need to be aware of between my 3 fingers?
There is no range threshold. If you're using Gnome 3 as your local desktop for testing, you'll probably find that it's swallowing your three-finger tap in an attempt to interpret it itself, rather than passing it on to vncviewer. I had to install a Gnome Shell extension to disable gestures ([this one here](https://extensions.gnome.org/extension/1140/disable-gestures/)).

>     * The delay added for detecting gestures seemed to be a bit long, but I guess this is easily tweaked?
Yes, this is easily tweaked. It's set to 250ms currently (this is how long you have to begin a gesture) but this is arbitrary and easily changed.

A couple of other things you could test for next time (since strictly speaking these are gestures now too), are left click, left click-and-hold, left click-and-drag.

samhed:
> I've noticed that a fast two-finger tap results in XInput2 registering Down-Up-Down-Up (two single-finger taps) rather than Down-Down-Up-Up (double-finger tap). Not sure if this is a peculiarity with XInput2, my particular touch screen/drivers, etc, or what. The consequence of this is that multi-finger taps need to be done quite "deliberately" in order to work, but even so, I found it felt quite natural to perform them, so glad you thought so too.

Yes I noticed that too, I think this might feel better if we eventually reduce the delay to ~100 ms or something like that.

> There is no range threshold. If you're using Gnome 3 as your local desktop for testing, you'll probably find that it's swallowing your three-finger tap in an attempt to interpret it itself, rather than passing it on to vncviewer. I had to install a Gnome Shell extension to disable gestures (this one here).

Ah, great! It works well now.

> A couple of other things you could test for next time (since strictly speaking these are gestures now too), are left click, left click-and-hold, left click-and-drag.

Seems to work well.

One thing I noticed, and am unsure if it's a problem or not is that; if you do a 2-finger drag, then release one finger, while continuing to drag the other finger, the eventual 1-finger drag doesn't generate any input (same is true for 3-finger drags).
After thinking about it a bit I can't come up with a different natural way to handle it. I think it's good as it is now with regards to this.

Generally, the feeling I get is that you have made a robust system that works well so far!

And btw, I was able to use map the touch input to the correct screen using this:

`xinput map-to-output 'Melfas LGD AIT Touch Controller' VGA-1`

samhed:
Another thing that I noticed just now is that double-clicks are generally a bit difficult to do using a touch-screen in vncviewer. It's difficult to do two taps in the same place.

In noVNC we have added a range-threshold for double-taps on touch devices. If the time and distance between two taps is close enough, we lie about the position of the second tap, saying that it's the same as the first one.

I get the feeling that Gnome has done something similar locally on my Fedora machine as well, double-tapping is really easy here compared to when using vncviewer.

I know this is a bit unrelated, but if you think this is an easy fix - perhaps something to consider?

aeneby:
> One thing I noticed, and am unsure if it's a problem or not is that; if you do a 2-finger drag, then release one finger, while continuing to drag the other finger, the eventual 1-finger drag doesn't generate any input (same is true for 3-finger drags).

There's probably two points for discussion here:

- Should removal of one finger end a multi-touch gesture, or should we wait for all fingers to be removed?

- If we do end the gesture, should the remaining finger(s) generate input, and if so, what input?

Currently, removal of any one finger will end a multi-touch gesture, and the remaining fingers will be ignored. If this proves to be an issue in usability testing then perhaps we can re-evaluate this approach.

> Another thing that I noticed just now is that double-clicks are generally a bit difficult to do using a touch-screen in vncviewer. It's difficult to do two taps in the same place.

I have to say that after testing a bit just now, I don't experience the same problem with double clicking. But this could be due to differences in our test platforms - perhaps you have a higher dpi, meaning smaller movements result in a bigger pixel difference for example. I'm using a 1920x1080 touch screen at 96dpi.

> In noVNC we have added a range-threshold for double-taps on touch devices. If the time and distance between two taps is close enough, we lie about the position of the second tap, saying that it's the same as the first one.

Yes if double-clicking is difficult then we could certainly look at something like this. Part of the solution might be to start measuring distance in physical units (mm) rather than pixels, which is something I want to do eventually anyway.

aeneby:
Squashed PR commits and updated with zoom/scroll gesture detection.

samhed:
Hi again Aaron!

> Squashed PR commits and updated with zoom/scroll gesture detection.

Very nice! I have been testing today, and it looks promising! I was immediately impressed by how well the scroll-gesture worked in this first revision! Well done!

I have a few comments:

* The gesture timeout needs a few changes; it feels more like a delay at the moment. In many cases it feels like the code shouldn't have to wait until it can decide which gesture it was, examples:
  * when you do a quick two-finger tap - when the second finger hits we can rule out 1-finger gestures, and when both fingers release we should be able to tell that it wasn't a 'pan' or 'pinch'
  * when you do a quick two-finger swipe/pan/pinch - as soon as the fingers have moved a certain threshold we can know what kind of two-finger gesture it is
* The coordinates of the resulting mouse events would probably feel more natural if they were positioned in the center between the fingers rather than at the position of the first finger
* The two-finger press gesture should send the right-click event when the fingers are released (and haven't moved too far) rather than after that 250 ms timeout. This would likely make it easier to distinguish this gesture from the other 2-finger ones as a bonus.
* The sensitivity of both the zoom and the scroll gestures feels too high. I haven't looked at the code yet, but the interval of how often scroll and ctrl+scroll input is sent should be distance based, and not trigger on every gesture event. Perhaps viewport can aggregate multiple gesture events and have distance thresholds for when to send the input to the server. Again, I haven't had time to look at the code yet, sorry.

And lastly, when using this I get the feeling that I miss to be able to right-click by using a "1-finger long-press". The largest PC platform for touch is Windows, and long-press is the only way to right-click using a touch screen and Windows. The Windows client/vncviewer platform is very important and I believe we should try to mimic their behavior as much as we can, so:

 * Would it be too much work to get 1-finger long-press to send right clicks as well?

aeneby:
Hey @samhed!

>       * when you do a quick two-finger tap - when the second finger hits we can rule out 1-finger gestures, and when both fingers release we should be able to tell that it wasn't a 'pan' or 'pinch'

This is already the case. Perhaps you're getting confused by the fact that a two-finger tap has to be made quite deliberately in order to work? This has nothing to do with the delay, but seems to be related to XInput2 - a fast two-finger tap will produce events in the order "Down->Up->Down->Up" (rather than "Down->Down->Up->Up"), which is impossible to distinguish from two single-finger taps.

>       * when you do a quick two-finger swipe/pan/pinch - as soon as the fingers have moved a certain threshold we can know what kind of two-finger gesture it is

Agreed - at the moment zoom/scroll gestures aren't recognised until the timeout, and this could be improved. Reducing the timeout would probably improve responsiveness here too.

>     * The coordinates of the resulting mouse events would probably feel more natural if they were positioned in the center between the fingers rather than at the position of the first finger

Agreed, I'll see what I can do.

>     * The two-finger press gesture should send the right-click event when the fingers are released (and haven't moved too far) rather than after that 250 ms timeout. This would likely make it easier to distinguish this gesture from the other 2-finger ones as a bonus.

This is already the case. All click gestures are registered as soon as a finger is released.

>     * The sensitivity of both the zoom and the scroll gestures feels too high. I haven't looked at the code yet, but the interval of how often scroll and ctrl+scroll input is sent should be distance based, and not trigger on every gesture event. Perhaps viewport can aggregate multiple gesture events and have distance thresholds for when to send the input to the server. Again, I haven't had time to look at the code yet, sorry.

Agree with you here too. We do need some kind of throttling mechanism, and it probably needs to be distance based as you say. Will look into this too.

> And lastly, when using this I get the feeling that I miss to be able to right-click by using a "1-finger long-press". The largest PC platform for touch is Windows, and long-press is the only way to right-click using a touch screen and Windows. The Windows client/vncviewer platform is very important and I believe we should try to mimic their behavior as much as we can, so:
> 
>     * Would it be too much work to get 1-finger long-press to send right clicks as well?

Well, this gesture would obviously conflict with left-click press-and-hold. Whether this is an issue or not depends on usage, but to be honest I can't really think of a case where left-click press-and-hold is useful. Perhaps we could add a compile-time option to specify this behaviour. But would a two-finger tap be right click as well in this case, or something else?

Appreciate the detailed feedback :)

samhed:
>> * when you do a quick two-finger tap - when the second finger hits we can rule out 1-finger gestures, and when both fingers release we should be able to tell that it wasn't a 'pan' or 'pinch'
>
>This is already the case. Perhaps you're getting confused by the fact that a two-finger tap has to be made quite deliberately in order to work? This has nothing to do with the delay, but seems to be related to XInput2 - a fast two-finger tap will produce events in the order "Down->Up->Down->Up" (rather than "Down->Down->Up->Up"), which is impossible to distinguish from two single-finger taps.

Hm, okay. I can see what you are talking about, I tested using 'xev' in the session. I guess there's not much we can do in that case.

>>* The two-finger press gesture should send the right-click event when the fingers are released (and haven't moved too far) rather than after that 250 ms timeout. This would likely make it easier to distinguish this gesture from the other 2-finger ones as a bonus.
>
>This is already the case. All click gestures are registered as soon as a finger is released.

Not for me; I recorded a video with my phone showing my fingers pressing on the touch screen:

https://www.youtube.com/watch?v=0usK3Tv1jEc

What could be happening here you think?

>>* Would it be too much work to get 1-finger long-press to send right clicks as well?
>
>Well, this gesture would obviously conflict with left-click press-and-hold. Whether this is an issue or not depends on usage, but to be honest I can't really think of a case where left-click press-and-hold is useful. Perhaps we could add a compile-time option to specify this behaviour. But would a two-finger tap be right click as well in this case, or something else?

Sounds good! Yes, a two-finger tap would still be a right click as well. This would mean just adding an extra way of right clicking.

samhed:
>>>when you do a quick two-finger tap - when the second finger hits we can rule out 1-finger gestures, and when both fingers release we should be able to tell that it wasn't a 'pan' or 'pinch'
>>
>>This is already the case. Perhaps you're getting confused by the fact that a two-finger tap has to be made quite deliberately in order to work? This has nothing to do with the delay, but seems to be related to XInput2 - a fast two-finger tap will produce events in the order "Down->Up->Down->Up" (rather than "Down->Down->Up->Up"), which is impossible to distinguish from two single-finger taps.
>
>Hm, okay. I can see what you are talking about, I tested using 'xev' in the session. I guess there's not much we can do in that case.

When using 'xinput test-xi2' locally on my workstation I can get things to work the way we want. A quick two-finger tap results in:

touchBegin->touchBegin->touchUpdate->touchEnd->touchEnd

Is there any timestamps or similar we can look at for these events? Perhaps the order the events are received can't be fully trusted?

> Not for me; I recorded a video with my phone showing my fingers pressing on the touch screen:
>
>https://www.youtube.com/watch?v=0usK3Tv1jEc
>
>What could be happening here you think?

Could the source of confusion be clicks vs mouse down here? In my mind we should do clicks for the two-finger taps, that is when the fingers are released we should send both mouse down and mouse up without any time in between.

aeneby:
> When using 'xinput test-xi2' locally on my workstation I can get things to work the way we want. A quick two-finger tap results in:
> 
> touchBegin->touchBegin->touchUpdate->touchEnd->touchEnd
> 
> Is there any timestamps or similar we can look at for these events? Perhaps the order the events are received can't be fully trusted?

I tested using `xinput --test-xi2` on my local desktop and actually get the same (incorrect) behaviour. So it must be something weird on my end. If it's working for you then perfect, we don't have a problem, so long as I'm the exception to the rule and not you :)

> Could the source of confusion be clicks vs mouse down here? In my mind we should do clicks for the two-finger taps, that is when the fingers are released we should send both mouse down and mouse up without any time in between.

Again, this is the case already :) what you were doing in the video was a right-button click-and-hold gesture, not a click. For the click-and-hold gesture, the click down will not register until the timeout, since we can't be sure until then what the gesture is. Once the fingers are released, the click will be released too. If you two-finger tap and release your fingers before the timeout, however, a click down and click up will be sent immediately - in that order - because we know it can't possibly be any other gesture once the fingers are released.

I think I get what you're saying, though. You're saying that the gesture you're making in the video should really be interpreted as an "exaggeratedly long" right click, with both the mouse down and mouse up events triggering only when the fingers are released. But this would mean we lose the ability to do a right-button click-and-hold, just like we lose the possibility to do a left-button click-and-hold if we interpret a single-finger long press to be a right click (stay with me... :)

I think we have to remember that we're not trying to design a touch interface here. We're trying to emulate mouse input using a touch screen, for applications and desktop environments which are typically mouse-driven. So even though click-and-hold gestures might not be commonplace in pure touch environments, if it's possible to do it with a mouse then we should probably make it possible to do with our fingers too (some right-click context menus require the button to be held down to stay open, for example).

aeneby:
Just to summarize, here are the points I've got written down so far:

- Recognition of scroll/zoom gestures as soon as movement threshold has been reached, rather than waiting for timeout

- Coordinates of a multi-touch tap should be an average of the fingers' coordinates, rather than that of the first touch

- Implement some kind of sensitivity adjustment for scroll/zoom gestures

- Add compile-time option for single-finger long-press to register right click

samhed:
> I tested using xinput --test-xi2 on my local desktop and actually get the same (incorrect) behaviour. So it must be something weird on my end. If it's working for you then perfect, we don't have a problem, so long as I'm the exception to the rule and not you :)

Hm, okay. But in that case, why does a quick two-finger tap not generate a right click for me?

Locally on my workstation `xinput --test-xi2` tells me very consistently that I get the events in the correct order (begin->begin->end->end). Shouldn't the code give us a right click in that case?

Should I enable some sort of debug logging to share with you?

>Again, this is the case already :) what you were doing in the video was a right-button click-and-hold gesture, not a click. For the click-and-hold gesture, the click down will not register until the timeout, since we can't be sure until then what the gesture is. Once the fingers are released, the click will be released too. If you two-finger tap and release your fingers before the timeout, however, a click down and click up will be sent immediately - in that order - because we know it can't possibly be any other gesture once the fingers are released.

Well explained, that makes sense. :+1: 

>I think I get what you're saying, though. You're saying that the gesture you're making in the video should really be interpreted as an "exaggeratedly long" right click, with both the mouse down and mouse up events triggering only when the fingers are released.

Correct, that is what I meant.

>But this would mean we lose the ability to do a right-button click-and-hold, just like we lose the possibility to do a left-button click-and-hold if we interpret a single-finger long press to be a right click (stay with me... :)
>
>I think we have to remember that we're not trying to design a touch interface here. We're trying to emulate mouse input using a touch screen, for applications and desktop environments which are typically mouse-driven. So even though click-and-hold gestures might not be commonplace in pure touch environments, if it's possible to do it with a mouse then we should probably make it possible to do with our fingers too (some right-click context menus require the button to be held down to stay open, for example).

You are right, but I actually think we can get the best of two worlds here! Looking at how it works on Windows:

 * 1-finger tap -> left click
 * 1-finger drag -> left click and drag
 * 1-finger long press -> right click (mousedown+mouseup when finger is released)
 * 1-finger long press and then drag -> right click and drag

When testing on our Windows machine here, it feels very natural. Do you think it would be possible to mimic this behavior at least with regards to left clicks? This would mean cancelling the long-press gesture as soon as you move the finger and at that point send a left-mousedown followed by the move and eventually left-mouseup as you release the finger.

We could wait with right click and drag, and I think an initial version of this would be fine without it. Feels like a natural future gesture to add.

>Just to summarize, here are the points I've got written down so far:
>
> * Recognition of scroll/zoom gestures as soon as movement threshold has been reached, rather than waiting for timeout
>
> * Coordinates of a multi-touch tap should be an average of the fingers' coordinates, rather than that of the first touch
>
> * Implement some kind of sensitivity adjustment for scroll/zoom gestures
>
> * Add compile-time option for single-finger long-press to register right click

Yes, that's great! And given my last comment regarding long-press gestures, it might even be good enough to do without having an option to enable it?

aeneby:
> Hm, okay. But in that case, why does a quick two-finger tap not generate a right click for me?
> 
> Locally on my workstation `xinput --test-xi2` tells me very consistently that I get the events in the correct order (begin->begin->end->end). Shouldn't the code give us a right click in that case?
> 
> Should I enable some sort of debug logging to share with you?

I see what you're saying, strange. What you could do is put some `vlog.info()` lines in `Viewport::handleSystemEvent()` for the `XI_TouchBegin` `XI_TouchUpdate` and `XI_TouchEnd` events. Then you can see what order the events are coming in, before they even get to the `GestureHandler` code. It would be interesting to see if they're still in the correct order.
 
> You are right, but I actually think we can get the best of two worlds here! Looking at how it works on Windows:
> 
>     * 1-finger tap -> left click
> 
>     * 1-finger drag -> left click and drag
> 
>     * 1-finger long press -> right click (mousedown+mouseup when finger is released)
> 
>     * 1-finger long press and then drag -> right click and drag
> 
> 
> When testing on our Windows machine here, it feels very natural. Do you think it would be possible to mimic this behavior at least with regards to left clicks? This would mean cancelling the long-press gesture as soon as you move the finger and at that point send a left-mousedown followed by the move and eventually left-mouseup as you release the finger.

I think this sounds sensible, and achievable, so long as we're happy to lose the ability to click and hold in one place ("long click"). I can have a look at how much we can get done now, and which parts we can perhaps look at in future.

samhed:
>I see what you're saying, strange. What you could do is put some vlog.info() lines in Viewport::handleSystemEvent() for the XI_TouchBegin XI_TouchUpdate and XI_TouchEnd events. Then you can see what order the events are coming in, before they even get to the GestureHandler code. It would be interesting to see if they're still in the correct order.

Ok, I'll do this tomorrow.

>>     * 1-finger tap -> left click
>> 
>>     * 1-finger drag -> left click and drag
>> 
>>     * 1-finger long press -> right click (mousedown+mouseup when finger is released)
>> 
>>     * 1-finger long press and then drag -> right click and drag
>> 
> 
>I think this sounds sensible, and achievable, so long as we're happy to lose the ability to click and hold in one place ("long click"). I can have a look at how much we can get done now, and which parts we can perhaps look at in future.

Great! And yeah, the ability to "long click" isn't necessary.

vncviewer/Viewport.cxx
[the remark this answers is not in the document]
CendioOssman: It's just the closing comment, yes. :)

It's definitely not just Linux, and for sanity it's usually best to verbatim copy the `#if` statement. I.e.:

```C
#ifdef FOO || BAR
#endif // FOO || BAR
```

CendioOssman:
Some build fixes needed:

```diff
diff --git a/vncviewer/CMakeLists.txt b/vncviewer/CMakeLists.txt
index 6e54bc69..498ddf06 100644
--- a/vncviewer/CMakeLists.txt
+++ b/vncviewer/CMakeLists.txt
@@ -12,7 +12,6 @@ set(VNCVIEWER_SOURCES
   OptionsDialog.cxx
   PlatformPixelBuffer.cxx
   Viewport.cxx
-  GestureHandler.cxx
   parameters.cxx
   keysym2ucs.c
   vncviewer.cxx)
@@ -32,7 +31,7 @@ if(WIN32)
 elseif(APPLE)
   set(VNCVIEWER_SOURCES ${VNCVIEWER_SOURCES} cocoa.mm osx_to_qnum.c)
 else()
-  set(VNCVIEWER_SOURCES ${VNCVIEWER_SOURCES} xkb_to_qnum.c)
+  set(VNCVIEWER_SOURCES ${VNCVIEWER_SOURCES} GestureHandler.cxx xkb_to_qnum.c)
 endif()
 
 if(WIN32)
@@ -51,18 +50,14 @@ endif()
 
 target_link_libraries(vncviewer rfb network rdr os ${FLTK_LIBRARIES} ${GETTEXT_LIBRARIES})
 
-if(UNIX)
-  target_link_libraries(vncviewer ${X11_Xi_LIB})
-endif()
-
 if(WIN32)
   target_link_libraries(vncviewer msimg32)
-endif()
-
-if(APPLE)
+elseif(APPLE)
   target_link_libraries(vncviewer "-framework Cocoa")
   target_link_libraries(vncviewer "-framework Carbon")
   target_link_libraries(vncviewer "-framework IOKit")
+else()
+  target_link_libraries(vncviewer ${X11_Xi_LIB})
 endif()
 
 install(TARGETS vncviewer DESTINATION ${BIN_DIR})
```

samhed:
I tested the latest commits as well, and it feels like a good improvement.

 * In my mind it's a lot easier to aim when using multi-finger gestures since the resulting input is now sent in the middle of the fingers

 * It's still a bit too sensitive when it comes to scrolling and zooming I think. The end-users will likely want to be able to scroll and zoom carefully while reading.

aeneby:
@CendioOssman, @samhed, 

3480dc0 is my suggested patch for fixing the grab issue in fullscreen. Feedback please. It attempts to grab all master pointer devices, and fails if it can't grab all of them. I don't see any way of knowing exactly which pointer device we should be grabbing, unless this could be user-specified somehow.

It also involves evaluating all attached devices each time a grab or ungrab occurs, which is potentially inefficient. An alternative to this would be to evaluate all devices on startup and store this list somewhere, but that would mean we lose the ability to hotplug - the user would need to restart vncviewer to use a newly attached touch device.

aeneby:
@samhed 

>     * It's still a bit too sensitive when it comes to scrolling and zooming I think. The end-users will likely want to be able to scroll and zoom carefully while reading.

I've implemented some more fine-tunable adjustments in ef6f1e4, and set them to something which I think works okay. See what you think. If it's still too sensitive then perhaps try tweaking the values and let me know what works for you.

CendioOssman:
I found a bug with regard to pointer grabbing. Set up a single monitor full screen, then start dragging stuff in the session. Once the pointer leaves the window, the dragging stops. This is a regression.

What used to happen was:

1. We grab the pointer using `XGrabPointer()` as we enter fullscreen and the pointer is inside the viewer window.
2. When we pushed the button we got a `ButtonPress` which FLTK translated to `FL_PUSH`
3. We handled the `FL_PUSH` in `Viewport`, which got FLTK to register `Viewport` as the "pushed" widget
4. All subsequent `MotionNotify` got translated to `FL_DRAG` and sent directly to `Viewport`
5. When the pointer left the window we got a `LeaveNotify`. However we also got a `MotionNotify` which makes FLTK ignore the `LeaveNotify` (because reasons)
6. Once we released the button FLTK stopped considering `Viewport` as "pushed" and started delivering events to `DesktopWindow`
7. `DesktopWindow` notices the pointer is outside the window and releases the grab

The new code breaks this as the `MotionNotify` and `ButtonPress` events are now gone:

1. We grab the pointer using `XIGrabDevice()` as we enter fullscreen and the pointer is inside the viewer window.
2. When we pushed the button we got a `XI_ButtonPress` which FLTK just ignores
3. All subsequent `XI_Motion` is also ignored by FLTK
4. When the pointer left the window we got a `LeaveNotify`. But since we nog longer get a `MotionNotify` FLTK now properly sends a `FL_LEAVE`
5. `DesktopWindow` notices the pointer has left the window and releases the grab

CendioOssman:
We may have more hidden issues on the same theme. FLTK also has the concept of "widget below the mouse" which also relies on `MotionNotify`. We may need to rethink the handling of grabs to sort everything out. Perhaps moving it to `Viewport`, which is really where the grabs are needed.

CendioOssman:
More on the same theme; it seems we stop getting core events once we select XInput2 events. This breaks other widgets in the same window, e.g. the scroll bars.

CendioOssman:
I dug deep enough in to the xserver code and found that it does indeed not deliver core events if it can deliver XInput 2 ones. It even seems like if *any* client listens for XInput 2 on a window then core events get filtered. See `DeliverDeviceEvents()` in `events.c`.

So we'll have to solve this in vncviewer somehow.

aeneby:
> I found a bug with regard to pointer grabbing. Set up a single monitor full screen, then start dragging stuff in the session. Once the pointer leaves the window, the dragging stops. This is a regression.
>
> What used to happen was:

Is this still the desired behaviour, though? That a drag should continue even if the pointer leaves the window? The amount of work that's been done to ensure that the drag continues makes me think there must have been a specific reason for wanting this, but I can't think of one.

> I dug deep enough in to the xserver code and found that it does indeed not deliver core events if it can deliver XInput 2 ones.

Yeah I think that's right too. So I guess anything which relies on core events will have to be modified.

CendioOssman:
> Is this still the desired behaviour, though? That a drag should continue even if the pointer leaves the window? The amount of work that's been done to ensure that the drag continues makes me think there must have been a specific reason for wanting this, but I can't think of one.

Anything else is rather annoying, I'd say. But that is the default behaviour you get without grabs, so it should continue to work with grabs.

> Yeah I think that's right too. So I guess anything which relies on core events will have to be modified.

I'd very much like to avoid having to rely on a patched FLTK. We've been there before and it was a pain. So let's see if we can find some other way of getting this to work...

aeneby:
Things have changed a lot since this PR was created, so closing it and I will make a new one. Most of the comments/suggestions here have been addressed in the new revision anyway I think.

CendioOssman:
For reference, the new one is #844.

Issue #592, opened 2018-02-02
Failed to connect to Socket


DidemnumVex:
Hi I'm running TigerVNC on my centos server. I have 7 user accounts on the server but only ONE is not working.  It was previously working just fine.  Have tried connecting from multiple remote machines with vncviewer and restarted the vnc server and the physical server.

The error log from the failed account:

```
Xvnc TigerVNC 1.1.0 - built May 11 2016 13:01:55
Copyright (C) 1999-2011 TigerVNC Team and many others (see README.txt)
See http://www.tigervnc.org for information on TigerVNC.
Underlying X server release 11704000, The X.Org Foundation


Thu Feb  1 13:26:32 2018
 vncext:      VNC extension running!
 vncext:      Listening for VNC connections on all interface(s), port 5907
 vncext:      created VNC server for screen 0
Couldn't exec xfce-session: Permission denied
gnome-session[19706]: WARNING: Could not make bus activated clients aware of DISPLAY=:7 environment variable: Failed to connect to socket /tmp/dbus-FZpL5vLpNq: Connection refused
gnome-session[19706]: WARNING: Could not make bus activated clients aware of GNOME_DESKTOP_SESSION_ID=this-is-deprecated environment variable: Failed to connect to socket /tmp/dbus-FZpL5vLpNq: Connection refused
gnome-session[19706]: WARNING: Could not make bus activated clients aware of SESSION_MANAGER=local/unix:@/tmp/.ICE-unix/19706,unix/unix:/tmp/.ICE-unix/19706 environment variable: Failed to connect to socket /tmp/dbus-FZpL5vLpNq: Connection refused
```

Error log from working account:

```

Xvnc TigerVNC 1.1.0 - built May 11 2016 13:01:55
Copyright (C) 1999-2011 TigerVNC Team and many others (see README.txt)
See http://www.tigervnc.org for information on TigerVNC.
Underlying X server release 11704000, The X.Org Foundation


Thu Feb  1 13:26:11 2018
 vncext:      VNC extension running!
 vncext:      Listening for VNC connections on all interface(s), port 5901
 vncext:      created VNC server for screen 0
GNOME_KEYRING_SOCKET=/tmp/keyring-lnEHz3/socket
SSH_AUTH_SOCK=/tmp/keyring-lnEHz3/socket.ssh
GNOME_KEYRING_PID=17324
Cleaned 26 files 9 still live

[...]

PR #838, opened 2019-05-24
Modern, systemd compatible session startup

### LINKED ISSUE #684: TigerVNC server only getting black screen on connect
I have a production machine (recently built and many hours of configuration have been put into it), so I tried things on a different machine I had sitting around first. 

I tried the suggestion in https://github.com/TigerVNC/tigervnc/issues/668 and downloaded https://bintray.com/tigervnc/stable/download_file?file_path=tigervnc-1.9.0.x86_64.tar.gz unpacked and moved to /

On my testing machine, I was able to connect correctly. I repeated the same exact TigerVNC install process on my production machine, and when I connect I just get a black screen.

The text machine is Xubuntu 18.04, and the projection machine is Ubuntu server 18.04 with the Xubuntu desktop added on top (sudo apt-get install xubuntu-desktop) because I wanted to use the server's install wizard to configure RAID.

I did a bunch of searching and found a different xstartup suggested here instead of the default created the first time vncserver is run: https://wiki.archlinux.org/index.php/Vncserver
`#!/bin/sh
unset SESSION_MANAGER
unset DBUS_SESSION_BUS_ADDRESS
exec startxfce4`

(I'm not sure why the line breaks aren't showing ^)

I don't get black (I am able to see the desktop), but the log file has tons of errors (attached).

I'm assuming the default xstartup is fine (since it was on the other machine), and that I have something else going on. 

[machinename-1.log](https://github.com/TigerVNC/tigervnc/files/2215080/machinename-1.log)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61139551-tigervnc-server-only-getting-black-screen-on-connect?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github).
</bountysource-plugin>

CendioOssman:
This is a major overhaul of how `vncserver` works and how it starts session. This is needed because the old school way we do things is no longer supported by current systems and desktop environments.

This does a number of changes to the system:

 * Sessions are started via PAM. This is necessary to get a proper login setup by systemd, and to get the correct SELinux context (and probably a lot of other things)
 * Sessions are started from a service as otherwise PAM cannot function correctly. This adds a new requirement on systemd as that's the only service definition we now have.
 * Sessions are started via `Xsession` as every distribution has its own quirks in how a desktop environment needs to be started and we cannot duplicate that effort. Note that this means that `~/.vnc/xstartup` is no longer supported.
 * Command line arguments have been removed from `vncserver` as it's not a good interface when it is a service, and it was redundant with the configuration files

This should fix #684, and probably a lot of other issues related to starting sessions.


kewlfft:
is it expected to fix https://github.com/TigerVNC/tigervnc/issues/592 as well? The need to run `dbus-launch` to run the xfce session instead of letting systemd manage it and create a new dbus.

CendioOssman:
It should, yes. D-BUS should be launched by the system in that case and not require anything explicit from us.

grulja:
What is the status of this change? Does it require some extra work? Or does it just need to be properly tested?

CendioOssman:
It still requires a lot of testing and polish, but it should mostly work. I just haven't been able to find the time to work on it.

So for anyone who has the time, please test this on various distributions, both with and without systemd.

darkdragon-001:
How to start a new desktop session when `xstartup` is no longer supported?

Currently, I use `dbus-launch --exit-with-session gnome-session`. I use it on a headless system where I want to create a purely virtual display with TigerVNC.

navysurfer:
In your `~/.vnc/config` add the line `session=gnome`.  The session line looks at the `*.desktop` files located in '/usr/share/xsessions'  If you don't have a `config` file, just copy the one from `/etc/tigervnc/vncserver-config-defaults` and adjust accordingly.

CendioOssman:
This should be working now, so please go ahead and test.

I've done some builds here for those that are unable to build themselves:

https://www.cendio.com/~ossman/tigervnc_systemd/

darkdragon-001:
In my setup, vncserver now seems to require `xinit`, but after installing it everything works fine for me! Looking forward to see this merged!

grulja:
I'm now trying to make this work on RHEL 8. There seems to be a problem with a missing dbus session. When starting it while being logged into GNOME, then the service fails to start, because GNOME fails to start. This works when I don't have a graphical session (runlevel 3). It was explained to me that on runlevel 5 (logged to GNOME), we already have a user bus, but it is not suitable for vnc session, while in runlevel 3 it will be created by gnome-session when there is none.
Xinit is supposed to create one, but it uses:
```
 if [ -z "$DBUS_SESSION_BUS_ADDRESS" ]; then
    eval `dbus-launch --sh-syntax --exit-with-session`
 fi
```
which won't happen as `DBUS_SESSION_BUS_ADDRESS` is set by pam_systemd. 

Solution to this is to add `dbus-run-session` to the executed command. I verified it works by putting this into the GNOME desktop file located /usr/share/xsessions. I also had some issues with SElinux, but I haven't looked into it yet properly as I spent some time investigating this. I will submit a PR if you want once I verify everything works even with SElinux. For now I had SElinux disabled. Thank you for pushing this change.

johnsimcall:
> ... When starting it while being logged into GNOME, then the service fails to start, because GNOME fails to start. This works when I don't have a graphical session (runlevel 3). It was explained to me that on runlevel 5 (logged to GNOME), we already have a user bus, but it is not suitable for vnc session, while in runlevel 3 it will be created by gnome-session when there is none.

Trying to be helpful here and summarize the [conversation](https://github.com/TigerVNC/tigervnc/pull/1011#issuecomment-624591088) that followed.  **You can't have a user login more than once, at the same time.**

You can't log into GNOME (or whatever) via the system's console and then startup that same user's VNC service.  The service will just exit saying `systemd[1]: vncserver@:1.service: Succeeded.`  Also, if your user is logged in via VNC and then you try to login via the system's console, it will hang the both sessions for a while, and eventually close both sessions.  The system console will return to the GDM login screen, and the VNC service will report that `systemd[1]: vncserver@:1.service: Succeeded.`

Thinking out loud... it might be nice if the `vncserver@:.service` file had a `Restart=...` line to handle this situation.  I'll experiment with my own drop-in snippet in the mean time.

FredericHamel:
This also broke my workflow. Why is xstartup no longer supported?
Is there a reason both to completely remove the xstartup file?
It should be possible to have both xstartup and  modern, systemd compatible session startup.

CendioOssman:
We are aware that this unfortunately broke some working setups. The problem was that in too many cases the `xstartup` model simply doesn't work. And there are years of now incorrect examples on the internet on how to setup `xstartup`. So it needed to go to make sure most users would instead use this new, more reliable method.

If you are starting a desktop environment, then `xstartup` is likely the wrong method and you need the new one. If you are starting something else, then you are likely doing something custom and can probably call `Xvnc` directly without using `vncserver`.

PR #845, opened 2019-06-15
Added option to specify RFB version to be used

DarkMaguz:
I need an option to force a specific version of the RFB protocol for use with tools like [vncdotool](https://github.com/sibson/vncdotool).


CendioOssman:
This looks like a workaround for some broken server. Adding a configuration option for such a reason is something we'd really like to avoid.

What is the issue you are seeing? What is the bug?

DarkMaguz:
> This looks like a workaround for some broken server. Adding a configuration option for such a reason is something we'd really like to avoid.

You are absolutely right. It is written specifically for broken, legacy and unmaintained tools.
I thought it to be a feature that others might find useful.

PR #847, opened 2019-06-18
Add "None" as menuKey option.  Add "WindowTitle" parameter

paherbst:
Parameter option MenuKey=None will disable the popup menu.
Parameter option WindowTitle will set the name of the window.


CendioOssman:
The native viewer already has the option of having no menu key. There is no support for changing the window title though. I'm not sure what the use case is there?

paherbst:
> The native viewer already has the option of having no menu key. There is no support for changing the window title though. I'm not sure what the use case is there?

If the native viewer already has option of no menu key, then it would be worthwhile to have the option in the Java viewer as well.

As for window title changes, tightvnc and ssvnc (based on tight) allow for the title to change, although i think this behaviour was inherited from the toolkit they use (libXt will process -title argument).  It is helpful to give the windows names when managing multiple vnc windows on a desktop.  It also allows for custom window manager rules based on window names.

CendioOssman:
Hmm... I think the patch is very insufficient in that case. We update the title based on the desktop name we get from the server. So a fixed name isn't really possible. Right now the principle is "<remote desktop name> - TigerVNC". Perhaps we can add a suffix at the end.

That should solve visibility. Not sure about the window manager rules though. Can they be specified using wildcards?

paherbst:
> Hmm... I think the patch is very insufficient in that case. We update the title based on the desktop name we get from the server. So a fixed name isn't really possible. Right now the principle is " - TigerVNC". Perhaps we can add a suffix at the end.
> 
> That should solve visibility. Not sure about the window manager rules though. Can they be specified using wildcards?

The Java viewer currently isn't implemented with " - TigerVNC", but simply displays the VNC server's name.  This proposed patch will give the user the option to have a fixed title.  If they don't specify it, then behavior-as-was goes into effect.

It's an optional argument most people can ignore and get the default behavior.  It just offers the user an override.

astrand:
My thoughts: This PR should be split into 2 separate; one for the menuKey option, and one for the issue with WindowTitle. Wrt the latter, we should strive for a consistent behaviour between the native and Java viewer. I agree with that an option for overriding the title could make sense. One solution could be an printf option string, ie default to:

"%s - TigerVNC"

This would give maximum flexibility - users could either set the title to a fixed string, or something which includes the remote desktop name.

samhed:
Agreed with @astrand. If you remove the WindowTitle stuff from this PR, this should be quite easy to merge.

paherbst:
Sounds good.  I'll do so (hopefully) this week.  I agree they should've been two requests.

CendioOssman:
No update. Closing.

PR #849, opened 2019-07-11
Do not return returncode indicating error when listing sessions

grulja:
Running "vncserver -list" will always return 1, which usually indicates an error, but this call is harmless and 0 should be returned instead.


samhed:
Thanks for the patch

PR #852, opened 2019-07-18
vncserver script: add -noxorgportcheck option

snejok:
- Most Linux distributions starts Xorg with "-nolisten tcp"
  option, so Xorg is not listening on 6000+n port


CendioOssman:
The check really has to do with if we want the port or not. So we should really be looking at the arguments we have and see if the user has specified `-listen tcp`.

Ideally we should also check the Xorg version for the default assumption, but perhaps it is long ago now that we can assume `-nolisten tcp` as the default...

PR #863, opened 2019-08-24
fix too early remove of iterator in VNCServerST::removeSocket

gschwind:
I do not know if someone has already noticed it but in VNCServerST::removeSocket the iterator is removed too early, and used afterward.

This patch fix the issue.

Best regards


gschwind:
Hello,

You are correct but, in this case the return statement line 196 break the for loop before issuing a ci++

Thus the current code seems ok for me, maybe a litle bit miss leading.

Best regards.

CendioOssman:
Ah, right. So then we should be good to merge this. Thanks for your contribution.

PR #871, opened 2019-09-28
Add -FixedPointer parameter which prevents cursor bitmap updates

Levitator1:
I had problems with lost cursor bitmap updates, so I decided to add a switch which disables them altogether, keeping the default mouse cursor. The switch is called -FixedPointer. TODO: Update the man page. There is also a partial NetBeans project in ./vncviewer, for the purpose of code-browsing.

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -243,6 +243,8 @@ static const char * dotcursor_xpm[] = {
 void Viewport::setCursor(int width, int height, const Point& hotspot,
                          const rdr::U8* data)
 {
+  if(fixedPointer) return;
+    
   int i;
 
   if (cursor) {
--- a/vncviewer/nbproject/Makefile-Debug.mk
+++ b/vncviewer/nbproject/Makefile-Debug.mk
@@ -1,0 +1,167 @@
+#
+# Generated Makefile - do not edit!
+#
+# Edit the Makefile in the project folder instead (../Makefile). Each target
+# has a -pre and a -post target defined where you can add customized code.
+#
+# This makefile implements configuration specific macros and targets.
+
+
+# Environment
+MKDIR=mkdir
+CP=cp
+GREP=grep
+NM=nm
+CCADMIN=CCadmin
+RANLIB=ranlib
+CC=gcc
+CCC=g++
+CXX=g++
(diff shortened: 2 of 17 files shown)


samhed:
Hi @Levitator1, in general we want to avoid adding code to work around bugs in other software.

Have you looked further into why your cursor goes missing? What VNC server are you using?

You could also have a look at the option to TigerVNC's viewer called "DotWhenNoCursor". If enabled it will show a dot when the server doesn't send a cursor.

Levitator1:
Hi,

The problem is not that the cursor is missing. The trouble is that bitmap
updates are lost or missing, so you get the wrong cursor bitmap until the
next successful update. The server is tigervncserver, provided here:
https://packages.debian.org/buster/tigervnc-standalone-server

I don't know at what point the bitmap updates are getting lost.

CendioOssman:
In that case it's a bug that we should try to fix.

Have you tried using a different VNC client? We need to determine if this is a bug in the client or the server.

Could you give us details on how your system is set up so we can try to replicate it?

Levitator1:
Initially, I used the tigervncclient that comes with Debian Buster and it
displayed the problem. So, I recompiled the latest client from the tigervnc
repo, and it displays the issue as well. I have tried other types of
client, but the ones that I looked at did not display cursor updates at all
-- just the default pointer. I am launching the server with:

*vncserver -rfbport 5901 -SecurityTypes Plain -localhost*

The session type is XFCE.

I am launching the client with:

.*/vncviewer localhost:1 -NoJpeg -MenuKey F10 -SecurityTypes Plain*

The client machine is an arm64 Chromebook running Debian Stretch under
Chrome OS. The server is an amd64 desktop computer running Debian Buster.

Levitator1:
Correction: The client is Debian Stretch under Chrome OS.

CendioOssman:
Hmmm... Can't say we have much experience running the client on ChromeOS. Do you have any other client device you could test with so we can see if that is the issue?

What other vnc client did you test? Both RealVNC and TightVNC should show the cursor correctly.

CendioOssman:
Something does sound a bit off in that case. Please let us know how it goes once you've tested on a different machine.

CendioOssman:
No update. Closing.

Issue #878, opened 2019-10-10
Please add "High Resolution" flag to Mac info.plist


appleguru:
Please add:

```
	<key>NSPrincipalClass</key>
	<string>NSApplication</string>
	<key>NSHighResolutionCapable</key>
	<string>YES</string>

```

to the Mac build's info.plist so that it isn't forced to "Open in low resolution" and the interface looks much better on modern high DPI Macs.

UI scaling already seems to be correct at high dpi, so this should be the only change needed. Please see attached screenshots.
<img width="377" alt="info_high" src="https://user-images.githubusercontent.com/595829/66611517-d3a0f280-eb8c-11e9-981b-320b7a7596f3.png">
<img width="377" alt="info_low" src="https://user-images.githubusercontent.com/595829/66611518-d3a0f280-eb8c-11e9-9da9-84f262eca5f2.png">
<img width="562" alt="connect_high" src="https://user-images.githubusercontent.com/595829/66611519-d4398900-eb8c-11e9-9cbb-fbf16eede044.png">
<img width="562" alt="connect_low" src="https://user-images.githubusercontent.com/595829/66611520-d4398900-eb8c-11e9-898c-d1f0215e0a7e.png">

CendioOssman:
I'm afraid we've already tried this in 6c0181c6f7241eaa9b19417c1729af463677c434, but it was reverted in aceb4e7bf702d5c26bad9e646854d69b68ee4d5a because of performance and quality issues.

Someone would need to sort that out before we enable it again.

Please see this issues reported for downstream ThinLinc:

https://www.cendio.com/bugzilla/show_bug.cgi?id=6980
https://www.cendio.com/bugzilla/show_bug.cgi?id=7047

CendioOssman:
Also see #369 for a more general discussion.

appleguru:
Thanks; I think this might warrant some more investigation with 64bit builds; I have not seen any performance issues so far with the setting enabled. Will try to do some more in depth testing later.

theLMGN:
<img width="1552" alt="image" src="https://user-images.githubusercontent.com/12257112/105491959-18cc3380-5caf-11eb-82f4-e44286b0ba58.png">
The scrolling still pretends it's running at LowDPI

CendioOssman:
[...]

PR #879, opened 2019-10-10
Update mac info.plist to support high dpi displays

appleguru:
https://github.com/TigerVNC/tigervnc/issues/878


CendioOssman:
As mentioned on #878, this unfortunately has unacceptable side effect. So more work is needed than just turning this feature on.

PR #880, opened 2019-10-16
Polkit fix

navysurfer:
To enable polkit, added getty.target to After, and changed ExecStart to use /bin/sh.

This will enable the required password prompts in gnome when trying to perform elevated tasks.

### DIFF
--- a/contrib/systemd/user/vncserver@.service
+++ b/contrib/systemd/user/vncserver@.service
@@ -14,12 +14,12 @@
 
 [Unit]
 Description=Remote desktop service (VNC)
-After=syslog.target network.target
+After=syslog.target network.target getty.target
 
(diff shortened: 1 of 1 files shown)


navysurfer:
Seems to only work intermittently.  Continuing to troubleshoot...

CendioOssman:
Unfortunately I think we need to resolve #838 to get good polkit interaction.

navysurfer:
Thanks for this.  I've been racking my brain on it for several days.  A psuedo workaround for me right now is to ssh into the server before the service is called and polkit starts as expected.  No idea why it works(something with PAM authentication).  Looking at your pull seems like there was a lot more at play than just a couple of config files.

PR #882, opened 2019-10-18
xkb map exhaustion fix #804

sergey1369:
ogurets/tigervnc - dynamic map


sergey1369:
Sorry, I've made a mistake. :(

PR #887, opened 2019-10-22
Added CentOS 8 Packaging support

evanburnsdev:
Adding support to master branch for CentOS 8 packaging.

### DIFF
--- a/.travis/centos8/Dockerfile
+++ b/.travis/centos8/Dockerfile
@@ -3,19 +3,12 @@ FROM centos:latest
 RUN yum -y update
 
 RUN yum -y groupinstall 'Development Tools'
-RUN yum -y install dnf-utils --nogpgcheck
-RUN yum -y install wget --nogpgcheck
-RUN yum -y install sudo --nogpgcheck
+RUN yum -y install dnf-utils
+RUN yum -y install sudo
 
 RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
 
 RUN yum config-manager --set-enabled PowerTools
-
-RUN wget http://mirror.centos.org/centos/7/os/x86_64/Packages/xorg-x11-server-source-1.20.4-7.el7.noarch.rpm
-
-RUN yum -y install xorg-x11-server-source-1.20.4-7.el7.noarch.rpm
-
-RUN rm xorg-x11-server-source-1.20.4-7.el7.noarch.rpm
 
 RUN yum -y update
 
--- a/.travis/centos8/Dockerfile
+++ b/.travis/centos8/Dockerfile
@@ -1,0 +1,27 @@
+FROM centos:latest
+
+RUN yum -y update
+
+RUN yum -y groupinstall 'Development Tools'
+RUN yum -y install dnf-utils --nogpgcheck
+RUN yum -y install wget --nogpgcheck
+RUN yum -y install sudo --nogpgcheck
+
+RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+
+RUN yum config-manager --set-enabled PowerTools
+
+RUN wget http://mirror.centos.org/centos/7/os/x86_64/Packages/xorg-x11-server-source-1.20.4-7.el7.noarch.rpm
+
+RUN yum -y install xorg-x11-server-source-1.20.4-7.el7.noarch.rpm
+
+RUN rm xorg-x11-server-source-1.20.4-7.el7.noarch.rpm
+
+RUN yum -y update
+
+RUN useradd -s /bin/bash -m rpm
+RUN echo >> /etc/sudoers
+RUN echo "rpm ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
+
+USER rpm
+WORKDIR /home/rpm
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ CMakeCache.txt
 Makefile
 Makefile.in
 config.h
+build/*
+.travis/centos8/rpmbuild/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ dist: xenial
 language: minimal
 
 env:
-  - DOCKER=centos8
   - DOCKER=centos6
   - DOCKER=centos7
+  - DOCKER=centos8
   - DOCKER=trusty
   - DOCKER=xenial
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ dist: xenial
 language: minimal
 
 env:
+  - DOCKER=centos8
   - DOCKER=centos6
   - DOCKER=centos7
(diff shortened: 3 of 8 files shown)


CendioOssman:
Nice, thanks. There seems to be something odd with the commit list still though as I'm seeing a stray merge commit in there. Do you think you could have a look at that?

navysurfer:
You guys made a few commits since I put in the pull request.  I merged it with mine to keep it up to date.

navysurfer:
I can rebase it again if you like.  Just let me know what you need.

CendioOssman:
No merges on your end, please. We can handle that here when we bring it in. :)

(if you really need to update it, then please do a rebase rather than a merge)

navysurfer:
Okay.  I'll get it cleaned up tonight.

navysurfer:
Rolled back to before merge.  Cleaned up travis.yml again.  Hopefully appeasing the git gods.  Let me know if you need anything else.

navysurfer:
Had some weird problem with my git author settings incorrect on my local machine.  Needed to force push to fix the commit.

.travis/centos8/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+RUN yum -y groupinstall 'Development Tools'
+RUN yum -y install dnf-utils --nogpgcheck
+RUN yum -y install wget --nogpgcheck
+RUN yum -y install sudo --nogpgcheck
+
+RUN dnf -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+
+RUN yum config-manager --set-enabled PowerTools
+
+RUN wget http://mirror.centos.org/centos/7/os/x86_64/Packages/xorg-x11-server-source-1.20.4-7.el7.noarch.rpm
```
bphinz: I don't think that taking the xorg-x11-server-source rpm from el7 is the right thing to do

navysurfer:
Unfortunately,  that's the only way to do it, as it was never included in the el8 repo.  I've compiled it in several different environments, with and without selinux enabled and haven't had any problems.
I've also installed the built from package in cent8 with no problems.  Ultimately, the call rests with you guys.  I'm not at my desk right now, but I could maybe track down the xorg upstream repo and download that to compile against.

bphinz:
Yeah, I don't know why the xorg-x11-server-source isn't available via yum,
but it can be built from the xorg-x11-server src rpm (still a PITA because
there are additional pre-reqs also not available without rebuilding their
parent src rpms...).  I'd rather see that than pulling the rpm from a
different distro release.

navysurfer:
Ok. I'll take a look at this weekend.

navysurfer:
Well 14 hours later...

xorg-x11-server depends on libmdx-devel and egl-wayland-devel, neither of which are available on the el8 repo.

After compiling all 3 from source rpms and installing in order I was able to compile tigervnc against the installed xorg-x11-server-source without relying on the el7 repo.  It triples the build time in Travis CI, but that's about the only correct way of doing it that I could think of.

A shortcut I thought about was just extracting the tar.bz2 source file from the xorg source rpm and editing the tigervnc spec file to use that for building.  It would work, but there would be no way to check for the build dependency.  I decided not to take that route.

Hopefully this is what you need.  I'm pretty burned out and don't think I'll be able to do much more for a few weeks. :)  Let me know what you think.

CendioOssman:
Yeah, all of this magic with the Xorg source is not something we want to maintain. And it's important to get this right or libvnc.so won't work properly.

Red Hat are building and shipping TigerVNC as part of RHEL 8, so the necessary packages must be there somewhere. Have you looked at their spec?

navysurfer:
Yeah, the src(not xorg-server-source) rpms are available to compile against but the compiled packages are not.  @bphinz also stated that above.
The the spec file for the tigervnc src rpm has a BuildRequired: xorg-x11-server-source but that package is not available for install.   It must be compiled from source, which I have done.  I've included a screenshot of the part of the spec for the packaged tigervnc available along with a search for the package.  Using my phone right now so it may be difficult to read.
Screenshot_20191119-042401_JuiceSSH

CendioOssman:
So it sounds like Red Hat has made a mistake somewhere. We should start by reporting it to them and see if they can provide a fix.

navysurfer:
More likely, the package responsible for xorg-server didn't include server-source for whatever reason.  I'll start an issue report with them but with how slow they move on sure itll be awhile before they get it corrected.  The last commit I made is the only currently working solution until they include the required packages.  If it doesn't fit with what you guys want, feel free to close.  I wish you guys all the best.

bphinz:
I agree that we should just file a bug report with them and accept the
overhead for now.  I've tested the rpm and it builds fine once you get the
prereq BS figured out.  @navysurfer - if you file a bugzilla ticket can you
reference the ticket number here so that we can track it?

navysurfer:
Will do.  I'll get it done tonight when I get home from work.

navysurfer:
Bug report submitted [Here](https://bugzilla.redhat.com/show_bug.cgi?id=1774228)

grulja:
Tigervnc Red Hat mainter here. The reason why xorg-x11-server-source is not available is that it makes sense to have it available only in the buildroot (in Brew, which is our build system for RHEL packages). I don't think that package will become available, unless there is a customer case for that, which I doubt there is and will be any. I can try to ask around and see if I can at least get some response from someone.

CendioOssman:
Can brew's repositories be made available in that case? I really think you should make it possible for your users to rebuild any package in RHEL. They (or consultants helping them out) might need the ability to rebuild a package for debugging or customization.

CendioOssman:
Hang on, this is using CentOS, not RHEL. So if this is build system related then I guess what RHEL does isn't relevant?

(I still think that RHEL should also ship this though so that every package can be rebuilt by customers, but it isn't necessary for this PR)

evanburnsdev:
CentOS is built off of RHEL and is nearly identical with differences being in distro support and subscriptions.  When a package is updated through RHEL, CentOS follows.

evanburnsdev:
Jan, a customer case would be anyone that wishes to build tigervnc from source would be required to build xorg(including libmdx and egl-wayland) to get the server-source package installed to copy over to the unix/xserver directory for building.

Pierre, if the package won't become available, I may have a workaround.  After digging through the xorg-server spec, it seems the server-source is just the src.rpm copied over to the /usr/share directory with a few file changes...  I was able to update the tigervnc.spec to do that work and it builds just fine without requiring the server-source package.  It would allow us to build tigervnc using only packages available from the distro's repository without any external dependencies.

The downside of doing it this way, is it could get very confusing for someone wanting to build tigervnc from source without doing a great deal of research.  The BUILDING.txt would have to have a section dedicated just to CentOS 8, which I'm fine with doing.

Jan, having the server-source package included in the repo would prevent everything I listed above.  From what you said, it looks like you're not responsible for that package(Adam Jackson?), but maybe you might have some luck convincing him to take a look at it.

Sorry for the long-winded message. :)
Evan

evanburnsdev:
No movement in over a month.

PR #894, opened 2019-10-25
Don't show the context menu hint when no hotkey is chosen

samhed:
If "Menu key" is set to "None" the viewer will display a hint saying "Press  to open the context menu". I think it's better to just not show the hint in this case.


CendioOssman:
LGTM

I wonder if we should validate `menuKey` before displaying it though. But that's a different issue...

PR #898, opened 2019-11-07
fix build vncviewer bug

itfanr:


CendioOssman:
I'm afraid this is not a correct fix for your issue. TigerVNC doesn't use those libraries so they should not be included in our dependencies.

Most likely there is something wrong with one of your libraries. E.g. FLTK. How is your build set up?

itfanr:
@CendioOssman 
I built it on ubuntu1804. And I encountered some errors during link. The environment is broken by me, but I make it with the changes.

Thank you.

PR #926, opened 2019-12-28
macOS and clang compilation-related updates

radonish:
- macOS key event names that were deprecated have been updated
- clang 11 compilation-related changes


radonish:
My intention was to keep the macOS/clang compilation fixes separate; will create independent PR.

PR #927, opened 2019-12-28
Fix gnutls X.509 certificate exporting

radonish:
Addresses #915 - vncviewer was erroneously reporting an out of memory failure due to how it was using gnutls' certificate export


CendioOssman:
Very odd. That code has been there since the start. I'm not sure how it used to work. Must have been some quirk in older versions of GnuTLS.

Thank you for attempting to fix this. However we don't want to hard code a limit on certificates, so I've pushed a more minimal fix in dbad687182ae9093efaf096a069eeafc18b22973.

PR #928, opened 2019-12-28
macOS and clang compilation-related updates

radonish:
- Updated deprecated macOS key event types
- Addressed compilation issues with clang 11

### DIFF
--- a/common/rfb/JpegCompressor.h
+++ b/common/rfb/JpegCompressor.h
@@ -49,7 +49,7 @@ namespace rfb {
 
     inline rdr::U8* getstart() { return start; }
 
-    inline int overrun(int itemSize, int nItems) {
+    inline size_t overrun(size_t itemSize, size_t nItems) {
       return MemOutStream::overrun(itemSize, nItems);
     }
 
--- a/common/rfb/SSecurityTLS.h
+++ b/common/rfb/SSecurityTLS.h
@@ -60,7 +60,6 @@ namespace rfb {
     gnutls_certificate_credentials_t cert_cred;
     char *keyfile, *certfile;
 
-    int type;
     bool anon;
 
     rdr::InStream* tlsis;
--- a/vncviewer/cocoa.mm
+++ b/vncviewer/cocoa.mm
@@ -145,9 +145,9 @@ int cocoa_is_keyboard_event(const void *event)
   nsevent = (NSEvent*)event;
 
   switch ([nsevent type]) {
-  case NSKeyDown:
-  case NSKeyUp:
-  case NSFlagsChanged:
+  case NSEventTypeKeyDown:
+  case NSEventTypeKeyUp:
+  case NSEventTypeFlagsChanged:
     return 1;
   default:
     return 0;
@@ -160,10 +160,10 @@ int cocoa_is_key_press(const void *event)
 
   nsevent = (NSEvent*)event;
 
-  if ([nsevent type] == NSKeyDown)
+  if ([nsevent type] == NSEventTypeKeyDown)
     return 1;
 
-  if ([nsevent type] == NSFlagsChanged) {
+  if ([nsevent type] == NSEventTypeFlagsChanged) {
     UInt32 mask;
 
     // We don't see any event on release of CapsLock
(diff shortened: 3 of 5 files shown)


CendioOssman:
Thanks for your contribution. Could you split this in to one commit per issue with a description of why this change is needed? It will make it easier to review.

common/rfb/JpegCompressor.h
```diff
@@ -49,7 +49,7 @@ namespace rfb {
 
     inline rdr::U8* getstart() { return start; }
 
-    inline int overrun(int itemSize, int nItems) {
+    inline size_t overrun(size_t itemSize, size_t nItems) {
```
radonish: The implicit conversion upsets clang11's default configuration and stricter gcc options.

radonish:
@CendioOssman, done. I split them off - all were simply to make the compilation successful with clang11.

Thanks

vncviewer/cocoa.mm
```diff
@@ -145,9 +145,9 @@ int cocoa_is_keyboard_event(const void *event)
   nsevent = (NSEvent*)event;
 
   switch ([nsevent type]) {
-  case NSKeyDown:
-  case NSKeyUp:
-  case NSFlagsChanged:
+  case NSEventTypeKeyDown:
+  case NSEventTypeKeyUp:
+  case NSEventTypeFlagsChanged:
```
CendioOssman: What do these changes mean in terms of minimum macOS SDK version?

common/rfb/SSecurityTLS.h
```diff
@@ -60,7 +60,6 @@ namespace rfb {
     gnutls_certificate_credentials_t cert_cred;
     char *keyfile, *certfile;
 
-    int type;
```
CendioOssman: Change looks good, but you can drop the bit about clang in the commit message. We want to keep commit message titles short for formatting reasons.

vncviewer/i18n.h
```diff
@@ -24,10 +24,10 @@
 
 /* Need to tell gcc that pgettext() doesn't screw up format strings */
 #ifdef __GNUC__
-static const char *
+__attribute__ ((format_arg (3))) static const char *
 pgettext_aux (const char *domain,
               const char *msg_ctxt_id, const char *msgid,
-              int category) __attribute__ ((format_arg (3)));
+              int category);
```
CendioOssman: I don't quite understand this one. This code should be for gcc only. Does clang also declare `__GNUC__`?

And why does it only complain about this one? We have similar things in `Exception.h` and `LogWriter.h`.

vncviewer/cocoa.mm
radonish: macOS 10.12+

samhed: and currently our minimum requirements are macOS 10.6+ @CendioOssman?

CendioOssman: Yeah, that's a bit too new for us for now. Has Apple actually removed the old ones? Or is it just a warning?

radonish: Warning promoted to error by default clang11 behavior. 10.12 is a 2016 macOS release; what is the line for too new?

common/rfb/SSecurityTLS.h
radonish: Done.

vncviewer/i18n.h
radonish: Yes, clang declares` __GNUC__`. The attribute usage in `i18n.h`/`gettext.h` is different than the two files you reference in that `i18n.h` includes `gettext.h`, which defines `pgettext_aux`, causing a conflict.

vncviewer/cocoa.mm
CendioOssman: Right now we're making sure it compiles with the 10.6 SDK, and runs on 10.9+.

If clang errors on those I think we'll have to use `#ifdef`s.

vncviewer/i18n.h
CendioOssman: Ah, I see. I think these files come from upstream gettext. Have you checked if they have similar fixes? It's easier if we copy what they do in case we want to update things.

CendioOssman:
These issues got fixed via #1037.

PR #941, opened 2020-01-14
Screen refresh timer.

xekkoman:
Added new startup argument:
name: ForceScreenRefresh
value: int - seconds
Starts timer that will send screen refresh command, after time in seconds passed.

That can solve problem described in issue 922 and can be useful in other cases.


CendioOssman:
Thanks for trying to fix this, but unfortunately I think this is a bit inefficient. You'll be requesting a lot of data from the server just to check if the connection is alive.

I suggest you have a look at using fences. The aren't supported by most servers though, but for the case of a TigerVNC server they should work fine.

xekkoman:
CendioOssman can you explain, about what fences you are talking?

CendioOssman:
These:

https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#clientfence

It's a very short message that can be sent to ping the server. It's currently not used in the client, but if you look at older versions it was used to synchronize pixel format changes. Look at `CConn.cxx` in some older version.

xekkoman:
CendioOssman, thank you for explanation.

PR #944, opened 2020-01-16
Fix JIS Keyboard dead keys in macOS

informationsea:
Underscore key does not work properly in macOS Japanese keyboard. This patch fixes the problem.


CendioOssman:
Thank you for your contribution. Unfortunately these files are automatically generated from https://github.com/qemu/keycodemapdb.

Have you checked if there is a fix available there? If so then we can re-generate our files based on their new data.

informationsea:
Thank you for the review. This problem is not fixed in upstream database. I will try to fix this problem in upstream project, but it may takes long time to find the right keycode for all columns. Can I patch https://github.com/TigerVNC/tigervnc/blob/master/vncviewer/Viewport.cxx#L1116 as a workaround, or do I need to wait for an upstream patch?

CendioOssman:
Patching the header files as you've done in this PR here is probably the lesser evil in that case.

But I don't think it is a requirement to fill out all of those columns. I think you just need to find the appropriate match in the first column (the Linux codes). Everything you don't know you can just leave blank.

For TigerVNC though it would be nice if you could figure out the the X11 XKB name. That would make sure things work on all our platforms. :)

Are you sure about 0x5f though? I cannot find any source confirming that. 0x73 seems to be the suggested value.

informationsea:
I confirmed 0x73 is the correct keycode on the my linux desktop and fix the source code. Thank you for the suggesting.

CendioOssman:
Great.

How did it go suggesting this change to keycodemapdb?

CendioOssman:
No response. Closing this PR.

PR #951, opened 2020-02-01
xserver: add no-op input thread init function

lheckemann:
This allows Xvnc to build with xorg-server 1.20.7, which requires OS
layers to implement a ddxInputThreadInit function when configured with
--enable-input-thread (the default).

relevant xorg-server commit: e3f26605d85d987da434640f52646d728f1fe919


CendioOssman:
LGTM. Thanks.

Issue #805, opened 2019-03-11
Display uses 1/4 of screen while mouse uses whole screen coordinates


lukego:
**Describe the bug**
VNC is showing my desktop at 1/4 size and occupying only the bottom-left quadrant of my viewer window. The mouse seems to be operate in the full coordinate space. I think the VNC desktop pixels need to be scaled up by a factor of 2 to work correctly.

I am running tigervnc server 1.9.0 on Linux (NixOS) and tigervnc client 1.9.0 (macOS homebrew) on a Macbook laptop with Retina display. The problem does not seem to occur on my desktop Mac with non-Retina display.

**To Reproduce**

You can install the exact VNC setup that I am using via Docker like this:

```
docker run --rm -ti -p 0.0.0.0:5901:5901 studioproject/master vnc
```

and then run `vncviewer localhost:1` outside of Docker on the host.

**Expected behavior**

VNC desktop uses whole viewer window.

**Screenshots**

<img width="1136" alt="Screenshot 2019-03-11 08 22 06" src="https://user-images.githubusercontent.com/13791/54135670-7450e900-43d7-11e9-82ab-1be9522f4224.png">

**Client (please complete the following information):**
- macOS Mojave.
 - TigerVNC 1.9.0.
- Downloaded from homebrew.

**Server (please complete the following information):**

- OS: Alpine Linux from Docker and Nix to install VNC.
- VNC serer: TigerVNC 1.9.0.

samhed:
If I recall correctly this didn't happen to me when I ran the vncviewer on macOS with scaling.

I'm unfamiliar with homebrew, but I believe it is possible that they are building with a different version of FLTK. Could you check what version they used?

Lastly, how come you used a client from homebrew? Doesn't one uploaded on bintray for the 1.9.0 release work for you? https://bintray.com/tigervnc/stable/tigervnc/1.9.0

lukego:
Thanks for the tip @samhed. I have been using the homebrew version just because that is my default way to install software. I switched to the stable binary from bintray and this is now working as expected.

Here is what I can immediately see about the dependency versions of the homebrew version:

```
$ otool -L $(which vncviewer)
/usr/local/bin/vncviewer:
	/System/Library/Frameworks/Carbon.framework/Versions/A/Carbon (compatibility version 2.0.0, current version 158.0.0)
	/System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa (compatibility version 1.0.0, current version 23.0.0)
[...]

PR #959, opened 2020-02-07
Fix for macOS Catalina / FLTK HiDPI issue (#805)

d235j:
Looks like Bug #805 is not an FLTK bug, but rather due to TigerVNC drawing directly into the CoreGraphics context that is backing the FLTK window.

This may not be the cleanest fix, but it does work as expected.

If I merely remove the call that applies CGAffineTransformInvert() on the transformation matrix, I get an image flipped in the vertical, probably because of how macOS uses the lower left hand corner as origin.

### DIFF
--- a/vncviewer/Surface_OSX.cxx
+++ b/vncviewer/Surface_OSX.cxx
@@ -142,6 +142,14 @@ void Surface::draw(int src_x, int src_y, int x, int y, int w, int h)
   // matrix as otherwise we get a massive performance hit
   CGContextConcatCTM(fl_gc, CGAffineTransformInvert(CGContextGetCTM(fl_gc)));
 
+  // Scale the image as necessary for the screen. If this is a HiDPI screen,
+  // the CGContext will have increased dimensions compared to the FL_Window.
+
+  double sf_h = (CGBitmapContextGetHeight(fl_gc) / Fl_Window::current()->h());
+  double sf_w = (CGBitmapContextGetWidth(fl_gc) / Fl_Window::current()->w());
+
+  CGContextScaleCTM(fl_gc, sf_h, sf_w);
+
   // macOS Coordinates are from bottom left, not top left
   y = Fl_Window::current()->h() - (y + h);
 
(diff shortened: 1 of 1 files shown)


d235j:
To add — the suggested workaround of using an older FLTK version does not work for me — all windows appear empty.

CendioOssman:
Nice, thanks. We'll have a look at this here and hopefully close this issue.

d235j:
It might be worth looking into why TigerVNC bypasses the FLTK drawing functions and blits into the underlying buffer directly...

CendioOssman:
The FLTK functions are unfortunately not efficient and flexible enough for our needs. We use low level drawing calls on all platforms, not just macOS.

samhed:
I tested both our own macOS build of the TigerVNC viewer and the one on bintray. I can't reproduce the problem. How are you building your viewer?

d235j:
@samhed 

I'm building FLTK as follows:
`./configure --prefix=/sources/fltk/p --enable-threads --enable-shared`

Then I'm building TigerVNC as follows:
`
mkdir build
cd build
PATH=$PATH:/usr/local/opt/gettext/bin cmake .. -DJPEG_INCLUDE_DIR=/usr/local/opt/libjpeg-turbo/include -DJPEG_LIBRARY=/usr/local/opt/libjpeg-turbo/lib/libjpeg.dylib -DFLTK_INCLUDE_DIR=/sources/fltk/p/include -DCMAKE_PREFIX_PATH="/usr/local/opt/gettext;/sources/p"
make -j9
`

This is on macOS 10.15.3, with latest Xcode.

What version of macOS are you  building on, and what build scripts are you using?

CendioOssman:
We (Cendio) use a cross compiler. But the official TigerVNC builds use something similar to your setup. @bphinz would know the details.

How do you create the app bundle though? Do you use our "dmg" target and our plist? Or do you have your own?

CendioOssman:
No response. Closing.

d235j:
@CendioOssman @samhed I seem to have missed this. I am not building the dmg, just running from the build directory to test. The fltk that is being used is the one in `/sources` that I have specified.

I would need more details regarding your build environment to be able to debug this.

CendioOssman:
I'm afraid you need to have a full app bundle to guarantee correct behaviour. Please try using `make dmg` and use that output.

Issue #188, opened 2015-08-18
Copying text to clipboard from Microsoft Edge causes unresponsiveness


jmontleon:
If I right click and copy text or use Ctrl-c to copy text from the Microsoft Edge web browsers address field TigerVNC and Edge become unresponsive. TigerVNC then Edge will become responsive again after several second.

I can copy text from the address field normally if TigerVNC is closed. I can even copy text from a page in Edge if TigerVNC is open. Copying addresses from Internet Explorers address bar/field causes no problem.

As far as I can tell it's just Edges address field when TigerVNC is open that causes them both to become unresponsive. I have accept and send clipboard disabled, but the same seems to happen with them enabled.

testworksau:
Sometimes I get strange clipboard behaviour from Excel too (where Excel is running on the same machine I'm running vncviewer.exe from). 

It can take two presses of CTRL-C for the text to be copied to the clipboard with the correct formatting.

Looks like another user of VNC had the same issue. My problem also went away when I closed the VNC Viewer down.

https://www.reddit.com/r/excel/comments/2aqcxq/why_do_i_have_to_ctrlc_a_cell_twice_to_get_it_to/ciygk0l

CendioOssman:
There is a new clipboard API available in Windows these days. Perhaps that will resolve this issue.

CendioOssman:
For reference, the new API is `AddClipboardFormatListener()` and friends.

PR #963, opened 2020-02-12
Windows Clipboard Polling

jua-mp:
Dear tigervnc-devs,

This is quite an ugly hack to try to avoid problems like in #188. At least, Excel seems to stop complaining about the clipboard being blocked.

It reads the clipboard data and sends it to the server while on FL_FOCUS instead of adding a clipboard listener with a callback. This is a Windows-only change.

You might be interested in something like this. Or not :smile:

Cheers,
Juanito

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -557,16 +557,10 @@ void Viewport::resize(int x, int y, int w, int h)
   Fl_Widget::resize(x, y, w, h);
 }
 
-
-int Viewport::handle(int event)
+void Viewport::sendClipboardData(const char* text, const size_t length)
 {
   char *filtered;
-  int buttonMask, wheelMask;
-  DownMap::const_iterator iter;
-
-  switch (event) {
-  case FL_PASTE:
-    filtered = convertLF(Fl::event_text(), Fl::event_length());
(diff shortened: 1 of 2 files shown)


samhed:
Which issue are you actually referring to? I can't find anything related to this in #88.

jua-mp:
Sorry, I meant #188. I just updated the description.

jua-mp:
Searching for the issue right now, I found this:

https://groups.google.com/forum/#!topic/tigervnc-users/jLEeOYLqSUs

The mentioned Excel issue is exactly what this apparently solves.

CendioOssman:
As you say, this is a bit hacky so I'm afraid it is not something we can merge.

Perhaps you can have a look at the new API mentioned in #188 and see if that resolves the issue?

jua-mp:
@CendioOssman, thanks for taking a look at this!

> As you say, this is a bit hacky so I'm afraid it is not something we can merge.

Fair enough.

> Perhaps you can have a look at the new API mentioned in #188 and see if that resolves the issue?

I had already seen that comment and looked for the new API, but wasn't able to find any infos on it. Do you happen to have a link somewhere?

I'm not sure I'll find the time to implement a "non-hacky" way, sorry! :frowning_face:

PR #1011, opened 2020-05-06
Make sure we start a dbus session

grulja:
As explained in https://github.com/TigerVNC/tigervnc/pull/838#issuecomment-624223381. We need to make sure a dbus session is started. I tested this on RHEL 8 and this makes it work for me while I'm logged in with the same user into GNOME.

### DIFF
--- a/unix/vncserver/vncserver.in
+++ b/unix/vncserver/vncserver.in
@@ -233,7 +233,7 @@ $ENV{XAUTHORITY} = $xauthorityFile;
 
 @cmd = ("xinit");
 
-push(@cmd, $Xsession, $session{'Exec'});
+push(@cmd, $Xsession, "dbus-run-session $session{'Exec'}");
 
 push(@cmd, '--');
 


CendioOssman:
Sorry, that is not our responsibility and would in fact break things.

Your use case is not supported by Red Hat or GNOME (or systemd) and I'm afraid it's not surprising it is breaking for you. Unfortunately you need to convince them to change before we can do anything.

grulja:
What do you mean my use case and not supported by Red Hat or GNOME. It's me who makes Tigervnc packages for RHEL and I want to be as close to upstream as possible, however currently it doesn't work for me as it is now, without this change. 

I was told yesterday on IRC from a collegue of mine:
```
[úterý 5. května 2020] [17:05:29 CEST] <halfline> no matter what session you pick, it should but run through dbus-run-session
[úterý 5. května 2020] [17:05:41 CEST] <halfline> pretty much any modern X session needs dbus
```

grulja:
And with https://github.com/TigerVNC/tigervnc/pull/1012 I'm able to run it with SELinux enabled.

grulja:
Now thinking, old xstartup way used to have `unset DBUS_SESSION_BUS_ADDRESS`. Wouldn't something like this make it work? That way xinit would start the dbus itself as this:
```
 if [ -z "$DBUS_SESSION_BUS_ADDRESS" ]; then
    eval `dbus-launch --sh-syntax --exit-with-session`
 fi
```
would pass now.

grulja:
> Now thinking, old xstartup way used to have `unset DBUS_SESSION_BUS_ADDRESS`. Wouldn't something like this make it work? That way xinit would start the dbus itself as this:
> 
> ```
>  if [ -z "$DBUS_SESSION_BUS_ADDRESS" ]; then
>     eval `dbus-launch --sh-syntax --exit-with-session`
>  fi
> ```
> 
> would pass now.

FYI: that works as well and I think it's the way to go. I'm just not sure where this should go, whether into the vncsession.c or into the vncserver perl script. I tested it quickly in the vncserver script so I don't have to build it again. Opinions?

CendioOssman:
I'm afraid you'll have to talk to your colleagues who are involved in session handling. Upstream has been very explicit that having multiple sessions as a user was a use case that was sacrificed in order to keep things simple. There is no longer such a thing as a session bus. It was replaced by a user bus. `DBUS_SESSION_BUS_ADDRESS` is deprecated and the bus must always be precent at `/run/user/<uid>/bus`.

We'd love for multiple sessions to be supported again, but we don't have the clout to influence the big players here. Until they change we will not cannot this use case either. 

(commands like `dbus-launch` and similar should really be removed from the distributions as they always do the wrong thing these days)

CendioOssman:
On a similar note, `pam_systemd.so` really should explicitly reject multiple sessions as right now users just get weird behaviour instead.

grulja:
Ok, I see, I was always trying to run the service for myself (same user). If I start it for a different user, then it works as expected. I will consider it as something we don't want to support.

darkdragon-001:
Maybe it makes sense to add this restriction to the documentation. By the way, which of the man pages listed on https://tigervnc.org/ describe the new systemd setup?

CendioOssman:
None so far. The home page gets updated when a new release is out, so there will be a new page there then.

PR #1012, opened 2020-05-06
[SELinux] Allow vnc_session_t type execute itself

wrabcak:
vncsession-start is running in SELinux vnc_session_t domain because of
"SELinuxContext=system_u:system_r:vnc_session_t:s0" option in systemd
vncserver@.service unit file. vncsession-start executing binary
vncsession with SELinux label/type vnc_session_t. This access was not
allowed in vncsession policy.

cc @grulja


CendioOssman:
Looks good. I wonder why I didn't see this in my testing...

PR #1026, opened 2020-05-21
Fix "vncserver -list" to have deterministic order and correct filter.

MarkMielke:
When listing sessions, use lexicographical order insead of directory order.

The filtering regular expression is vulnerable to misinterpretation, and
include a strange reference to ` $\ ` that is likely a typo:

       if ($file =~ /$host:(\d+)$\.pid/) {

It only works because ` $\ ` is normally empty, so it is acting like:

       if ($file =~ /$host:(\d+).pid/) {

This pattern is vulnerable if `$host` contains any regular expression
characters, including `.`, and also the `.` before pid is not literal.

This has been corrected to:

       if ($file =~ /^\Q$host\E:(\d+)\.pid$/) {

### DIFF
--- a/unix/vncserver
+++ b/unix/vncserver
@@ -699,8 +699,8 @@ sub List
     closedir(dir);
     print "\nTigerVNC server sessions:\n\n";
     print "X DISPLAY #\tPROCESS ID\n";
-    foreach my $file (@filelist) {
-	if ($file =~ /$host:(\d+)$\.pid/) {
+    foreach my $file (sort @filelist) {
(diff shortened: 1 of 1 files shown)


MarkMielke:
Only posting this against 1.10-branch, as the list function seems removed from master.

CendioOssman:
Thank you for your contribution. Unfortunately we're not planning on making any more releases on the 1.10 branch, and as you've noticed this particular feature is no longer present in the new code.

PR #1028, opened 2020-05-21
Fix "vncserver" display check to better identify available display nu…

MarkMielke:
…mbers.

There are a number of situations than can occur when the X
server is not shut down cleanly. These situations can leave
files around that "vncserver" has previously mis-identified
as evidence that the display number is still in use. These
changes reduce the occurrences of these false positives,
ensuring that "vncserver" handles these situations gracefully.

Instead of checking for existence of `/tmp/.X<n>-lock`, the
code will now extract the PID from this file and confirm that
a process exists with the same PID. This will eliminate
false positives in the case that this file references a PID
that no longer exists. The Xorg server does not have a problem
with fixing this file when it next starts up. It is only
important to avoid using the port if it is still in use.

Instead of checking for existence of `/tmp/.X11-unix/X<n>`, the
code will now attempt to connect to the socket to confirm that
there is a server process listening on this UNIX domain socket.
The Xorg server does not have a problem with fixing this file
when it next starts up. It is only important to avoid using
the UNIX domain socket if it is still in use.

The ordering of the code has changed, such that that check for
a lock file happens first. This should ensure that the fast
path does not require to bind() or connect() to any sockets
which can both have side effects. The slow path will still
do all checks, so there is no risk to this change.

The check for existence of `/usr/spool/sockets/X11/<n>` has
been removed. This file is only relevant on HP-UX, and
TigerVNC dropped support for HP-UX in commit 31e5aa3ddd2.

The section of code updated now requires Perl 5. This
removes legacy cruft that should not be necessary on any
supported platforms.

### DIFF
--- a/unix/vncserver
+++ b/unix/vncserver
@@ -522,41 +522,46 @@ sub GetDisplayNumber
 
 sub CheckDisplayNumber
 {
-    local ($n) = @_;
+    my($n) = @_;
 
-    socket(S, $AF_INET, $SOCK_STREAM, 0) || die "$prog: socket failed: $!\n";
-    eval 'setsockopt(S, &SOL_SOCKET, &SO_REUSEADDR, pack("l", 1))';
-    if (!bind(S, pack('S n x12', $AF_INET, 6000 + $n))) {
-	close(S);
-	return 0;
-    }
-    close(S);
+    use Socket;
 
-    socket(S, $AF_INET, $SOCK_STREAM, 0) || die "$prog: socket failed: $!\n";
-    eval 'setsockopt(S, &SOL_SOCKET, &SO_REUSEADDR, pack("l", 1))';
-    if (!bind(S, pack('S n x12', $AF_INET, 5900 + $n))) {
-	close(S);
-	return 0;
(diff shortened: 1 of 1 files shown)


MarkMielke:
Once merged to 1.10, I will propose comparable changes to the master branch.

MarkMielke:
I believe these fixes will address a number of startup failures with 1.10 and older versions. I know the systemd changes are intended to reset some of this. However, almost everyone is likely still using the "vncserver" script for the time being.

CendioOssman:
Thank you for your contribution. Unfortunately we're not planning on making any more releases on the 1.10 branch. If you can rebase this for the master branch then this would be very welcome.

MarkMielke:
Due to the time between releases, as well as due to the major change in functionality coming with 1.11 related to how Xvnc is launched (systemd, ...), I think there is a need for a 1.10 branch to be maintained with fixes that apply to 1.10. But, I already keep one myself, so this is really a question of whether anybody else is interested in it or not. I'll close this one, and I've posted #1066 with a rebase to master.

PR #1029, opened 2020-05-21
Change streams to be asynchronous

CendioOssman:
Major restructuring of how streams work. Neither input nor output streams are now blocking. This avoids stalling the rest of the client or server when a peer is slow or unresponsive.

Note that this puts an extra burden on users of streams to make sure they are allowed to do their work once the underlying transports are ready (e.g. monitoring fds).

This PR also contains a bunch of other fixes needed or otherwise related to streams.


CendioOssman:
This should in theory work, but needs more testing. Could someone test WinVNC for me?

CendioOssman:
Tested Xvnc, as well as vncviewer on Linux and Windows. A tweak for the bandwidth estimation is needed, but otherwise it seems to work well. So I'm going to merge this now.

PR #1031, opened 2020-05-23
Enhanced ComparingUpdateTracker to crop changed blocks

MarkMielke:
64x64 changed block can be large for fine changes such as cursor
movement and typing in terminal windows, or an update to a clock.

If the block can be efficiently cropped, this will reduce latency
and bandwidth. Every pixel cropped is a pixel less to analyze, encode,
transmit, and decode.

The previous code already detected the top of the change in order
to determine if the block had changed. However, it did not use
this information to reduce the size of the change rectangle, nor
did it calculate any of the other edges.

The new code introduces detection of the other edges, and uses
the information to build a reduced area change rectangle. This
has the additional effect of reducing the number of discrete pixel
values in the change block which may allow a more efficient
encoding algorithm to be selected.

As this section of code is performance sensitive, the method
of detecting the edges has been optimized to quickly fall back
to pessimistic values as soon as a single comparison fails on
each edge. In the case that full 64x64 block are changing,
there will be three extra comparisons per block.

In cases where the change rectangle can be reduced from 64x64,
the reduced size of the change rectangle represents reduced
effort to encode, transfer, and decode the contained pixels.

In the case of images with high frequency changes, which
specifically includes text, the lossy JPEG encoding can be
highly distorted, especially with JPEG level 6 or less. The
quick flash from a distorted JPEG to a lossless JPEG can
appear as a flickering to some people. This effect was more
obvious when the surrounding area is not expected to change,
but is being distorted anyways due to being part of the 64x64
blocking algorithm.

In the case of a user typing in a terminal window, this change
may commonly reduce the number of pixels updated with every
character typed from 4096 pixels (64x64) to 640 pixels (32x20)
or less.

### DIFF
--- a/common/rfb/ComparingUpdateTracker.cxx
+++ b/common/rfb/ComparingUpdateTracker.cxx
@@ -120,6 +120,10 @@ void ComparingUpdateTracker::compareRect(const Rect& r, Region* newChanged)
   rdr::U8* oldData = oldFb.getBufferRW(r, &oldStride);
   int oldStrideBytes = oldStride * bytesPerPixel;
 
+  // Used to efficiently crop the left and right of the change rectangle
+  int minCompareWidthInPixels = BLOCK_SIZE / 8;
+  int minCompareWidthInBytes = minCompareWidthInPixels * bytesPerPixel;
+
   for (int blockTop = r.tl.y; blockTop < r.br.y; blockTop += BLOCK_SIZE)
   {
     // Get a strip of the source buffer
@@ -139,19 +143,91 @@ void ComparingUpdateTracker::compareRect(const Rect& r, Region* newChanged)
       int blockRight = __rfbmin(blockLeft+BLOCK_SIZE, r.br.x);
       int blockWidthInBytes = (blockRight-blockLeft) * bytesPerPixel;
 
+      // Scan the block top to bottom, to identify the first row of change
       for (int y = blockTop; y < blockBottom; y++)
       {
         if (memcmp(oldPtr, newPtr, blockWidthInBytes) != 0)
         {
-          // A block has changed - copy the remainder to the oldFb
-          newChanged->assign_union(Region(Rect(blockLeft, blockTop,
-                                               blockRight, blockBottom)));
-          for (int y2 = y; y2 < blockBottom; y2++)
+          // Define the change rectangle using pessimistic values to start
+          int changeHeight = blockBottom - y;
+          int changeLeft = blockLeft;
+          int changeRight = blockRight;
+
+          // For every unchanged row at the bottom of the block, decrement change height
+          {
+            const rdr::U8* newRowPtr = newPtr + ((changeHeight - 1) * newStrideBytes);
+            const rdr::U8* oldRowPtr = oldPtr + ((changeHeight - 1) * oldStrideBytes);
+            while (changeHeight > 1 && memcmp(oldRowPtr, newRowPtr, blockWidthInBytes) == 0)
+            {
+              newRowPtr -= newStrideBytes;
+              oldRowPtr -= oldStrideBytes;
+
+              changeHeight--;
+            }
+          }
+
+          // For every unchanged column at the left of the block, increment change left
+          {
+            const rdr::U8* newColumnPtr = newPtr;
+            const rdr::U8* oldColumnPtr = oldPtr;
+            while (changeLeft + minCompareWidthInPixels < changeRight)
+            {
+              const rdr::U8* newRowPtr = newColumnPtr;
+              const rdr::U8* oldRowPtr = oldColumnPtr;
+              for (int row = 0; row < changeHeight; row++)
+              {
+                if (memcmp(oldRowPtr, newRowPtr, minCompareWidthInBytes) != 0)
+                  goto endOfChangeLeft;
+
+                newRowPtr += newStrideBytes;
+                oldRowPtr += oldStrideBytes;
+              }
+
+              newColumnPtr += minCompareWidthInBytes;
+              oldColumnPtr += minCompareWidthInBytes;
+
+              changeLeft += minCompareWidthInPixels;
+            }
+          }
+        endOfChangeLeft:
+
+          // For every unchanged column at the right of the block, decrement change right
+          {
+            const rdr::U8* newColumnPtr = newPtr + blockWidthInBytes;
+            const rdr::U8* oldColumnPtr = oldPtr + blockWidthInBytes;
+            while (changeLeft + minCompareWidthInPixels < changeRight)
+            {
+              newColumnPtr -= minCompareWidthInBytes;
+              oldColumnPtr -= minCompareWidthInBytes;
+
+              const rdr::U8* newRowPtr = newColumnPtr;
+              const rdr::U8* oldRowPtr = oldColumnPtr;
+              for (int row = 0; row < changeHeight; row++)
+              {
+                if (memcmp(oldRowPtr, newRowPtr, minCompareWidthInBytes) != 0)
+                  goto endOfChangeRight;
+
+                newRowPtr += newStrideBytes;
+                oldRowPtr += oldStrideBytes;
+              }
+
+              changeRight -= minCompareWidthInPixels;
+            }
+          }
+        endOfChangeRight:
+
+          // Block change extends from (changeLeft, y) to (changeRight, y + changeHeight)
+          newChanged->assign_union(Region(Rect(changeLeft, y, changeRight, y + changeHeight)));
+
+          // Copy the change from fb to oldFb to allow future changes to be identified
+          for (int row = 0; row < changeHeight; row++)
           {
             memcpy(oldPtr, newPtr, blockWidthInBytes);
             newPtr += newStrideBytes;
             oldPtr += oldStrideBytes;
           }
+
+          // No further processing is required for this block
           break;
         }
 


MarkMielke:
In terms of the visual artifact that I described here: https://groups.google.com/d/msg/tigervnc-users/Pq3ziWLsLvc/VfZ6Tye_BwAJ

This fix doesn't fully resolve the problem. However, upon understanding what the real problem is, this was the first obvious fix to apply that would provide benefit to many scenarios beyond the I one I was facing, while also improving and simplifying the problem for me. With this change, the visual artifact is reduced from 1, 2, or 4 (depending upon update alignment) full 64x64 blocks, to narrow strips. Visually, it means that the artifact was previously fairly large on both X and Y axis, but after this change, the artifact is typically less than pixels tall, and 64 pixels wide. It means the artifact is still visible left and right, but not above and below. With JPEG level 8, the artifact is very nearly invisible now, however if AutoSelect reduces to JPEG level 6, it is still perceptible. There are a few other ideas I have for improving this further, but I want to get this first change out of the way first.

CendioOssman:
Unfortunately this is a rather performance sensitive part of the code. Have you done some performance testing under different scenarios to see what the cost of this is? We don't want to trade one issue for another.

MarkMielke:
Yes. I tried to resurrect the original vncbenchtools and used the data sets from it, and was unable to see a difference. However, I wasn't sure that the tests were really valid for my use case either, and was going to think about it some more - but decided to post this patch anyways. I have also deployed it to a few hundred people over the weekend for additional soaking, with no concerns raised.

The change itself is specifically designed to both minimize overhead in the worst case, and benefit in non-worst cases. First, it only activates the new code if a change block is detected, so it has no real overhead for unchanged regions. For changed regions, here are the scenarios:

1. The entire block has changed. This is the worst case scenario, where it adds one check to see if the last row has also changed with a calculation and a memcmp(). This is close to zero overhead on any modern machine.
2. The top of the block has not changed, but the entire bottom has changed. In this case, it will add one check to see if the last row has also changed with a calculation and a memcmp(). But, it will be a savings, because the block sent to be encoding will start at the first line of change, not at the first line of the block. The more lines that didn't change, the more savings.
3. The bottom of the block has not changed, but the entire top has changed. In this case, it will add an additional calculation and memcmp() per row that didn't change. But, it will be a savings, because the block send to be analyzed/encoded/transferred/decoded will end at the last line of change, not at the end of the block. The more lines that didn't change, the more savings.
4. The top and bottom has not changed, but a portion in the middle has changed. This is similar result of 2. and 3. combined.

It's fairly minor - but beyond just not analyze/encode/transfer/decode, it also reduces the memcpy() to only the rows that have changed.

I have a more changes to make after this, but starting with the simple one. The general summary here is that the lossless refresh code has made a few elements like this much more obvious, to the point that I can see them as visible "shimmer", and the fixes will not just eliminate the visual artifacts, but also make the existing code more efficient both in CPU, and in network bandwidth.

In this case, there is often a double savings, as it reduces both the lossy JPEG update which is usually sent initially, and it also reduces the lossless refresh JPEG update that is sent a little later. This reduces CPU, and it reduces network bandwidth. I don't think the vncbenchtools properly captures this aspect, as it doesn't simulate an interactive user with short pauses between actions. It just streams all the RFB changes quickly without any gaps.

MarkMielke:
Now that I understand what is going on - I can also report that the "shimmer" doesn't need a slow VPN to reproduce. It can even happen over Gbit/s links. To see it yourself, try something like the following:

1. Turn off AutoSelect. AutoSelect is the variable that was making it come and go for me, making it difficult to reproduce.
2. Set CompressionLevel to 2. This will enable the ComparingUpdateTracker.
3. Set the JPEG level to 4. In my case - JPEG level 8 is nearly imperceptible and the default with AutoSelect, but it would sometimes drop to JPEG level 6 due to network performance. For me, it is plainly visible with 6, but I'm suggesting setting it to 4 to help with anybody who can't see it at 6.
4. Have one terminal with text in it. It doesn't matter what the text is, as long as it is not solid rectangles. It should be noisy.
5. Have another window - perhaps another terminal, offset to the right of the first terminal window. It should be overlapping, but text from the first terminal should be visible to the left.
6. Now drag the other window up and down.

The text to the left of this window should now start to "shimmer" very obviously. This is because it is first sending a lossy JPEG, where JPEG is not really appropriate for images with high frequencies such as text, and the lossless JPEG which reintroduces the high frequencies. Basically the characters get speckles added all over due to the high frequencies getting removed from the image, and then the speckles get cleared up. Because the lossless refresh is 50ms delay, this makes it much more obvious for human perception as 1/20th of a second or more should be visible in areas with high contrast such as text.

If you can't see it - try ensuring the first terminal has a white solid background, and dark text, as this contrast should make it even more obvious.

This first fix, has the effect of improving overall performance for these very common case (at least for us) of working in a terminal (or similar application), and as a side effect, reducing the chance and severity of the visual artifact. It does this by trimming the rectangle where the artifact might occur down to the minimum vertical impact, although for performance reasons it still leaves the horizontal impact as-is.

My next changes will likely address the horizontal, and also one or more fixes to the lossless refresh code and the code that chooses JPEG as the best choice for this high frequency use case. But, first trimming the change rectangle down is important to make sure that the problem being solved is the correct one. The problem shouldn't be "how to losslessly repair an unnecessary set of lossy changes sent", which is what the current implementation unfortunately degrades to, and what is the underlying cause to the "shimmer". :-(

Also: If you want to see things go REALLY bad with the lossless refresh code, set CompressionLevel = 1 above, and make a full screen terminal. Now start moving your cursor around, or using a pulsing cursor, or do basically anything. You will see the entire screen become speckled with JPEG artifacts, and then clear up with every minor screen update. This is hugely expensive for CPU and network, with unclear benefit of lossless refresh. :-(

MarkMielke:
Above I talked about "speckled with JPEG artifacts". This is mostly the impact for black vs white text. I should mention that the effect is more prounounced if you also add color. So, what makes my terminal including "ls" output, and Emacs even worse - is that I use color coding for syntax and different file types (as I expect most people do these days). This adds more gradients of colors, and causes something worse than speckling... the Hue, or probably the Saturation of the colours changes. The colours change from what they should be, to a higher or lower frequency due to JPEG quantization due to elimination of high frequencies. This means that not only are there random speckles all around the characters, but the font strokes themselves shimmer between the colour they should be, and they colour they are with lossy JPEG encoding on coloured text.

CendioOssman:
Since this is mostly about CPU usage and not compression ratio, I'd recommend using `perf` to see if there are any significant changes.

Use GNOME 3 as the desktop environment, as that really tortures the comparison code. Then put together some test case you can repeat somewhat identical. Then do `perf record -p <Xvnc pid> --call-graph=dwarf` and perform your test case. `perf report` will then show you how large portion of CPU usage is because of the comparisons.

Run that test with the original code, and with the new, and see how much they differ. Hopefully the reduction in `memcpy()` offsets the extra `memcmp()` and CPU usage will be similar as before.

MarkMielke:
I'll try to come up with something scientific. However, I will note that this is not a comparison of memcpy() vs memcpy(). Although memcpy() happens in this loop, the real effort is in the size of the change rectangles, as these must first be encoded with a lossy JPEG, and later encoded with a lossless JPEG. So, it is really memcmp() vs JPEG encoding that we are talking about exchanging here.

In the ideal case, memcmp() eliminates almost all of the JPEG encoding work. In the worst case, there is an extra memcmp() per 64x64 block, and no savings on JPEG encoding. For regular every day updates, including a user typing in a terminal or in a github.com comment, I have found that it typically reduces the rectangle from 64x64 to 64x20, resulting in 1/3 as much JPEG lossy encoding, followed by 1/3 as much JPEG lossless encoding.

MarkMielke:
I haven't tested the latest iteration yet, nor had time to prove the CPU impact is positive. However, I noted that the new Region code in master was causing this change to be unmergeable, so I have rebase it.

Due to effort to prove the CPU impact is positive, I may add to this change prior to re-testing and posting results. The original intent was to get a relatively simple change with less concern.

In the mean time ... the original change has been widely used without any problems reported.

MarkMielke:
I simplified the code a bit and made the comments a lot better - may improve performance, although with modern compiler it shouldn't make a difference.

MarkMielke:
I added a first go at cropping the left and right, but added it as a second commit for now to separate the changes for review.

To minimize runtime overhead it only considers cropping from left to right in increments of BLOCK_SIZE/4 == 16 pixels. The approach is to first start from the left and crop 16 pixels at a time from the left until the first change is encountered. Then it will start from the right and crop 16 pixels at a time until the first change is encountered. The change rectangle that will be used for encoding, and transmitting is cropped, but the copy at the end is unchanged (as the objective isn't really to copy fewer bytes in memory, but to only encode and transfer the smallest area of change.

The biggest benefit here is regular horizontal typing, cursor movement, or cursor flashing in a terminal or form field. In this case, it will first have been cropped down to 64x20 or so, and then it will be cropped down to 32x20 or 16x20. The pixels to encode and transfer will have dropped from 64x64 = 4096, to 32x20 = 640 which would normally be encoded and sent with every keystroke. This further improves the odds of using a lossless encoding up front due to the smaller rectangle and lower number of colour values to encode.

I am using this change myself to confirm no issues (performance or bugs) and if no issues or further improvements, will report back with some sort of believable benchmark. I may also collapse the two changes into one, unless you would prefer to accept the first one on its own, and the second one later.

MarkMielke:
Initial results for perf, watching the same YouTube clip filling a majority of the screen...

For both changes:

```
21.08%     0.43%  Xvnc     Xvnc                   [.] rfb::ComparingUpdateTracker::compareRect
20.65% rfb::ComparingUpdateTracker::compareRect
10.17% __memcpy_ssse3_back
8.69% __memcmp_sse4_1
1.30% rfb::Region::setOrderedRects                                                                                                                                         
```

But, this result seemed suspicious against the others, so I redid it at the end and got:

```
26.97%     0.72%  Xvnc     Xvnc                   [.] rfb::ComparingUpdateTracker::compareRect
26.25% rfb::ComparingUpdateTracker::compareRect
19.86% __memcpy_ssse3_back
5.19% __memcmp_sse4_1
0.71% rfb::Region::setOrderedRects                                      
```

For just the height crop (although the original version of it from last month):

```
29.25%     0.96%  Xvnc     Xvnc                   [.] rfb::ComparingUpdateTracker::compareRect
28.29% rfb::ComparingUpdateTracker::compareRect
21.97% __memcpy_ssse3_back
5.07% __memcmp_sse4_1
0.70% rfb::Region::setOrderedRects
```
Before both changes:

```
29.75%     1.14%  Xvnc     Xvnc                   [.] rfb::ComparingUpdateTracker::compareRect
28.61% rfb::ComparingUpdateTracker::compareRect
23.53% __memcpy_ssse3_back
3.88% __memcmp_sse4_1
0.50% rfb::Region::setOrderedRects
```

Overall, every test is very noisy without an exact match. But, I believe the above shows it isn't getting worse, and it might be getting better. Youtube video isn't the use case I am interested in, but I agree we don't want to make it worse, and it might make it better.

MarkMielke:
Ok. Having tested it a little more:

With normal auto select and VPN (bandwidth + latency) overhead reducing it to JPEG compress level 6 is no longer causing a percetible issue for me for my original use case. Yay!

If I reduce the JPEG compress level much lower, such as 2 or less, it's still plainly obvious that certain types of refreshes are causing lossy refresh of a larger block, followed by a lossless refresh of a smaller block. But, instead of being 64x64, 128x64, 64x128, or 128x128 (depending upon where the change is), it is now much smaller. An example here is that when doing cursor up or down in certain areas of colored text such that it affects two rows of characters (increasing the number of discrete color values), if the character being updated is effectively cropped by units of BLOCK_SIZE/4 effectively there is no visible artifact to the left or right, and it is likely doing a lossless encoding of the upper and lower portion. However, certain characters do not crop evenly by units of BLOCK_SIZE/4, and for these characters there could be up to 15 additional pixels on the left and right of the actual change, and these pixels can be seen as distorting (lossy) and then correcting (lossless).

Overall, I think I am proposing an improvement that has very low overhead (or may even reduce overhead!), that improves the visible behaviour of TigerVNC, and improves the network efficiency of TigerVNC.

Please let me know what is required to get both changes merged. Unless you prefer otherwise, I am going to squash the commits into one, align the variable names a little better, and possibly optimize it further (although I don't want it to be so hard to read that nobody understands it! :-( )

MarkMielke:
I optimized the code a little more for efficiency, and improved the readibility.

The latest iteration will crop from the left and right in units of 8 pixels (BLOCK_SIZE/8) instead of 16 (BLOCK_SIZE/4). This further reduces the visual artifact (there was one place I could still see it with JPEG level 6) and improves the efficiency of sending only the portions of the screen that actually changed to the client, without having any apparent impact on the performance for full screen updates such as the YouTube clip. Playing the same clip as earlier:

```
26.07%     0.95%  Xvnc     Xvnc                   [.] rfb::ComparingUpdateTracker::compareRect
25.12% rfb::ComparingUpdateTracker::compareRect
19.21% __memcpy_ssse3_back
4.88% __memcmp_sse4_1
0.66% rfb::Region::setOrderedRects
```
The numbers still look like noise to me more than proving an increase in efficiency. But, they don't seem to raise any alarms either.

MarkMielke:
FYI that with hundreds of people using this patch in the last two months, there has been zero complaints, and I've found a broad set of places where it has been a huge clearly visible improvement in performance. The essence of it is to not encode, send, and often re-send (lossless refresh) changes that never changed to begin with.

samhed:
Thank you for this contribution @MarkMielke! And thanks for taking the time to explain and test it as well.

Above you wrote:
>The entire block has changed. This is the worst case scenario, where it
adds one check to see if the last row has also changed with a
calculation and a memcmp(). This is close to zero overhead on any
modern machine.

When I read this I instantly think of video playback or something similar where large sections of the screen change constantly. I see that you measured the performance when running a youtube video - how come we don't see a slight performance deterioration in this case you think?

MarkMielke:
My conjecture:

The cost is not really the memcmp(). CPU are fast. memcmp() is fast. To believe that memcmp() is slow, requires presuming that the CPU is slow, which we should not be presuming.

The cost *is* likely related to cache lines, and the time required to load data from RAM into cache. That is, a cache hit will result in a very fast memcmp(), and a cache miss will result in a very slow memcmp(). The memcmp() itself is the least significant factor. It is merely the trigger for an operation that might be expensive.

The code as I've implemented checks all four sides:

1. Top - already being checked by the previous code, so no change
2. Bottom - new - however, is it really new? When the bottom row is eventually encoded, it would have to be loaded into cache, so likely - the new code is just pulling it into cache earlier, for what basically amounts to a net zero change +/- noise
3. Left and Right - as the left and right check are really checking the Left and Right of the Top, and the Top is likely already in cache from the above check, these memcmp() are likely also very fast.

In the case that the video has resulted in the entire block changing - there is some theoretical overhead above, which upon re-evaluating in the context of cache lines, and upon evaluation of real life samples, seems to be very difficult to measure.

In the case that the video has *not* changed every block on the screen, but some blocks are entirely changed, some blocks are partially changed, and some blocks are not changed at all, the new code will be a net gain, as the blocks that are not fully changed will be cropped down to the region of change (with rectangular dimensions), and less data will be encoded, transferred to the client, and decoded on the client.

There is also the case of "full screen video" vs "video in a portion of the screen only", and this then depends upon the alignment of the 64x64 change blocks with the video as to what the impact is. For example, let us say that the video is inside Firefox or Chrome canvas, with non-video elements (Youtube logos and interfaces?) as well as a video element. If the video exact aligns (x1, y1, x2, y2) with 64x64 blocks, such that x1 % 64 == 0, y1 % 64 == 0, x2 % 64 == 0, y2 % 64 == 0, and the video is changing every block with every frame, then this worst case scenario might be real (within the limits observed and theorized about above). But, what if it's not? What if x1 % 64 == 32, and y1 % 64 == 28? At this point: 1) We are marking the unchanged region outside the video as "changed", which is causing it to be encoded, sent, and decoded, and 2) JPEG artifacts such as my original concern become visible with lower JPEG settings. For 2), this is because there is a high frequency blip between the area that is not changing, outside the video, and the portion that is changing, inside the video. Because of how JPEG works, and because the portion outside the video is being encoded *with* the video, this will result in artifacts as the portion in the video will affect the portion outside the video, and the portion outside the video will affect the portion inside the video. At some later time, "JPEG lossy refresh" will come back and clean up the portion outside the video, but this delay between it being "speckled" and it being restored to "unspeckled" is sometimes visible and distracting.

The above conjecture is convincing to me that the video case of every 64x64 block changing every frame is not significantly effected. Maybe it is, maybe it isn't, but I doubt you'll be able to prove it one way or the other - just as I wasn't able to prove it one way or the other. It looks like noise to me, and it probably depends upon exactly what is in the video, and where exactly the video aligns with the 64x64 blocks. Most videos are probably not like this, anyways - but are either a subset of the screen, or not every pixel is changing with every frame.

MarkMielke:
Also, I should have mentioned this point which ties the original performance concern, to caching. I think there has been a presumption that this code is performance sensitive due to operations performed within it. However, if you look at the pixel analysis code, and the encoding code, including the JPEG encoding - these are much more complex than a simple memcmp(). To wonder that adding a memcmp() would create a problem - should really start with "well, why doesn't JPEG or other encoding cause a problem, which does a lot more than memcmp()?"

I think the original performance sensitivity is probably also about cache lines. With this interpretation - the question isn't primarily about "how complex are the operations on each block", but should be "how big of a block will fit entirely in cache for computers that we are targeting?"

It may be that in 2010 and before, a block size of 16x16 was the right size, if the cache size was much lower. At some point this was increased to 64x64, because this allows us to take advantage of the larger cache sizes, and perform more efficient analysis of the block, such as memcmp() of 64x4=256 bytes at a time, instead of memcmp() of 16x4=64 bytes at a time.

I'm not saying we can do anything we please, and it will always be dwarfed by the time to load the 64x64 block into cache. But, I think the change I am proposing is a good compromise between a minimum number of checks to determine if optimization is possible, and if optimization is posisble, then reducing the total amount of work required which results in better performance overall.

MarkMielke:
In explaining my thinking, I noticed:

If the entire screen fits into cache (doubtful? not sure?), then ComparingUpdateTracker might have the bulk of the cost, as it is responsible for loading the frame into cache before a later step runs the encoder and benefits from ComparingUpdateTracker running before it. However,

If the entire screen does not fit into cache, then ComparingUpdateTracker might be loading the region it is inspecting into cache, pushing out other data from cache, and then later the change regions identified are re-loaded into cache, also each pushing out other data from cache.

I think the second explanation is the most likely and relevant, and this makes me wonder if integrating the ComparingUpdatetracker with the EncodeManager, such that the change blocks are iteratively identified and encoded before moving onto the next block, would eliminate the need to load the fb into cache a second time.

This type of change might be intrusive, and the existing code seems to work pretty well for many real-life cases including video, so there might not be an appetite or requirement to make this change. But, I'm just noting that if you explain the overhead of ComparingUpdateTracker in terms of cache lines, and you explain the overhead of EncodeManager and the encoders in terms of cache lines, it seems unfortunate that they don't overlap in their execution path. :-)

CendioOssman:
I've done some testing here as well, and unfortunately GNOME 3 is no longer a good test case. It seems they (or Mesa) have fixed whatever bug they had and they now report only the areas that actually changed.

Fortunately Firefox still re-renders the entire window for many animated pages, so it was useful as an alternative.

With that in place I can see marginal overall change in how much time the comparison takes up (~9% in my test). I can see that `memcmp()` is now a larger proportion compared to `memcpy()` before, but that's expected.

Oddly enough I'm seeing more time spent in socket functions. But it's not readily apparent why. It might just be more throughput. So I'm not going to let that block this change.

The code also looks good, so I'm going to go ahead and merge this. Thank you for your work in writing this code and testing things out.

PR #1034, opened 2020-05-26
[SELinux] Allow vnc session create ~/.vnc

wrabcak:
Allow vnc session labeled with SELinux vnc_session_t domain to create
directory in user homedir ~/.vnc labeled with SELinux label xdm_home_t.

Because also process vncpasswd creates ~/.vnc directory, file transition that
unconfined_t SELinux domain (label for the process) can create the
directory with the same label needs to be added.

Signed-off-by: Lukas Vrabec <lvrabec@redhat.com>

### DIFF
--- a/unix/vncserver/selinux/vncsession.fc
+++ b/unix/vncserver/selinux/vncsession.fc
@@ -17,6 +17,9 @@
 #  USA.
 #
 
+HOME_DIR/\.vnc(/.*)?      gen_context(system_u:object_r:xdm_home_t,s0)
+HOME_ROOT/\.vnc(/.*)?      gen_context(system_u:object_r:xdm_home_t,s0)
+
 /usr/sbin/vncsession			--	gen_context(system_u:object_r:vnc_session_exec_t,s0)
 /usr/libexec/vncsession-start		--	gen_context(system_u:object_r:vnc_session_exec_t,s0)
 
--- a/unix/vncserver/selinux/vncsession.te
+++ b/unix/vncserver/selinux/vncsession.te
@@ -19,6 +19,11 @@
 
 policy_module(vncsession, 1.0.0);
 
+gen_require(`
+    attribute userdomain;
+    type xdm_home_t;
+')
+
 type vnc_session_exec_t;
 corecmd_executable_file(vnc_session_exec_t)
 type vnc_session_t;
@@ -41,6 +46,16 @@ allow vnc_session_t self:capability { kill chown dac_override dac_read_search fo
 allow vnc_session_t self:process { getcap setsched setexec setrlimit };
 allow vnc_session_t self:fifo_file rw_fifo_file_perms;
 
+manage_files_pattern(vnc_session_t, xdm_home_t, xdm_home_t)
+manage_fifo_files_pattern(vnc_session_t, xdm_home_t, xdm_home_t)
+manage_sock_files_pattern(vnc_session_t, xdm_home_t, xdm_home_t)
+manage_lnk_files_pattern(vnc_session_t, xdm_home_t, xdm_home_t)
+userdom_user_home_dir_filetrans(vnc_session_t, xdm_home_t, dir, ".vnc")
+userdom_admin_home_dir_filetrans(vnc_session_t, xdm_home_t, dir, ".vnc")
+
+userdom_admin_home_dir_filetrans(userdomain, xdm_home_t, dir, ".vnc")
+userdom_user_home_dir_filetrans(userdomain, xdm_home_t, dir, ".vnc")
+
 miscfiles_read_localization(vnc_session_t)
 
 kernel_read_kernel_sysctls(vnc_session_t)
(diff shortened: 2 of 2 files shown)


wrabcak:
CC @grulja

CendioOssman:
Is this a recent change? Using `vncpasswd` on my Fedora 31 results in the type `user_home_t`.

grulja:
> Is this a recent change? Using `vncpasswd` on my Fedora 31 results in the type `user_home_t`.

This is on RHEL 8. For a regular user it is `user_home_t`, but this change was made so it works also for `root` user. 

Note: I know it's not a good idea to run a vncserver as `root`, but some people do it anyway.

CendioOssman:
Shouldn't these changes be just for `admin_home_t` in that case?

wrabcak:
This solution allows write only to xdm_home_t which is label for ~/.vnc or /root/.vnc. The solution should be "more secure" because there is no need that vnc session to manage whole /root/ dir .

CendioOssman:
My point was that it sounds like ~/.vnc should be user_home_t for normal users, but xdm_home_t for root. But I can't see that in the suggested rules.

wrabcak:
@CendioOssman, 
No, in both situations ~/.vnc should be xmd_home_t. It's handled in vncsession.fc file on lines 20 and 21 and in vncsession.te lines 53 and 54.

CendioOssman:
I'm afraid I'm still missing something.

I read the initial description as `vncpasswd` already creating `~/.vnc` as `xdm_home_t`, and this PR is to make `vncserver` do the same?

However when I try this on either RHEL 8 or Fedora 31, I get `user_home_t`:

```
[cendio@dhcp-254-219 ~]$ cat /etc/redhat-release 
Red Hat Enterprise Linux release 8.1 (Ootpa)
[cendio@dhcp-254-219 ~]$ rm -rf .vnc/
[cendio@dhcp-254-219 ~]$ vncpasswd
Password:
Verify:
Would you like to enter a view-only password (y/n)? n
A view-only password is not used
[cendio@dhcp-254-219 ~]$ ls -Zd .vnc
unconfined_u:object_r:user_home_t:s0 .vnc
```

grulja:
Using `vncpasswd`, which is the first thing users will run to create a password, will create `~/.vnc` as `user_home_t` for a regular user and `admin_home_t` for the `root` user. I think what we were working on with @wrabcak was to make it to use `xdg_home_t` instead for both regular users and the `root` user.

CendioOssman:
So this is for a future improvement in the reference policy?

Can we do this in a way that works with both old and new systems?

grulja:
I have tested it now (again), because before I had a slightly different version.

It does work for the root user, however it doesn't work for regular users, because it is missing this rule:
`userdom_user_home_dir_filetrans(unconfined_t, xdm_home_t, dir, ".vnc")`

wrabcak:
Good catch @grulja! PR updated.

grulja:
To summarize this change:
Current behaviour on Fedora 32 or RHEL 8 (without this patch) is that `/root/.vnc/` will have `admin_home_t` context and regular users with `$HOME/.vnc/` will have `user_home_t` context.

With this patch on RHEL 8 (or even Fedora 32 once this becomes available) both the root with  `/root/.vnc/` and regular users with `$HOME/.vnc/` will have `xdg_home_t` context.

CendioOssman:
Assuming that `vncsession` creates the directory though? Not if the user first runs `vncpasswd`?

grulja:
Tested now:
1) I removed $HOME/.vnc for both a regular user and the root user
2) Started /usr/libexec/vncsession-start :display_number (for each)
3) Running `ls -Zld $HOME/.vnc` for both returns **xdm_home_t**

CendioOssman:
But that is not the normal/only case. Users will be running `vncpasswd` first or their session won't be functional. How will that be handled? What do you get on your system if you run `rm -rf ~/.vnc; vncpasswd`?

grulja:
I've tested both variants:
1) Running vncsession-start as first, as mentioned in https://github.com/TigerVNC/tigervnc/pull/1034#issuecomment-655541483
2) Running vncpasswd as first, this is what I've been using all the time, mentioned for example here https://github.com/TigerVNC/tigervnc/pull/1034#issuecomment-646065045

In both cases the user (root or a regular user) didn't have `$HOME/.vnc` created.

CendioOssman:
You're right. I got a test machine up and running, and I can confirm even vncpasswd is affected. I'm not quite sure how though. I assume it is these two lines:

```m4
userdom_admin_home_dir_filetrans(unconfined_t, xdm_home_t, dir, ".vnc")
userdom_user_home_dir_filetrans(unconfined_t, xdm_home_t, dir, ".vnc")
```

If so, then this should be clearly marked as it doesn't really fit in here as the file is meant for vncserver and friends. Any bonus stuff should be easy to spot.

What about users that aren't `unconfined_t` then? Every domain should be affected by this transition rule. I hope there is a way to specify that?

wrabcak:
Hi @CendioOssman ,

If you want to allow this transition for all possible user domain you can use attribute `userdomain` instead of `unconfined_t` . 

userdomain attribute contains all these types: 
```
Type Attributes: 1
   attribute userdomain;
	auditadm_t
	dbadm_t
	guest_t
	logadm_t
	secadm_t
	staff_t
	staff_wine_t
	sysadm_t
	unconfined_t
	user_t
	user_wine_t
	webadm_t
	xguest_t
```

CendioOssman:
Seems more future proof IMO.

Could you do that change, and also make it more obvious this rule is not like the others? Then we can merge this.

grulja:
Hi @CendioOssman, the review has been updated as requested. Can we merge it now?

PR #1036, opened 2020-05-28
Send response for refused SetDesktopSize if the client is expecting one

matt335672:
Currently, if the user sets the `AcceptSetDesktopSize` parameter to 0, the client receives no response at all to a SetDesktopSize message.

If the client has previously received an ExtendedDesktopSize rectangle, this not compliant with [the community wiki](https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#setdesktopsize) specification.

This PR sends an ExtendedDesktopSize rectangle with (x,y) set to (1.1) "Resize is administratively prohibited" as a response to SetDesktopSize if AcceptSetDesktopSize is set to 0 and the client has previously requested the `ExtendedDesktopSize` pseudo-encoding.


CendioOssman:
Looks good. Thanks for finding this.

The same issue should exist for the access check just above as well. Do you think you could adjust your commit to fix them both?

matt335672:
How's that?

Let me know if you want me to squash.

Thanks for the positive response.

CendioOssman:
Nice, thanks. And yes, please squash them. And you can remove the encoding check. A client should not be sending us SetDesktopSize messages without proper support for responses.

matt335672:
Thanks for the clear guidance. Removing the encoding check makes for a more natural code flow.

CendioOssman:
Nice. Just one final code style issue; we tend to avoid {} for single line blocks. So if you can adjust that then I can merge this. :)

PR #1048, opened 2020-06-20
Fix division by zero exception in SSecurityPlain.

MarkMielke:
If using SSecurityPlain and the user specifies an empty username
and password, it will invoke InStream::checkNoWait(0) which will
cause a division by zero when calculating the number of available
items.
    
Enhance InStream::check() to behave properly when asked for
zero items, or zero sized items.
    
Add comments to InStream::check(), InStream::checkNoWait(),
and InStream::readBytes() to document expected behaviour
when requested to check or read zero items, or an item with
zero size.

### DIFF
--- a/common/rdr/InStream.h
+++ b/common/rdr/InStream.h
@@ -39,15 +39,19 @@ namespace rdr {
     // itemSize bytes.  Returns the number of items in the buffer (up to a
     // maximum of nItems).  If wait is false, then instead of blocking to wait
     // for the bytes, zero is returned if the bytes are not immediately
-    // available.
+    // available. If itemSize or nItems is zero, check() will return zero.
 
     inline size_t check(size_t itemSize, size_t nItems=1, bool wait=true)
     {
       size_t nAvail;
 
+      if (itemSize == 0 || nItems == 0)
+        return 0;
+
       if (itemSize > (size_t)(end - ptr))
         return overrun(itemSize, nItems, wait);
 
+      // itemSize cannot be zero at this point
       nAvail = (end - ptr) / itemSize;
       if (nAvail < nItems)
         return nAvail;
@@ -58,8 +62,12 @@ namespace rdr {
     // checkNoWait() tries to make sure that the given number of bytes can
     // be read without blocking.  It returns true if this is the case, false
     // otherwise.  The length must be "small" (less than the buffer size).
+    // If length is zero, checkNoWait() will return true.
 
-    inline bool checkNoWait(size_t length) { return check(length, 1, false)!=0; }
+    inline bool checkNoWait(size_t length)
+    {
+      return length == 0 || check(length, 1, false) > 0;
+    }
 
     // readU/SN() methods read unsigned and signed N-bit integers.
 
@@ -94,6 +102,7 @@ namespace rdr {
     }
 
     // readBytes() reads an exact number of bytes.
+    // If length is zero, readBytes() will return immediately.
 
     void readBytes(void* data, size_t length) {
       while (length > 0) {


MarkMielke:
I believe this issue has hit a few of our users - it finally hit me, and more recently hit another user. Without this change, the checkNoWait(0) as caused by the user clicking "Ok" without authentication always fails with "Floating point exception" and causes Xvnc to crash. To some degree, this makes it a security issue as any user without a password can crash any session that is using SSecurityPlain. I'm surprised it hasn't hit a lot more users!

I know CendioOssman has a non-blocking branch that may address this indirectly. However, that code isn't yet merged and I think this change is important enough to fix first. Also, it is a bit ambiguous what it means to assert that there is at least 0 bytes available. Should it return true or false? In the case of checkNoWait(), it is asking "give me at least 1 items of size 0", so if zero bytes are available should that be a that 1 item is available, or 0 item is available? The "Floating point exception" is exposing the flaw in this logic.

I think if no bytes are expected, it shouldn't call checkNoWait(0) or the non-blocking branch equivalent. So, this code should apply either way.

CendioOssman:
Thanks for finding this issue. I think this should be fixed inside `InStream::check()` though as any caller might do this. Could you have a look at that? Calling `check(0)` should be safe, as well as `readBytes(0)` or similar.

As for #1029, it will solve this. But the plan is to make a release before that, so this PR would be nice to get in to the release.

MarkMielke:
I did review both options, and determined that calling `checkNoWait(0)` was ambiguous, as it is unclear whether it should return true or false. For example, purpose `checkNoWait(0)` should be interpreted as `checkNoWait(1)`, given that an actual read() would block.. Essentially, the division by zero is surfacing as an amusing classical mathematic problems... What is 5 / 0 equal to? If the code was implemented differently, such as #1029, it wouldn't surface the problem any more, but there is still a problem in terms of communicating intent. Is the intent to ensure that a `read()` will not block? What will `read(0)` do?

So, the minimum change for me, was that unless the code wants to read bytes, it should not be checking for those bytes to be available in the first place. I thought this would avoid the question, and most clearly communicate design intent to the code reviewer.

But, given that it will fail with a division by zero today - perhaps there is argument to either do both, or document the special 0 case, and add a comment in the calling code that "if called with 0, we expect it will return true"? This just seemed messier to me.

Also, I think the same check that we want to read before calling check() would apply after #1029. The only reason it might not is if the special case of 0 is documented with specific behaviour, and it is clear enough to a reader that the case exists, and what will happen in the case.

MarkMielke:
Despite my comment about: "What will read(0) do?" 

I do acknowledge that the code does exactly this a few lines later, twice. So, I suppose it should be documented what it does. :-)

CendioOssman:
> I did review both options, and determined that calling `checkNoWait(0)` was ambiguous, as it is unclear whether it should return true or false. For example, perhaps `checkNoWait(0)` should be interpreted as `checkNoWait(1)`, given that an actual read() would block.. Essentially, the division by zero is surfacing as an amusing classical mathematic problems... What is 5 / 0 equal to? If the code was implemented differently, such as #1029, it wouldn't surface the problem any more, but there is still a problem in terms of communicating intent. Is the intent to ensure that a `read()` will not block? What will `read(0)` do?

The intent is indeed that `checkNoWait(x)` guarantees that `read(x)` will not block. And reading 0 bytes should always succeed without blocking, otherwise we have a bug. Hence `checkNoWait(0)` should always return true.

MarkMielke:
Retesting will require a bit of effort - but please check if the form of the new change is compatibility with your thinking.

MarkMielke:
(I use a patched branch of 1.10.1 for internal use for my real-life usage of TigerVNC, as master is not necessarily stable...)

MarkMielke:
I think I introduced the exact problem I was describing as possible... ambiguity of results... might need to update it further... checking...

MarkMielke:
Testing passes.

CendioOssman:
Nice. Thanks.

PR #1063, opened 2020-07-15
Correction to socket error connection message

ayoder770:
When connecting from a client to the vncserver, if the socket connection fails, the error popup currently states "unable connect to socket". This pull request contains a one-line code change to make this error "unable to connect to socket" to be more correct and consistent with the other socket errors. Thank you


CendioOssman:
Great. Thanks.

Issue #1073, opened 2020-08-05
Pixel byte order regression with commit cc647db


gearnau:
Hello,

I recently moved to TigerVNC1.10 and I am afraid that a regression was introduced, when compared to 1.7
It used to be OK when I was doing this:  SSH from intel to SPARC system, X11 tunelling. Run vncviewer to localhost

Now with 1.10, I get colours messed up the same way as shown in #738

_Originally posted by @gearnau in https://github.com/TigerVNC/tigervnc/issues/738#issuecomment-669129678_

gearnau:
How to reproduce:
1. login on x86 system
2. ssh -X <sparc>
3. vncserver
4. vncviewer 0:1

Colours are OK with 1.7 and not OK with 1.10
Both vnc server and viewer are running on SPARC
However, X11 is on x86
Both 1.7 and 1.10 show that they opt for little endian rgb888 but colours are garbled on 1.10

johnmartin-oracle:
Backing out changeset cc647db, in particular the change to Surface_X11.cxx, does restore the correct behavior.

I need more time to digest this problem but I believe the fundamental issue is the changeset forces a compile
time assumption that the byte-order of the system on which vncviewer is run matches the system on which
the X window is rendered.  In the failing case both vncviewer and vncserver are run on a big endian SPARC
but the rendered window is on a little endian x86:

  (x86 little endian) ssh -X  ---- X forwarding ----->   (SPARC big endian) vncviewer ---> vncserver/Xvnc

CendioOssman:
You guys are really trying to create the worst conditions possible for this code. :)

Unfortunately this seems to be a bit of a pain to figure out as the X Render specification is not very clear on how you determine byte ordering. I'm not sure I'll have time to look at it before a 1.12.0 release.

(any particular reason you are using X forwarding, which is slow, instead of running vncviewer on your client?)

gearnau:
Simple reason: the thin clients from which we access have ssh and X forwarding but no "onboard" tigervnc.

gearnau:
Simple reason: the thin clients from which we access have ssh and X forwarding but no "onboard" tigervnc.

johnmartin-oracle:
Can we just use the X Visual mask to determine BGRA ordering at runtime?

https://github.com/johnmartin-oracle/tigervnc/commit/adec29a62cff83054defff2ad6ddacaa81d5b3c4

CendioOssman:
[...]

PR #1079, opened 2020-08-19
Runtime selection of BGRA ordering

johnmartin-oracle:
For issue #1073 replace the compile time selection of BGRA ordering in changeset cc647db with
runtime selection using the X Visual mask.


CendioOssman:
As mentioned on #1073, there is unfortunately no strong connection between visuals and XRender (it has its own equivalent since it needs things like alpha). So we need something else to examine.

PR #1084, opened 2020-08-28
ARGB runtime XImage byteorder selection

johnmartin-oracle:
Respin of the original runtime ARGB selection pull request to solve the regression introduced by #cc647db.
and being tracked under #1073.  This uses XImageByteOrder() instead.  Tested on Solaris/SPARC server
and Solaris/x86 client.

### DIFF
--- a/vncviewer/Surface_X11.cxx
+++ b/vncviewer/Surface_X11.cxx
@@ -123,17 +123,17 @@ void Surface::alloc()
   // we find such a format
   templ.type = PictTypeDirect;
   templ.depth = 32;
-#if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
-  templ.direct.alpha = 0;
-  templ.direct.red   = 8;
-  templ.direct.green = 16;
-  templ.direct.blue  = 24;
-#else
-  templ.direct.alpha = 24;
-  templ.direct.red   = 16;
-  templ.direct.green = 8;
-  templ.direct.blue  = 0;
-#endif
+  if (XImageByteOrder(fl_display) == MSBFirst) {
+    templ.direct.alpha = 0;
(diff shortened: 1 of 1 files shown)


CendioOssman:
Great. This is hopefully more likely to be correct.

However have you been able to confirm it somewhere? Just testing isn't sufficient as the return value of this function has historically varied depending on the graphics card and not just the CPU architecture.

johnmartin-oracle:
I only confirmed it by looking at the source.  The value is what is reported
by the X server (((_XPrivDisplay)(dpy))->byte_order).  It will be up to the
X server to set the correct value given the render path (CPU + GPU).
It is not just reporting the byte order of the CPU on which the library
call is made.

CendioOssman:
Alright. Have you tested this locally on the big-endian machine as well, or just via X11 forwarding?

johnmartin-oracle:
> Alright. Have you tested this locally on the big-endian machine as well, or just via X11 forwarding?

Not yet as I don't have local access to a SPARC system with a monitor attached.  I will ask.

gearnau:
I have tested it in the typical fail way: SSH from x86 to a SPARC and then tunnel vncviewer. The colors are not garbled anymore so it looks ok.
I have not tested (because I cannot compile for it) a "normal" vncviewer from x86 to a VNC service running on SPARC. However, I have tested the opposite: vncviewer from SPARC to a VNC service running on x86: works fine as well.

CendioOssman:
Just to confirm, you were running vncviewer on a local display on that SPARC machine?

gearnau:
> Just to confirm, you were running vncviewer on a local display on that SPARC machine?

Yes, sorry for the very long delay before reply...

CendioOssman:
Great, thanks.

PR #1101, opened 2020-09-16
Do not assume sysconfdir is always /etc

metalefty:
Refer @CMAKE_INSTALL_FULL_SYSCONF@ instead.


CendioOssman:
Looks good. Although it seems like `vncserver` and `vncserver@.service` also has a few hard coded references.

Could you have a look at those as well?

metalefty:
Sure.

metalefty:
Paths like `/etc/X11` are left as is because they're system path not sysconfdir configured during TigerVNC build.

PR #1114, opened 2020-09-26
     Generate libtigervnc library from COMMON, which is convenient for applications to use the VNC protocol.   Support MSVC compiler

KangLin:


CendioOssman:
There are a lot of different changes in this PR. I'm afraid we need to have them split up more in to individual commits with a clear description on what each commit improves (and why).

KangLin:
> There are a lot of different changes in this PR. I'm afraid we need to have them split up more in to individual commits with a clear description on what each commit improves (and why).

I am editing. There are several goals:
1. Generate libtigervnc library from COMMON, which is convenient for applications to use the VNC protocol.
2. Add dynamic library
3. Support MSVC compiler

CendioOssman:
Ah, all right. I'll mark this PR as a draft then and you can switch it back to a normal PR once you feel you are done.

KangLin:
The following functions have been completed:
1.   Generate libtigervnc library from COMMON, which is convenient for applications to use the VNC protocol.
 2. Support MSVC compiler

Please review.

CendioOssman:
I'm afraid your commits are still a bit too big. Please split them up in such a way that you can describe the purpose of each commit in the commit message. We need to understand why you've done things the way you have in order to review it.

PR #1115, opened 2020-09-29
Use /run instead of /var/run which is just a symlink

grulja:
See https://lwn.net/Articles/436012/.


CendioOssman:
I supposed enough time has passed that we can assume `/run` is now standard. Or are there any holdout distributions?

Still, I'm afraid this patch is incomplete as it only changes where systemd should look for the file. You also need to change the code that actually writes the file.

grulja:
> I supposed enough time has passed that we can assume `/run` is now standard. Or are there any holdout distributions?

I'm not aware of any. Even RHEL 7 should be using /run already as it's based on Fedora 19.

> Still, I'm afraid this patch is incomplete as it only changes where systemd should look for the file. You also need to change the code that actually writes the file.

Right, should be fixed now.

PR #1127, opened 2020-10-17
Support msvc

KangLin:


CendioOssman:
This is a good start. But I'll need you to write a bit on each commit why it is needed. We don't have a MSVC setup here to test with so we need to understand why things need to be changed this way.

KangLin:
There is no big problem, mainly the compiler compatibility error.

KangLin:
There are added ci of appveyor.

KangLin:
I has add ci. so that please merge it!

CendioOssman:
As mentioned before, you need to add some more details to the commits so we understand why these changed are needed.

KangLin:
It's already very detailed. It is mainly a compilation error caused by the order of windows.h and Winsock.h.

kgold2:
I'm not a developer, just a user, but I'd really appreciate a MSVC project.  The build instructions are Linux-centric and it's unclear what software has to be installed for a Windows build.  "Double click the solution file, click build" would be quite nice.

Or a precompiled binary!

bphinz:
We provide pre-compiled binary files.  Please see the tigervnc.org homepage
where you can find links to the latest release as well as nightly builds.

Building TigerVNC on Windows has never been easy, mostly because there are
so many prerequisites that themselves are complicated to build on Windows.
I gave up trying to do that a long time ago because it's so much easier to
use a cross-compiler.

CendioOssman:
This PR is unfortunately not progressing in a way that would allow us to merge it. Closing.

PR #1137, opened 2020-10-25
OS X Japanese Keyboard Support

gsittyz:
Added OS X JIS Keyboard keys (kVK_JIS_Eisu, kvK_JIS_Kana). The Kana and Eisu keys are used to alter IME behavior, but currently the Kana key produces an unwanted space character and the Eisu key does not work.

### DIFF
--- a/vncviewer/cocoa.mm
+++ b/vncviewer/cocoa.mm
@@ -358,6 +358,9 @@ static const int kvk_map[][2] = {
   { kVK_ANSI_Keypad7,           XK_KP_7 },
   { kVK_ANSI_Keypad8,           XK_KP_8 },
   { kVK_ANSI_Keypad9,           XK_KP_9 },
+  // Japanese Keyboard Support
+  { kVK_JIS_Eisu,               XK_Muhenkan },
+  { kVK_JIS_Kana,               XK_Hiragana_Katakana },
 };
 
 int cocoa_event_keysym(const void *event)


CendioOssman:
Thank you for your contribution, and sorry for the late reply.

I'm having problem reproducing this here. I'm not getting any key events at all for these keys. How have you configured your keyboard settings to get this to work?

And shouldn't it be `XK_Eisu_toggle` instead of `XK_Muhenkan`?

gsittyz:
> I'm having problem reproducing this here. I'm not getting any key events at all for these keys. How have you configured your keyboard settings to get this to work?

Do you mean Eisu(英数)/Kana(かな) keys (on a Mac JIS keyboard) do not produce any key events, or these keys do produce key events but tigerVNC doesn't capture them? 
In my MacBook Air (13-inch, Early 2015, JIS keyboard, macOS 10.15.7), the app called Key Codes (https://apps.apple.com/jp/app/key-codes/id414568915) shows 
```
Key Down
	Characters:	 
	Unicode:		32 / 0x20
	Keys:		 
	Key Code:	102 / 0x66
	Modifiers:	256 / 0x100

Key Up
	Characters:	 
	Unicode:		32 / 0x20
	Keys:		 
	Key Code:	102 / 0x66
	Modifiers:	256 / 0x100
```
by pressing an Eisu key, and
```
Key Down
	Characters:	 
	Unicode:		32 / 0x20
	Keys:		 
	Key Code:	104 / 0x68
	Modifiers:	256 / 0x100

Key Up
	Characters:	 
	Unicode:		32 / 0x20
	Keys:		 
	Key Code:	104 / 0x68
	Modifiers:	256 / 0x100
```
by pressing a Kana key.
I didn't completely test my modification works (Sorry, I didn't have much time, so I just read the code and added modification to it). If my modification doesn't work, I'll build the app with my modification and investigate the problem more concretely.

> And shouldn't it be XK_Eisu_toggle instead of XK_Muhenkan? 

I referenced https://github.com/cb22/macbook12-spi-driver/pull/35, which maps a Mac Eisu key to a Muhenkan key.

As a matter of fact, this is a complicated problem. Mac has a special JIS keyboard layout and behavior: the Eisu key (located on the left of the Space) disables IME, while the Kana key (located on the right of the Space) enables IME. However, the ordinary JIS keyboard layout (OADG 109A keyboard) has the Muhenkan key on the left of the Space key, and the Henkan key on the right. The Hiragana/Katakana key is on the right of the Henkan key. 

As far as I know, on the iBus-Mozc (most popular Japanese IME), The Hiragana/Katakana key (which usually fires XK_Hiragana_Katakana) behaves like the Mac Kana key. 
On the other hand, there is no similar key / keysym to the Mac Eisu key with the default setting of iBus-Mozc. Therefore, we need to alter the iBus-Mozc setting anyway.
If we map the Mac Eisu key to XK_Eisu_toggle, many users should pay attention to both xmodmap and the IME setting. However, mapping to XK_Muhenkan makes people focus only on the IME setting. 

XK_Eisu_toggle is a candidate of the Mac Eisu alternative, and its behavior is somewhat similar but somewhat different from the Mac Eisu key. The problem is that CapsLock key (keycode 66) often generates XK_Eisu_toggle, which is often set as a `lock` modifier on `xmodmap` on a generic Japanese keyboard setting. If we map the Mac Eisu key to XK_Eisu_toggle, many Japanese users need to pay attention to both xmodmap and IME settings, which might be frustrating.

```
xmodmap:  up to 4 keys per modifieo r, (keycodes in parentheses):

shift       Shift_L (0x32),  Shift_R (0x3e)
lock        Eisu_toggle (0x42)
control     Control_L (0x25),  Control_R (0x69)
mod1        Alt_L (0x40),  Alt_R (0x6c),  Meta_L (0xcd)
mod2        Num_Lock (0x4d)
mod3      
mod4        Super_L (0x85),  Super_R (0x86),  Super_L (0xce),  Hyper_L (0xcf)
mod5        ISO_Level3_Shift (0x5c),  Mode_switch (0xcb)
```
```
$ xmodmap -pke | grep Eisu
keycode  66 = Eisu_toggle Caps_Lock Eisu_toggle Caps_Lock
```

The Muhenkan key locates on the same position as Mac Eisu key, but its default behavior is quite different from the Mac Eisu key. The Muhenkan key does not play an important role in xmodmap modifiers. Also, Japanese users considered it a legacy key, and some of them configure the IME setting so that XK_Muhenkan bahaves like the Mac Eisu key. If we map Mac Eisu key to XK_Muhenkan, the user can focus only on the IME setting. Although the default behavior is different, I think mapping to XK_Muhenkan is a better solution in terms of simplicity and the range of keymap influence.

gsittyz:
I forgot to say that code_map_osx_to_qnum should be fixed in order to get these two keys to work. I'm going to fix keycodemapdb in qemu, but it will take a certain amount of time.

CendioOssman:
> 
> Do you mean Eisu(英数)/Kana(かな) keys (on a Mac JIS keyboard) do not produce any key events, or these keys do produce key events but tigerVNC doesn't capture them?
> In my MacBook Air (13-inch, Early 2015, JIS keyboard, macOS 10.15.7), the app called Key Codes (https://apps.apple.com/jp/app/key-codes/id414568915) shows

Ah. I have a PC keyboard (so a 109A I assume). Can those be used with macOS?

Thank you for the extra details. I'll try to get through it soon. This is a bit new to me so bare with me. :)

I've also found some other bugs in the keycodemapdb with regards to Japanese keyboards, so I can include your fixes as well with a PR if you want.

gsittyz:
I managed to debug the TigerVNC app in Xcode, and confirmed that pressing the Eisu/Kana key on Mac produced a key event, so the problem is key mapping in TigerVNC.

> Ah. I have a PC keyboard (so a 109A I assume). Can those be used with macOS?

You can use the keyboard, but Muhenkan, Henkan and Hiragana/Katakana keys do nothing to Mac in the default setting.
You might be able to emulate Eisu and Kana keys with Karabiner-Elements.
https://www.sumahoinfo.net/entry/mac-karabiner-elements-setting/ (sorry it's Japanese) will help.
The key point is the image above "下の画像のように、プルダウンから日本語キーボードに即したキーを選択できるため、迷うことなく設定できます" (please search this sentence).

In the "Simple modifications" tab, map
- PCキーボードの無変換キー(=Muhenkan key on a PC keyboard, left of the space key) to 英数キー(=Eisu key)

and 

- PCキーボードの変換キー(=Henkan key on a PC keyboard, right of the space key) to かなキー(=Kana key).

Then you can use Muhenkan key as Eisu key and Henkan key as Kana key.

Note: PCキーボードのかなキー means Hiragana/Katakana(/Romaji) key on a PC keyboard, which is on the right of Henkan key.

CendioOssman:
> If we map the Mac Eisu key to XK_Eisu_toggle, many Japanese users need to pay attention to both xmodmap and IME settings, which might be frustrating.

I can understand that, but we try to avoid getting too deep in trying to get the perfect keyboard experience. TigerVNC should just let you access your system remotely. Bugs, problems and all. Otherwise we get easily caught up in a much larger discussion on what keyboard layouts should look like, which quickly gets very subjective.

In this case I think it is relevant to see what would happen if you would plug that Apple keyboard in to a PC. If I'm reading the data from XKB correctly such a keyboard should be configured with `applealu_jis` for everything to work correctly. This does a number of things:

 * It maps the Hangul and Hanja keycodes to Eisu and Kana. It seems Apple decided to reuse these keys. (see https://github.com/freedesktop/xkeyboard-config/blob/master/keycodes/macintosh#L201)

 * It maps those keys to the symbols `Hiragana_Katakana` and `Eisu_toggle`. (see https://github.com/freedesktop/xkeyboard-config/blob/master/symbols/macintosh_vndr/jp#L29)

I'm hoping this can be sufficient for most. And for other cases we should get keycodemapdb properly updated so those users can use `RawKeyboard` and configure things however they want.

CendioOssman:
> 	Unicode:		32 / 0x20

This is concerning. I would prefer not looking at the keycode to determine the key. Especially now that it seems like the same keycodes are used for Korean layouts. Do you happen to know how those look like on Apple devices?

CendioOssman:
Hm... It seems Apple keyboards doesn't have those keys on their Korean layouts. And constants they use directly refer to the Japanese keys. So I guess we can go ahead with that for now.

Who knows what happens if you use a Korean PC keyboard on a mac though...

gsittyz:
> I'm hoping this can be sufficient for most. And for other cases we should get keycodemapdb properly updated so those users can use `RawKeyboard` and configure things however they want.

OK, I understand that.

> Who knows what happens if you use a Korean PC keyboard on a mac though...

I do not speak Korean but I found an (unofficial) article about the han/yeong and hanja keys.
Nothing happens in Mac when we press the han/yeong or hanja key (on an external keyboard).

> 그리고 한/영 키나 한자 키를 눌러도 아무런 반응이 없어서 우리를 더 좌절하게 만들죠.
> And when we press the Korean/English key or the Chinese character key, nothing happens, which makes us more frustrated.
https://godoftyping.wordpress.com/2018/04/16/mac-%EB%A7%A5%EC%97%90%EC%84%9C-%EC%9C%88%EB%8F%84%EC%9A%B0-%ED%82%A4%EB%B3%B4%EB%93%9C-%EC%82%AC%EC%9A%A9%ED%95%98%EA%B8%B0/

According to https://www.w3.org/TR/uievents-code/ :
- Mac kana key is equivalent to the han/yeong key? 
- Mac eisu key is equivalent to the hanja key?

According to https://namu.wiki/w/%ED%95%9C%EC%98%81%20%ED%82%A4 and https://namu.wiki/w/%ED%95%9C%EC%9E%90%20%ED%82%A4 :
- han/yeong key toggles input modes (Hangul <-> English)
- hanja key converts Hangul characters into Chinese characters

CendioOssman:
> 
> According to https://www.w3.org/TR/uievents-code/ :
> 
> * Mac kana key is equivalent to the han/yeong key?
> 
> * Mac eisu key is equivalent to the hanja key?

Great find. It unfortunately contradicts the XKB data though.

The USB specification confirms the mapping of `Lang1`/`Lang2` to Hangul/Hanja at least:

https://usb.org/document-library/hid-usage-tables-121

And this patch for FreeBSD confirms that Apple are reusing those scan codes in their USB keyboards:

https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=204709

That patch also maps Kana to Hangul, and Eisu to Hanja. So there is a majority of sources for that mapping. So that seems more likely to be correct.

I don't suppose you have an Apple USB JIS keyboard and a Linux machine so we can check if the XKB data is indeed incorrect?

CendioOssman:
I've suggested fixes for what we've found so far here:

https://gitlab.com/keycodemap/keycodemapdb/-/merge_requests/8

gsittyz:
> I don't suppose you have an Apple USB JIS keyboard and a Linux machine so we can check if the XKB data is indeed incorrect?
I'm afraid I don't have an Apple USB JIS keyboard....

CendioOssman:
I've committed the keycode mapping fixes as 0d22c7bd365637889c05dc3ad08652a3e719c46a. Do you think you can adjust the Eisu keysym here and then we can merge this PR as well?

gsittyz:
Sorry I'm tied up right now.... May I work on this next week?

CendioOssman:
That is perfectly all right. No rush. :)

CendioOssman:
Have you had a chance to look at this? :)

CendioOssman:
I've gone ahead and merged this and #1142. If you have any more input feel free to comment though and we can fix things up in later commits.

PR #1142, opened 2020-10-31
Windows JIS Keyboard Support

gsittyz:
Added vkey mappings for Japanese keyboards because the special keys for Japanese input do not work on TigerVNC currently.

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -1060,6 +1060,16 @@ int Viewport::handleSystemEvent(void *event, void *data)
 
     self->handleKeyPress(keyCode, keySym);
 
+    // We don't get reliable WM_KEYUP for these
+    switch (keySym) {
+    case XK_Zenkaku_Hankaku:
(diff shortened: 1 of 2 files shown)


CendioOssman:
Thank you for your contribution. I'm having some issues getting Japanese layout working in Windows so I can test this. Could you share how you've set things up?

gsittyz:
I think this instruction will help.
https://www.thewindowsclub.com/install-japanese-keyboard-on-windows-10

CendioOssman:
For some reason it just throws errors on my main test machine. Fortunately it worked better on a second machine, so I've been able to test this.

Your mapping look fine, but there virtual key codes used are unfortunately not unique to Japanese so there is a problem with conflicts here.

I had a look at Firefox and they do a similar mapping, but only if the current layout is Japanese. So I think it's probably best if we do the same. I'll add some commits to your branch. Please have a look and see if you think it looks reasonable.

CendioOssman:
@gsittyz ?

PR #1176, opened 2021-01-11
Have a history of used vnc servers to select from in the vncviewer

johahauf:


CendioOssman:
For reference, this is a follow up to #1155.

CendioOssman:
Looks good. Thanks for getting this finished up. :)

PR #1186, opened 2021-01-15
Ask for reconnect if connection is lost

johahauf:
My vpn connection is sometimes unstable. To make the re-connect more convenient instead of exiting after an error now an option is offered to reconnect. Just a convenience feature which saves a few clicks.

In addition added *~ to gitignore to ignore my emacs backup files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 *.lo
 .deps
 .libs
+*~
 
 CMakeFiles
 CMakeCache.txt
--- a/vncviewer/vncviewer.cxx
+++ b/vncviewer/vncviewer.cxx
@@ -87,9 +87,8 @@ char vncServerName[VNCSERVERNAMELEN] = { '\0' };
 
 static const char *argv0 = NULL;
 
-static bool inMainloop = false;
 static bool exitMainloop = false;
-static const char *exitError = NULL;
+string exitError;
 
 static const char *about_text()
 {
@@ -109,28 +108,10 @@ static const char *about_text()
   return buffer;
 }
 
-void exit_vncviewer(const char *error, ...)
+void exit_vncviewer(const std::string& error)
 {
-  // Prioritise the first error we get as that is probably the most
-  // relevant one.
-  if ((error != NULL) && (exitError == NULL)) {
-    va_list ap;
-
-    va_start(ap, error);
-    exitError = (char*)malloc(1024);
-    if (exitError)
-      (void) vsnprintf((char*)exitError, 1024, error, ap);
-    va_end(ap);
-  }
-
-  if (inMainloop)
-    exitMainloop = true;
-  else {
-    // We're early in the startup. Assume we can just exit().
-    if (alertOnFatalError && (exitError != NULL))
-      fl_alert("%s", exitError);
-    exit(EXIT_FAILURE);
-  }
+  exitError = error;
+  exitMainloop = true;
 }
 
 bool should_exit()
@@ -144,18 +125,30 @@ void about_vncviewer()
   fl_message("%s", about_text());
 }
 
-void run_mainloop()
+static int run_mainloop(const char* vncserver, network::Socket* sock)
 {
+  exitMainloop = false;
+
+  CConn cc(vncServerName, sock);
   int next_timer;
 
-  next_timer = Timer::checkTimeouts();
-  if (next_timer == 0)
-    next_timer = INT_MAX;
+  while (!exitMainloop) {
+    next_timer = Timer::checkTimeouts();
+    if (next_timer == 0)
+      next_timer = INT_MAX;
 
-  if (Fl::wait((double)next_timer / 1000.0) < 0.0) {
-    vlog.error(_("Internal FLTK error. Exiting."));
-    exit(-1);
+    if (Fl::wait((double)next_timer / 1000.0) < 0.0) {
+      vlog.error(_("Internal FLTK error. Exiting."));
+      exit(-1);
+    }
   }
+
+  int ret = 0;
+  if (!exitError.empty() && alertOnFatalError)
+    ret = fl_choice("%s.\nTry reconnect?", "No", "Yes", 0, exitError.c_str());
+  exitError.clear();
+
+  return ret;
 }
 
 #ifdef __APPLE__
@@ -431,8 +424,8 @@ potentiallyLoadConfigurationFile(char *vncServerName)
       vncServerName[VNCSERVERNAMELEN-1] = '\0';
     } catch (rfb::Exception& e) {
       vlog.error("%s", e.str());
-      exit_vncviewer(_("Error reading configuration file \"%s\":\n\n%s"),
-                     vncServerName, e.str());
+      exit_vncviewer(std::string("Error reading configuration file \"")
+                     +vncServerName +"\":\n\n" +e.str());
     }
   }
 }
@@ -632,8 +625,8 @@ int main(int argc, char** argv)
     // TRANSLATORS: "Parameters" are command line arguments, or settings
     // from a file or the Windows registry.
     vlog.error(_("Parameters -listen and -via are incompatible"));
-    exit_vncviewer(_("Parameters -listen and -via are incompatible"));
-    return 1; /* Not reached */
+    exit_vncviewer("Parameters -listen and -via are incompatible");
+    return 1;
   }
 #endif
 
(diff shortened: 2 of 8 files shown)


vncviewer/vncviewer.cxx
```diff
@@ -503,6 +503,30 @@ static int mktunnel()
 }
 #endif /* !WIN32 */
 
+static int mainloop(const char* vncserver, network::Socket* sock)
```
samhed, line 506: Feels strange to have one function called `mainloop` and another called `run_mainloop`, could you find a different name for this one? Otherwise we should just keep this code in the `main` function.

.gitignore
```diff
@@ -4,7 +4,7 @@
 *.lo
 .deps
 .libs
-
+*~
```
samhed, line 7: Would you mind splitting thischange to it's own commit?

samhed (changes requested):
Sounds nice, thanks for the PR! I have a few items that should be looked at.

.gitignore
johahauf: No could do this

vncviewer/vncviewer.cxx
johahauf: `main` is already quite lengthy so I definitely would vote for putting this in its own method. Good names for the method are welcome, I will also again make up my mind about this

vncviewer/vncviewer.cxx
```diff
@@ -503,6 +503,30 @@ static int mktunnel()
...
+
+  inMainloop = true;
+  exitMainloop = false;
+  while (!exitMainloop)
+    run_mainloop();
+  inMainloop = false;
+
+  delete cc;
+
+  if (exitError != NULL && alertOnFatalError)
```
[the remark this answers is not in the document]
johahauf: The description for `alertOnFatalError` I am aware of is: 
> Display a dialog with any fatal error before exiting.
So the dialog is still shown, just with the additional option to retry. If the option is off no dialog is shown and no retry is done.

vncviewer/vncviewer.cxx
```diff
@@ -503,6 +503,30 @@ static int mktunnel()
...
+    return fl_choice("%s. Retry?", "No", "Yes", 0, exitError);
```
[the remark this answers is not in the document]
johahauf: As far as I understand the code we reach this line only in case before a successful connection was done. In case a connection could not be established the code errors out in `exit_vncviewer` around line 122. This is at least what I observed in my tests.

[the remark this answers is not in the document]
samhed: Oh, so this PR doesn't handle a connection drop while working, so to speak? That makes the title of the PR quite misleading

vncviewer/vncviewer.cxx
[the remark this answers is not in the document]
samhed: Yeah, thats fine if the "Retry" only applies to the initial connection.

vncviewer/vncviewer.cxx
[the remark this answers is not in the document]
johahauf: No, sorry, the other way round. This PR only handles the case when the connection before was successful. But not the case when no connection can be established initially.

vncviewer/vncviewer.cxx
johahauf: My proposal would be to rename `run_mainloop` to `timed_wait` as this is only a wait timer and not a main loop. And than rename my new method `mainloop` into `run_mainloop`. What do you think about this?

.gitignore
samhed: what do you mean?

johahauf: Sorry, I meant YES could do this...

vncviewer/vncviewer.cxx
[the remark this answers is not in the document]
samhed: Ah, okay. In that case, wouldn't "Reconnect?" be a better question compared to "Retry?"?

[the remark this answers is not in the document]
johahauf: Yes, will change it.

vncviewer/vncviewer.cxx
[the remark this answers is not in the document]
samhed: Since you explained above that this does apply to a successful connection I think this feature might be more important than showing an error before exiting, for some users.

I think we should add info about the reconnect-feature to the description of that parameter at the very least. Perhaps even split this behavior to it's own parameter.

vncviewer/vncviewer.cxx
samhed: Yeah, that sounds good to me.

.gitignore
samhed: Thanks

samhed (commented):
This looks good I think. I see you had to make changes after the main loop was modified on master. I'd recommend rebasing this PR on master and squishing the two commits aside from the gitignore one.

I have yet to test the actual function however.

vncviewer/vncviewer.cxx
```diff
@@ -144,18 +144,36 @@ void about_vncviewer()
...
 }
 
-void run_mainloop()
+static int run_mainloop(const char* vncserver, network::Socket* sock)
 {
-  int next_timer;
+  if(exitError != NULL) {
+    free(exitError);
+    exitError = NULL;
+  }
```
CendioOssman, line 152: This looks a bit like it's trying to hide errors. I think it's clearer if you reset `exitError` after you've run `fl_choice()` and the user has asked to reconnect.

vncviewer/vncviewer.cxx
```diff
@@ -144,18 +144,36 @@ void about_vncviewer()
...
+      vlog.error(_("Internal FLTK error. Exiting."));
+      exit(-1);
+    }
   }
+  inMainloop = false;
+
+  delete cc;
+
+  if (exitError != NULL && alertOnFatalError)
+    return fl_choice("%s. Try reconnect?", "No", "Yes", 0, exitError);
```
CendioOssman, line 174: Have you checked what kind of error messages we might have? I'm concerned we might need some text before `exitError` to explain what is going on.

vncviewer/vncviewer.cxx
```diff
@@ -144,18 +144,36 @@ void about_vncviewer()
...
+    if (Fl::wait((double)next_timer / 1000.0) < 0.0) {
+      vlog.error(_("Internal FLTK error. Exiting."));
+      exit(-1);
+    }
   }
+  inMainloop = false;
+
+  delete cc;
+
+  if (exitError != NULL && alertOnFatalError)
```
CendioOssman, line 173: I'm not sure this is a safe principle. We've not previously made the distinction between errors that are fatal for the connection, and errors that are fatal for the entire process.

Have you gone through the calls to `exit_vncviewer()` and see how they are used?

We should probably also split `exit_vncviewer()` to two functions to handle different levels of errors.

CendioOssman (changes requested):
I think this is a good idea and would be nice to have. But I have to caution that we are going in to somewhat uncharted territory here. The code has so far been written with the idea that the viewer terminates after each connection. So there might be bugs once we start doing multiple connections.

Have you looked through the code anything to check that things do proper cleanup after each connection?

johahauf:
Here how the message looks on my end:
image

samhed:
Good points by @CendioOssman 

I built your vncviewer and did some simple testing, it seems your dialog doesn't show when killing the VNC server or when running 'vncconfig -disconnect' on the remote. Is that intended behavior? Is there any way to easily trigger it?

johahauf:
My intention was mainly to handle cases where a connection was interrupted due to network issues and not so much about retries in case the server dies. The 2 cases I tried are:
1. Disable VNC while connected to my PC at the office
2. Turn of WLAN while having a VNC open.
Both worked on my end also multiple times. I could not see any trouble with this, however I did not do any memory checking to see if there are leaks somewhere in the code. Enhancing this functionality beyond my described use cases I can investigate to make it a more general retry, but of course this will also enlarge the possibility for unclean states to be revealed. Let me know if I should invest into this direction or not.

johahauf:
Spend some hours to catch all cases to ask for a retry. Now also in case of wrong password or server stops working.
I tested various cases and could not identify issues, of course some risk still exists that the retry will reveal coding issues which were hidden so far by terminating in case of any error.

PR #1188, opened 2021-01-19
TEMP: Attempt docker registry mirror

CendioOssman:


CendioOssman:
Travis seems to have fixed their quota with Docker, so this doesn't seem to be needed.

PR #1191, opened 2021-01-23
Xvnc: Added the ability to listen on both a Unix socket and TCP port.

jlesage:
Added an option to allow Xvnc to listen on both a Unix socket and TCP port.  When the option is not set, the same behaviour as before is kept (i.e. setting a unix socket path disables listening on TCP port).

### DIFF
--- a/unix/x0vncserver/x0vncserver.cxx
+++ b/unix/x0vncserver/x0vncserver.cxx
@@ -261,7 +261,9 @@ int main(int argc, char** argv)
     if (rfbunixpath.getValueStr()[0] != '\0') {
       listeners.push_back(new network::UnixListener(rfbunixpath, rfbunixmode));
       vlog.info("Listening on %s (mode %04o)", (const char*)rfbunixpath, (int)rfbunixmode);
-    } else {
+    }
+
+    if ((int)rfbport != -1) {
       if (localhostOnly)
         createLocalTcpListeners(&listeners, (int)rfbport);
       else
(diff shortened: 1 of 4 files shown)


CendioOssman:
Oh. I guess we should have thought about this scenario when we did this.

We have a lot of options already, so ideally I'd like to avoid adding more if we don't have to. It's a good principle to keep existing behaviour, but I think the unix socket function is new enough and rarely used so we could forgo that this time. So I'd like to suggest changing the behaviour to having both TCP and Unix by default. We could add that specifying `rfbport=-1` disables TCP.

Out of curiosity, what does your use case look like where you need both?

jlesage:
Thanks for your feedback!  I will adjust the PR with your suggestion.

In my case, `Xvnc` is used in Docker container to provide UI of the implemented application.  Users typically access the UI through a web browser with noVNC.  In this scenario, a Unix socket between the web server and Xvnc is used (no need to expose a network port).  However, users also have the choice to enable access of the UI via a typical VNC client, thus the need to also access Xvnc through a TCP port.

CendioOssman:
Please also have a look at keeping x0vncserver in sync. And updating the man pages.

jlesage:
Sure, I will do.

Also, I assume that inetd must continue to disable listening on both TCP and unix socket ?

CendioOssman:
I'd say so, yes. inetd mode is a special way of connecting so other ways are likely not relevant.

jlesage:
Here you go.  I tried to minimize the number of changes in `vncExtInit.cc` to ease the review (else the indentation of the unix and tcp port code blocks would need to be changed).

CendioOssman:
Looks great. Thanks. :)

PR #1192, opened 2021-01-30
Support for middle mouse emulation using ALT+leftClick.

or-tal-0:
Useful for MacBook and other touchpads where it's hard or impossible
to press the left and right buttons simultameously

### DIFF
--- a/po/tigervnc.pot
+++ b/po/tigervnc.pot
@@ -310,7 +310,11 @@ msgid "View only (ignore mouse and keyboard)"
 msgstr ""
 
 #: vncviewer/OptionsDialog.cxx:726
-msgid "Emulate middle mouse button"
+msgid "Emulate middle mouse button left+right"
+msgstr ""
+
+#: vncviewer/OptionsDialog.cxx:731
+msgid "Emulate middle mouse button ALT+left"
 msgstr ""
 
 #: vncviewer/OptionsDialog.cxx:732
--- a/tests/unit/emulatemb.cxx
+++ b/tests/unit/emulatemb.cxx
@@ -18,6 +18,7 @@
 
 #include <stdio.h>
 #include <vector>
+#include <map>
 #include <unistd.h>
 
 #include <rfb/Rect.h>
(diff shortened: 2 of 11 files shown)


CendioOssman:
Well that's annoying. I would had hoped that macOS had native support for middle click on a touch pad.

A quick Google suggests that Cmd+Click is the common workaround, so I think your patch is on the right track. This is complicated code though so I'd like to see updates to the unit test that checks for correct behaviour. Could you have a look at that?

or-tal-0:
Hi,

Sure, I'll try to add test this weekend.
I also noticed that the MacOS version does not support high resolution
(NSHighResolutionCapable), Is there already work done to add support? Will
you accept a patch for that? and if so, any advice?

Thanks,
Or

CendioOssman:
>  I also noticed that the MacOS version does not support high resolution (NSHighResolutionCapable), Is there already work done to add support? Will you accept a patch for that? and if so, any advice?

A patch would be very welcome. But it's not trivial. We've tried it before and had performance issues when we turned it on. We also have some scaling issues in the current code. See #878 for more details.

CendioOssman (changes requested):
Great. So the unit tests should lock down some of the behaviour.

The existing code uses a state machine to make sure there aren't any odd missed cases. Have you looked at having a state machine here as well? I'd very much prefer to have something that is easy to compare to.

Please also try to keep the same terminology, e.g. "filter" and "send.

or-tal-0:
I've changed the code to use a state machine. Please review from scratch.

or-tal-0:
Hi, Did you get a chance to re-review this PR? I've changed the code to use a state machine.
Thanks,

CendioOssman:
I'm afraid I haven't gotten to this yet. It is not forgotten though. :)

po/tigervnc.pot
```diff
@@ -310,7 +310,11 @@ msgid "View only (ignore mouse and keyboard)"
 msgstr ""
 
 #: vncviewer/OptionsDialog.cxx:726
-msgid "Emulate middle mouse button"
+msgid "Emulate middle mouse button left+right"
+msgstr ""
+
+#: vncviewer/OptionsDialog.cxx:731
+msgid "Emulate middle mouse button ALT+left"
```
CendioOssman, line 317: We update this as part of the release so you don't need to include changes to it here.

tests/unit/emulatemb.cxx
```diff
@@ -483,5 +748,16 @@ int main(int argc, char** argv)
   testDragAndTimeout();
   testDragAndRelease();
 
+  testMBEMKeyPress(false);
+  testMBEMKeyPress(true);
```
CendioOssman, line 752: Feel free to be more verbose in the naming in the test as it should be easy to see here what this tests.

vncviewer/EmulateMB.h
```diff
@@ -21,29 +21,50 @@
...
 
 class EmulateMB : public rfb::Timer::Callback {
 public:
-  EmulateMB();
+  EmulateMB(rdr::U32 emulateMBModKey);
 
   void filterPointerEvent(const rfb::Point& pos, int buttonMask);
 
+  bool filterKeyPress(int keyCode, rdr::U32 keySym);
+  bool filterKeyRelease(int keyCode);
```
CendioOssman, line 33: Let's have the same symmetry with `filterPointerEvent()` in that these call things on their own, and hence have no return value.

vncviewer/EmulateMB.h
```diff
@@ -21,29 +21,50 @@
...
+
 protected:
   virtual void sendPointerEvent(const rfb::Point& pos, int buttonMask)=0;
 
   virtual bool handleTimeout(rfb::Timer *t);
 
+  virtual void writeKeyEvent(rdr::U32 keySym, rdr::U32 keyCode, bool down)=0;
```
CendioOssman, line 40: Again, let's try to have symmetry with the existing stuff. So I'd like a `sendKeyPress()` and `sendKeyRelease()` here.

vncviewer/EmulateMB.h
```diff
@@ -21,29 +21,50 @@
...
+
 private:
+  typedef const signed char Action[3];
+  static Action mbStateTab[8][5];
```
CendioOssman, line 44: Let these be local to the `.cxx` file, just like the existing state table.

vncviewer/EmulateMB.cxx
```diff
@@ -194,23 +205,198 @@ static const signed char stateTab[11][5][3] = {
...
 }
 
 void EmulateMB::filterPointerEvent(const rfb::Point& pos, int buttonMask)
+{
+  if (emulateMiddleButton)
+    filterPointerEventLR(pos, buttonMask);
+  else if (emulateMiddleButtonMod)
+    filterPointerEventMod(pos, buttonMask);
+  else
+    sendPointerEvent(pos, buttonMask);
```
CendioOssman, line 391: This is really nice! Keeps things neat and not having to solve both modes in the same code.

vncviewer/EmulateMB.cxx
```diff
@@ -194,23 +205,198 @@ static const signed char stateTab[11][5][3] = {
...
-EmulateMB::EmulateMB()
-  : state(0), emulatedButtonMask(0), timer(this)
+/*
+ * 2) Pressing a modifier + left button to emulate middle button:
+ *
+ * At each state, we need handlers for the following events
+ *   0: Modifier pressed
+ *   1: Modifier released
+ *   2: Left button pressed / released
+ *   3: Other button pressed / released
```
CendioOssman, line 215: Unfortunately multiple buttons can be pressed/released in the same event. So things need to be structured like the existing state table.

I'm not sure, but it might also be an advantage to change the keyboard handling to the same line of thinking. I.e. track what state the keys are in, rather than the change. It might be confusing that for mouse we react to current state, but for keys we react to the change.

vncviewer/EmulateMB.cxx
```diff
@@ -194,23 +205,198 @@ static const signed char stateTab[11][5][3] = {
...
+ * action = 0: Dont Emulate
+ * action < 0: Emulate
+ *
+ * Note that a key pressed event (where the key is not the emulation modifier) always sends the key.
+ */
+
+EmulateMB::Action EmulateMB::mbStateTab[8][5] = {
+/* 0 ground */
+  {
+    {  0,  0,  1 },   /* mod down -> mod down */
```
CendioOssman, line 240: Is this how other such systems behave? I would suspect that pressing Alt here and then releasing it on left mouse down would be more robust.

vncviewer/EmulateMB.cxx
```diff
@@ -194,23 +205,198 @@ static const signed char stateTab[11][5][3] = {
...
+    writeKeyEvent(emulateMiddleButtonModifierKey, modKeyCode, false);
+}
+
+bool EmulateMB::filterKeyPressRelease(bool is_press, int keyCode, rdr::U32 keySym)
+{
+  if (!emulateMiddleButtonMod)
+    return false;
+
+  Action *action;
+  if (keySym == emulateMiddleButtonModifierKey || keyCode == modKeyCode) {
```
CendioOssman, line 324: I would not trust `modKeyCode` unless this is a release. And ideally you know that Alt is pressed.

vncviewer/EmulateMB.h
```diff
@@ -21,29 +21,50 @@
...
+
 private:
   int state;
   int emulatedButtonMask;
   int lastButtonMask;
   rfb::Point lastPos, origPos;
   rfb::Timer timer;
+
+  // Variables for mod-button1 emulation
+  rdr::U32 emulateMiddleButtonModifierKey;	// The key sym used for emulation
```
CendioOssman, line 64: We only support Alt, so this can be a constant.

vncviewer/EmulateMB.h
```diff
@@ -21,29 +21,50 @@
...
+  int mbState;
```
CendioOssman, line 65: Can't we use the existing `state`? Both modes cannot be active at the same time.

vncviewer/EmulateMB.h
CendioOssman, line 33: You probably need a reset function as well. Unfortunately you aren't guaranteed to get a release event for every press. So we need a way to reset the state machine when focus is lost.

tests/unit/emulatemb.cxx
```diff
@@ -33,31 +34,295 @@ static const int right          = 0x04;
...
+  emulateMiddleButton.setParam(false);
+  emulateMiddleButtonMod.setParam(enabled);
+
+  test.handleKeyPress(a_code, a_sym);
+  test.handleKeyRelease(a_code);
+
+  test.handleKeyPress(cmd_code, cmd_sym);
+  test.handleKeyPress(a_code, a_sym);
+  test.handleKeyRelease(a_code);
+  test.handleKeyRelease(cmd_code);
```
CendioOssman, line 161: Try to split things like these to separate test cases. If things go wrong it makes it easier to see what is broken.

tests/unit/emulatemb.cxx
```diff
@@ -483,5 +748,16 @@ int main(int argc, char** argv)
...
+  testMBEMKeyPress(false);
+  testMBEMKeyPress(true);
+
+  testMBEMModifier(false);
+  testMBEMModifier(true);
+
+  testMBEMClick();
+  testMBEMKeyAndClick();
+  testMBEMRightClick();
+  testMBEMMultiClick();
```
CendioOssman, line 760: I'm lacking tests with more keys pressed. I.e. if I press Ctrl+Alt and then left click. Should that be a middle click? Or a left click? What does other systems do in that case?

CendioOssman (changes requested):
Nice. I think it is on the right way. But there are some questions that need to be sorted out. Please see my comments.

CendioOssman:
Closing due to lack of activity.

PR #1196, opened 2021-02-05
Try reconnect if connection aborted or cannot be established

johahauf:


samhed:
Continued from #1186

samhed (changes requested):
I've tested it now and it works great as far as I can tell. However, it seems like you have included a lot of irrelevant and unwanted changes in this commit? Could you look that over?

johahauf:
Hi @samhed,
thanks for testing and the positive feedback. However I am not sure which changes you think are unrelated? I reviewed all the error messages and updated the comments where necessary. Also I changed the error handling from va_args to a std::string as this makes the code a way simpler an eases the handling of the error messages.

samhed:
> Also I changed the error handling from va_args to a std::string as this makes the code a way simpler an eases the handling of the error messages.

That's what I think is unrelated, the code uses `va_args` in many places, if you want this sort of change that fits into it's own PR, not here. Also, you can't just remove the call to the translation function `_()`.

vncviewer/vncviewer.cxx
```diff
@@ -87,9 +87,8 @@ char vncServerName[VNCSERVERNAMELEN] = { '\0' };
 
 static const char *argv0 = NULL;
 
-static bool inMainloop = false;
 static bool exitMainloop = false;
-static const char *exitError = NULL;
+string exitError;
```
CendioOssman, line 91: Should still be static.

vncviewer/vncviewer.cxx
```diff
@@ -113,24 +112,19 @@ void exit_vncviewer(const char *error, ...)
...
 
-  if (inMainloop)
-    exitMainloop = true;
-  else {
-    // We're early in the startup. Assume we can just exit().
-    if (alertOnFatalError && (exitError != NULL))
-      fl_alert("%s", exitError);
-    exit(EXIT_FAILURE);
-  }
+  exitError = errorString;
```
CendioOssman, line 126: I don't think it's legal to assign NULL to a string.

vncviewer/vncviewer.cxx
```diff
@@ -633,12 +651,12 @@ int main(int argc, char** argv)
...
     vlog.error(_("Parameters -listen and -via are incompatible"));
     exit_vncviewer(_("Parameters -listen and -via are incompatible"));
-    return 1; /* Not reached */
+    return 1;
   }
 #endif
 
   if (listenMode) {
-    std::list<SocketListener*> listeners;
+    list<SocketListener*> listeners;
```
CendioOssman, line 659: Please avoid unrelated cleanups like this in the same commit.

CendioOssman (changes requested):

vncviewer/CConn.cxx
```diff
@@ -272,12 +272,16 @@ void CConn::socketEvent(FL_SOCKET fd, void *data)
       exit_vncviewer(_("The connection was dropped by the server "
                        "before the session could be established."));
     } else {
-      exit_vncviewer();
+      exit_vncviewer(_("Connection error: %s\n\n"), e.str());
     }
+    recursing = false;
+    return;
```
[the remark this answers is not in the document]
johahauf: exit_vncviewer is no longer calling exit immediately so recursing and return is needed here.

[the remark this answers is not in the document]
CendioOssman: `exit_vncviewer()` only calls `exit()` during early errors. In a case like this it would just break the mainloop.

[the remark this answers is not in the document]
johahauf: actually `exit_vncviewer()` in this version never calls `exit` but only breaks the mainloop. The mainloop than waits in the re-connect dialog for the user. During this wait for the user input I saw the error message again and again...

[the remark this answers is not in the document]
CendioOssman: Hmm... that suggests some cleanup isn't done. This should get cleared by the destructor for `CConn` once it goes out of scope.

I think you'll have to debug a bit more to see why this error message keeps repeating.

CendioOssman:
Have you been able to have a look at the comments, @johahauf?

johahauf:
Sorry, was busy the last weeks, but updated the solution now to differentiate between exit due to interrupt and abort.
Please have a look and let me know your opinion.

samhed (approved):
Looks good, seems you have addressed all @CendioOssman's comments.

I've tested the functionality again and it works as advertised. I tested a connection to a VNC server running on a VM and disconnecting the machine from the network. After a while vncviewer gave up and displayed the alert. After enabling the network again I chose to reconnect which worked well.

vncviewer/vncviewer.cxx
```diff
@@ -109,28 +108,47 @@ static const char *about_text()
...
+    errorString = (char*)malloc(1024);
+    if (errorString) {
+      (void) vsnprintf((char*)errorString, 1024, error, ap);
+      vlog.error("%s",errorString);
+    }
     va_end(ap);
   }
 
-  if (inMainloop)
-    exitMainloop = true;
```
CendioOssman, line 127: This function could unfortunately be called when the mainloop has been started, so this still needs to be handled.

vncviewer/vncviewer.cxx
```diff
@@ -109,28 +108,47 @@ static const char *about_text()
...
-      fl_alert("%s", exitError);
-    exit(EXIT_FAILURE);
+  exit(1);
+}
+
+
+void exit_vncviewer(const char *error, ...)
+{
+  // Prioritise the first error we get as that is probably the most
+  // relevant one.
```
CendioOssman, line 136: Same thing here. This comment is no longer respected.

vncviewer/vncviewer.cxx
```diff
@@ -144,20 +162,43 @@ void about_vncviewer()
...
-    vlog.error(_("Internal FLTK error. Exiting."));
-    exit(-1);
+    if (Fl::wait((double)next_timer / 1000.0) < 0.0) {
+      vlog.error(_("Internal FLTK error. Exiting."));
+      exit(-1);
+    }
   }
 }
 
+static void mainloop(const char* vncserver, network::Socket* sock)
```
CendioOssman, line 182: `mainloop()` and `run_mainloop()` are a bit too similar. Can we tweak the naming to clarify what the difference is?

vncviewer/vncviewer.cxx
```diff
@@ -678,9 +717,8 @@ int main(int argc, char** argv)
           }
       }
     } catch (rdr::Exception& e) {
-      vlog.error("%s", e.str());
```
CendioOssman, line 681: Did you have a specific plan when moving these?

vncviewer/vncviewer.cxx
```diff
@@ -109,28 +108,47 @@ static const char *about_text()
...
-  else {
-    // We're early in the startup. Assume we can just exit().
-    if (alertOnFatalError && (exitError != NULL))
-      fl_alert("%s", exitError);
-    exit(EXIT_FAILURE);
+  exit(1);
+}
+
+
+void exit_vncviewer(const char *error, ...)
```
CendioOssman, line 133: This should probably be renamed given that it no longer reliably exits vncviewer.

CendioOssman (changes requested):
Sorry for the delay. I missed that you had updated the code.

There are still a few tweaks I'd like to see. But I think all the main parts look good now. So some last polish and we can merge this. :)

vncviewer/vncviewer.cxx
johahauf: abort_vncviewer is already logging the error so no need to do it twice I thought.

vncviewer/vncviewer.cxx
johahauf: No good idea.... I tend to merge `run_mainloop` and `mainloop` into one method, as `run_mainloop` is just about 10 lines of code and `mainloop` is also small. Would you agree?

vncviewer/vncviewer.cxx
CendioOssman: The message given to `abort_vncviewer()` is something we show to the user, which isn't always the same thing we want to put in a log. So I think it is best to separate the two, just like we had before.

vncviewer/vncviewer.cxx
CendioOssman: That's my point, it shouldn't be calling `exit()` and just abruptly terminating things. At least not if we've gotten to the point of the mainloop. Getting the ability to clean things up is always preferred.

vncviewer/vncviewer.cxx
CendioOssman: Works for me.

vncviewer/vncviewer.cxx
CendioOssman: Referring to "vncviewer" might no longer be relevant for this function as it is now more about terminating the current connection, not the entire viewer.

vncviewer/vncviewer.cxx
johahauf: `exit_vncviewer` is called before the main loop is entered in vncviewer.cxx. Only in touch.cxx it might be called when we are already in the main loop and I read the code like it should have all cases a terminating character (`return; //Not reached` is directly after the call to `exit_vncviewer`). So I see no need for a graceful shutdown. While writing this maybe the better name is to call this `terminate_vncviewer` to make it clear that this is not a graceful shutdown.

vncviewer/vncviewer.cxx
johahauf: OK fine, it just looked inconsistent to me that sometimes something is logged and in other cases not and thought the easiest is to have the logging at a central place.

vncviewer/vncviewer.cxx
CendioOssman: That comment was probably just a copy-paste error.

Shutting down as gracefully as we can should always be preferred unless there is some technical limitations preventing us from doing so. Otherwise we might leave stray stuff around, not signal other programs properly, or various other issues.

johahauf: Than my proposal would be to limit `exit_vncviewer` to vncviewer.cxx and have all other places to call `abort_vncviewer` which does a graceful termination/reconnect. Is this the direction you are thinking?

CendioOssman: No, I still think it's a good separation of one aborting the connection, and the other aborting the viewer. There are some errors where a reconnect will not be able to fix things (hopefully few though).

So this function here should be able to break the mainloop in a way that does not trigger a reconnect. A simple extra flag should be able to solve that.

CendioOssman:
Just a heads up that I'll need some time before I can have a look at this as I'm currently swamped with Cendio stuff.

CendioOssman:
I feel like we're going in circles a bit here, so I've adjusted the things I feel need changing and committed everything as 44b085adee9a0ecf697b48e38d2434d920385928. So this is effectively merged now.

Issue #619, opened 2018-03-25
[REQUEST] Mouse synchronization/grabbing/capture/wrapping


cpixl:
I'm trying to develop an alternative for Nvidia GameStream or Steam In-Home Streaming, but using VNC (+ PulseAudio) due to the architecture/hardware/OS limitations of the former technologies.

I've been able to run CS:GO and BioShock Infinite perfectly inside `vncserver` (using VirtualGL/`optirun` + some hacks on Steam command launcher), but unfortunately these FPS games (and probably most of FPS games) are known to use the following steps to allow 360° view rotation:

- reposition the mouse to the center of the screen
- poll/watch mouse movements
- move the 360° view according to relative distance of the mouse to the center of the screen
- repeat

Without being able to reposition the mouse to the center, the game will spin endlessly because the relative distance of the mouse to the center of the screen will always be enormous.

I've been reading on the internet reports of other VMs/VNCs lacking this feature and I found some relevant discussions ([Parallels](https://forum.parallels.com/threads/spinning-screen-in-games-in-mouse-look.94590/), [Splashtop](https://support-splashtoppersonal.splashtop.com/hc/en-us/community/posts/222608427-Trouble-controlling-game-with-mouse), [Splashtop again](https://support-splashtoppersonal.splashtop.com/hc/en-us/community/posts/222583907-Mouse-control-and-gaming-), [Splashtop AGAIN](https://support-splashtoppersonal.splashtop.com/hc/en-us/community/posts/222589447-Solution-to-the-mouse-problem-when-gaming), [Synergy conflict](http://www.guldmyr.com/blog/spinning-camera-in-fps-games/)) about these implementations (or lack of).

Anyway, my feature request is the same as all of these linked topics: Is it possible to sync/grab/capture/wrap the mouse inside the VNC viewer? This feature is implemented for keyboard on fullscreen mode only, but doesn't apply to mouse movements.

CendioOssman:
XWayland has pointer warping detection/emulation that could be used as a reference:

https://lists.x.org/archives/xorg-devel/2016-April/049351.html
https://cgit.freedesktop.org/xorg/xserver/tree/hw/xwayland/xwayland-input.c

We also need to look at a VNC protocol extension for this.

Enverex:
[...]

PR #1198, opened 2021-02-08
Add support for notifying clients about pointer movements

lhchavez:
This change adds support for the VMware Mouse Position
pseudo-encoding[1], which is used to notify VNC clients when X11 clients
call `XWarpPointer()`[2]. This function is called by SDL[3] (and other
similar libraries)  when they detect that the server does not support
native relative motion, like some RFB clients.

With this, RFB clients can choose to adjust the local cursor position
under certain circumstances to match what the server has set. For
instance, if pointer lock has been enabled on the client's machine and
the cursor is not being drawn locally, the local position of the cursor
is irrelevant, so the RFB client can use what the server sends as the
canonical absolute position of the cursor. This ultimately enables the
possibility of games (especially FPS games) to behave how users expect
(if the clients implement the corresponding change).

You can try this out end-to-end with noVNC with
https://github.com/novnc/noVNC/pull/1520 applied!

Part of: #619

1: https://github.com/rfbproto/rfbproto/blob/master/rfbproto.rst#vmware-cursor-position-pseudo-encoding
2: https://tronche.com/gui/x/xlib/input/XWarpPointer.html
3: https://hg.libsdl.org/SDL/file/28e3b60e2131/src/events/SDL_mouse.c#l804

### DIFF
--- a/common/rfb/ClientParams.cxx
+++ b/common/rfb/ClientParams.cxx
@@ -30,7 +30,7 @@ ClientParams::ClientParams()
     compressLevel(2), qualityLevel(-1), fineQualityLevel(-1),
     subsampling(subsampleUndefined),
     width_(0), height_(0), name_(0),
-    ledState_(ledUnknown)
+    cursorPos_(0, 0), ledState_(ledUnknown)
 {
   setName("");
 
@@ -83,6 +83,11 @@ void ClientParams::setCursor(const Cursor& other)
 {
   delete cursor_;
   cursor_ = new Cursor(other);
+}
+
+void ClientParams::setCursorPos(const Point& pos)
+{
+  cursorPos_ = pos;
 }
 
 bool ClientParams::supportsEncoding(rdr::S32 encoding) const
@@ -178,6 +183,13 @@ bool ClientParams::supportsLocalCursor() const
   if (supportsEncoding(pseudoEncodingCursor))
     return true;
   if (supportsEncoding(pseudoEncodingXCursor))
+    return true;
+  return false;
+}
+
+bool ClientParams::supportsCursorPosition() const
+{
+  if (supportsEncoding(pseudoEncodingVMwareCursorPosition))
     return true;
   return false;
 }
--- a/common/rfb/ClientParams.h
+++ b/common/rfb/ClientParams.h
@@ -77,6 +77,9 @@ namespace rfb {
     const Cursor& cursor() const { return *cursor_; }
     void setCursor(const Cursor& cursor);
 
+    const Point& cursorPos() const { return cursorPos_; }
+    void setCursorPos(const Point& pos);
+
     bool supportsEncoding(rdr::S32 encoding) const;
 
     void setEncodings(int nEncodings, const rdr::S32* encodings);
@@ -91,6 +94,7 @@ namespace rfb {
     // Wrappers to check for functionality rather than specific
     // encodings
     bool supportsLocalCursor() const;
+    bool supportsCursorPosition() const;
     bool supportsDesktopSize() const;
     bool supportsLEDState() const;
     bool supportsFence() const;
@@ -110,6 +114,7 @@ namespace rfb {
     PixelFormat pf_;
     char* name_;
     Cursor* cursor_;
+    Point cursorPos_;
     std::set<rdr::S32> encodings_;
     unsigned int ledState_;
     rdr::U32 clipFlags;
--- a/common/rfb/SMsgWriter.cxx
+++ b/common/rfb/SMsgWriter.cxx
@@ -42,7 +42,8 @@ SMsgWriter::SMsgWriter(ClientParams* client_, rdr::OutStream* os_)
   : client(client_), os(os_),
     nRectsInUpdate(0), nRectsInHeader(0),
     needSetDesktopName(false), needCursor(false),
-    needLEDState(false), needQEMUKeyEvent(false)
+    needCursorPos(false), needLEDState(false),
+    needQEMUKeyEvent(false)
 {
 }
 
@@ -269,6 +270,14 @@ void SMsgWriter::writeCursor()
   needCursor = true;
 }
 
+void SMsgWriter::writeCursorPos()
+{
+  if (!client->supportsEncoding(pseudoEncodingVMwareCursorPosition))
+    throw Exception("Client does not support cursor position");
+
+  needCursorPos = true;
+}
+
 void SMsgWriter::writeLEDState()
 {
   if (!client->supportsEncoding(pseudoEncodingLEDState) &&
@@ -293,6 +302,8 @@ bool SMsgWriter::needFakeUpdate()
   if (needSetDesktopName)
     return true;
   if (needCursor)
+    return true;
+  if (needCursorPos)
     return true;
   if (needLEDState)
     return true;
@@ -339,6 +350,8 @@ void SMsgWriter::writeFramebufferUpdateStart(int nRects)
     if (needSetDesktopName)
       nRects++;
     if (needCursor)
+      nRects++;
+    if (needCursorPos)
       nRects++;
     if (needLEDState)
       nRects++;
@@ -453,6 +466,18 @@ void SMsgWriter::writePseudoRects()
     }
 
     needCursor = false;
+  }
+
+  if (needCursorPos) {
+    const Point& cursorPos = client->cursorPos();
+
+    if (client->supportsEncoding(pseudoEncodingVMwareCursorPosition)) {
+      writeSetVMwareCursorPositionRect(cursorPos.x, cursorPos.y);
+    } else {
+      throw Exception("Client does not support cursor position");
+    }
+
+    needCursorPos = false;
   }
 
   if (needSetDesktopName) {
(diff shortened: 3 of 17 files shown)


lhchavez:
The failures were caused by Docker authentication limits. I swear I did not break the build ^^;;

CendioOssman:
Nice. I'll try to have a look and comment.

Have you also looked at adding support to our viewer? We prefer if features are fully implemented in both client and server so we can test properly without external projects.

lhchavez:
> Nice. I'll try to have a look and comment.
> 
> Have you also looked at adding support to our viewer? We prefer if features are fully implemented in both client and server so we can test properly without external projects.

not originally, but let me see how hard would it be.

lhchavez:
> > Nice. I'll try to have a look and comment.
> > Have you also looked at adding support to our viewer? We prefer if features are fully implemented in both client and server so we can test properly without external projects.
> 
> not originally, but let me see how hard would it be.

Turns out most of the functionality was already there! Used this little snippet of code[1] because fltk does not support relative mouse events and the cursor eventually hit a wall at the edges of the screen and broke everything unless we also mirrored the cursor-warping that was done remotely.

1: https://www.mail-archive.com/fltk-opengl@easysw.com/msg00023.html

vncviewer/DesktopWindow.cxx
```diff
@@ -322,6 +324,27 @@ void DesktopWindow::setCursor(int width, int height,
...
+#elif defined(__APPLE__)
+  CGPoint new_pos;
+  new_pos.x = pos.x;
+  new_pos.y = pos.y;
+  CGWarpMouseCursorPosition(new_pos);
+#else // Assume this is Xlib
+  Window rootwindow = DefaultRootWindow(fl_display);
+  XWarpPointer(fl_display, rootwindow, rootwindow, 0, 0, 0, 0,
+               pos.x, pos.y);
+#endif
```
CendioOssman, line 344: The coordinates here look wrong. The VNC screen might not be at coordinates 0,0, so everything here needs to be adjusted by an offset.

CendioOssman (changes requested):
This looks really nice. Good work!

Please also merge the two fix commits in too the others so we just have the two original commits. You can rebase and force push to fix up this.

How is this best tested? And what other servers and clients implement this?

vncviewer/DesktopWindow.cxx
lhchavez: haha, this is my first change of a client application, so i basically stumbled upon a solution that worked for me in this one case (translation: please suggest what offset should be added).

i kinda guess that the `viewport->x/y()` is what we want, but have no way of confirming since i don't know any situations other than the one i can test locally.

lhchavez:
> This looks really nice. Good work!
> 
> Please also merge the two fix commits in too the others so we just have the two original commits. You can rebase and force push to fix up this.

Done

> 
> How is this best tested? And what other servers and clients implement this?

So far only noVNC and this are the only open source server/clients that will implement this (after a quick Sourcegraph search). To test, run any SDL game that captures the cursor. In a pinch, running https://gist.github.com/lhchavez/50f4c31fc3faf811c3a572148479ef1b (requires Python3 and pygame) should do the trick.

vncviewer/DesktopWindow.cxx
CendioOssman: I see two test scenarios:

 1. Viewport doesn't match window. Just disable "Resize remote session..." in the options and make sure the session is larger or smaller than your client window
 2. Client is on different monitors. You need two monitors for this, and disable "Enable full-screen mode over all monitors"

I would suspect you need both `viewport->x/y()` and `window()->x/y()` to properly get screen coordinates.

lhchavez: both cases seemed to work with a combination of `x_root() + viewport->x()`. thanks!

vncviewer/DesktopWindow.cxx
```diff
@@ -322,6 +324,29 @@ void DesktopWindow::setCursor(int width, int height,
 }
 
 
+void DesktopWindow::setCursorPos(const rfb::Point& pos)
+{
+  if (!mouseGrabbed) {
+    // Do nothing if we do not have the mouse captured.
+    return;
+  }
```
CendioOssman, line 332: We probably shouldn't do this if we aren't focused as well.

vncviewer/DesktopWindow.cxx
```diff
@@ -322,6 +324,29 @@ void DesktopWindow::setCursor(int width, int height,
...
+#elif defined(__APPLE__)
+  CGPoint new_pos;
+  new_pos.x = pos.x + x_root() + viewport->x();
+  new_pos.y = pos.y + y_root() + viewport->y();
+  CGWarpMouseCursorPosition(new_pos);
+#else // Assume this is Xlib
+  Window rootwindow = DefaultRootWindow(fl_display);
+  XWarpPointer(fl_display, rootwindow, rootwindow, 0, 0, 0, 0,
+               pos.x + x_root() + viewport->x(),
+               pos.y + y_root() + viewport->y());
```
CendioOssman, line 345: Have you tried this with Wayland? It doesn't seem to work for me here. :/

vncviewer/DesktopWindow.cxx
```diff
@@ -51,6 +51,8 @@
 
 #ifdef __APPLE__
 #include "cocoa.h"
+#include <CoreGraphics/CGGeometry.h>
+#include <CoreGraphics/CGRemoteOperation.h>
```
CendioOssman, line 55: The compiler can't find these here. I'm using an older SDK and gcc instead of clang. Hower `#include <Carbon/Carbon.h>` works for me. Does that work for you?

CendioOssman (changes requested):

CendioOssman:
Doesn't seem to work on Windows either. macOS kind-a works, but there seems to be something generally wrong there unrelated to your changes.

lhchavez:
> Doesn't seem to work on Windows either. macOS kind-a works, but there seems to be something generally wrong there unrelated to your changes.

splitting this change into server-side (the one i *really* care about) and client-side, since it seems that the client-side will take more time.

vncviewer/DesktopWindow.cxx
lhchavez: wouldn't that make it hard to exit the client? in any case, moved to #1212.

vncviewer/DesktopWindow.cxx
lhchavez: no, and i don't have a wayland environment :(

vncviewer/DesktopWindow.cxx
lhchavez: i don't have a macOS device ^^;; i was just fiddling until i got it working on CI. what's preferable in these cases? to drop support for macOS altogether? (in any case, moved to #1212).

CendioOssman:
We don't want you disappearing on us once the server is merged though, so I prefer the one PR. :)

CendioOssman:
The fix for macOS seems to be this at least:

```diff
diff --git a/vncviewer/DesktopWindow.cxx b/vncviewer/DesktopWindow.cxx
index bb51019e..385a9e88 100644
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -188,6 +187,14 @@ DesktopWindow::DesktopWindow(int w, int h, const char *name,
 
   // Show hint about menu key
   Fl::add_timeout(0.5, menuOverlay, this);
+
+  // By default we get a slight delay when we warp the pointer, something
+  // we don't want or we'll get jerky movement
+#ifdef __APPLE__
+  CGEventSourceRef event = CGEventSourceCreate(kCGEventSourceStateCombinedSessionState);
+  CGEventSourceSetLocalEventsSuppressionInterval(event, 0);
+  CFRelease(event);
+#endif
 }
 
 
```

vncviewer/DesktopWindow.cxx
CendioOssman: How so? The warping should generally be in the way in such cases, not helping?

vncviewer/DesktopWindow.cxx
CendioOssman: It works for me with those changes, so go with that. :)

lhchavez:
> We don't want you disappearing on us once the server is merged though, so I prefer the one PR. :)

don't worry about that, i promise right here and now to finish the whole thing. disappearing once the server is merged through will contribute to building a bad relationship with you, and I already have a lot of PRs open in repositories you own, so it's in my best interest to see through all of them (within reason: i'm probably not going to go buy a Mac just to test these changes ^^;; ).

but on the other hand, throw a man a bone! this is the part of the change that i have a vested interest in. i see the vncviewer part as "building goodwill" (other less charitable folks call it the "open source contributor tax").

lhchavez:
i have no idea why Travis is complaining about `HOWTO.md` O_O

CendioOssman:
> i have no idea why Travis is complaining about `HOWTO.md` O_O

Unrelated and fixed on `master` now. :)

lhchavez:
> > i have no idea why Travis is complaining about `HOWTO.md` O_O
> 
> Unrelated and fixed on `master` now. :)

yay, thanks! i was stumped for a bit there as to how the other PR wasn't getting in the same situation.

lhchavez:
is this change also good to merge?

CendioOssman (approved):
Yup. The only notes where about the client side of things.

samhed:
For reference, the vncviewer side of this: #1212

PR #1203, opened 2021-02-12
Add documentation to the new systemd support

grulja:
This document describes the basics users need to do in order to start Tigervnc's systemd service properly.


grulja:
Note: this needs some grammar correction from someone who is a native english speaker.

HOWTO.md
```diff
@@ -0,0 +1,109 @@
+# Changes in recent Tigervnc versions
+Previous Tigervnc versions had a wrapper script called `vncserver`. This script
+could be run as a user manually to start *Xvnc* process. The usage was quite 
+simple as you just run:
+```
+$ vncserver :x [vncserver options] [Xvnc options]
+```
+and that was it. It worked just fine, however, there were issues when users
```
CendioOssman, line 8: It worked fine for some cases, but far from all. So I think we should use more cautious wording here.

HOWTO.md
```diff
@@ -0,0 +1,109 @@
...
+```
+:x=user
+```
+For example you can have
+```
+:1=test
+:2=vncuser
+```
+
+## Configure Xvnc options
```
CendioOssman, line 27: Should we have a man page reference for getting all available options?

HOWTO.md
```diff
@@ -0,0 +1,109 @@
...
+alwaysshared
+```
+### Note:
+There is one important option you need to set and that option is the session you
+want to start. E.g when you want to start GNOME desktop, then you have to use:
+```
+session=gnome
+```
+This should match the name of a session desktop file from `/usr/share/xsessions`
+directory.
```
CendioOssman, line 61: This setting is right now optional as it will pick the first desktop it finds.

HOWTO.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Note:
+If you used Tigervnc before with your user and you already created a 
+password, then you have to make sure the `$HOME/.vnc` folder created by 
+`vncpasswd` have the correct *SELinux* context. You either can delete this 
+folder and recreate it again by creating the password one more time, or 
+alternatively you can run:
+```
+$ restorecon -RFv /home/<USER>/.vnc
+```
```
CendioOssman, line 79: Are we not backwards compatible with older folders?

HOWTO.md
```diff
@@ -0,0 +1,109 @@
...
+$ sudo systemctl start vncserver@:x
+```
+Run it as a regular user in case the user has permissions to run `sudo`. 
+Don't forget to replace the `:x` by the actual number you configured in the
+user mapping file. For example:
+```
+$ systemctl start vncserver@:1
+```
+This starts a Tigervnc server for user `test` with GNOME session.
```
CendioOssman, line 96: Should we also mention `systemctl enable`?

CendioOssman (commented):
Not a native speaker, but looks good to me. :)

HOWTO.md
grulja: I'm not sure, I think they will not have correct context in case they were created before we installed our SELinux policy.

HOWTO.md
grulja: It is optional, but I would still recommend to specify it, because the very first desktop might not work correctly. This situation for example happens on RHEL 8 with Kiosk mode installed.

grulja:
Just for the information, the CI failure is not because of my change.

HOWTO.md
CendioOssman: That's not great and I would like a future patch to improve that. Many users will have existing folders with password files in them.

CendioOssman:
This HOWTO is specific to the vncserver service, so I think it is best if it lives in that directory. Could you also change CMake so it gets installed?

grulja:
> This HOWTO is specific to the vncserver service, so I think it is best if it lives in that directory. Could you also change CMake so it gets installed?

Done.

HOWTO.md
grulja: I will try to look into that.

CendioOssman:
Thanks.

PR #1206, opened 2021-02-13
Add "High Resolution" mode for OSX

### LINKED ISSUE #878: Please add "High Resolution" flag to Mac info.plist

junhaoliao:
Fixes #878 

This commit fixes the HiDPI incompatibility of vncviewer on OSX. 

Now the sizes of the render images are multiplied by the backingScaleFactor and can be displayed normally on the screen: https://developer.apple.com/documentation/appkit/nsscreen/1388385-backingscalefactor?language=objc

# Before
## NSHighResolutionCapable: False
image
## NSHighResolutionCapable: True
image

# After:
## NSHighResolutionCapable: True
image
## NSHighResolutionCapable: True
image

# Known Issues:
The vncviewer viewport sizes are not committed to the server at launch, and the control still mismatches with the image in the beginning. The user will need to either specify "-Maximize" in the arguments, or manually resize the viewer after launching the viewer. This issue is no longer observed when I compile the viewer with FLTK 1.4.0:
https://www.fltk.org/doc-1.4/drawing.html#ssect_DrawingUnit:~:text=DPI%2Dindependent

(Sorry this is my first time submitting a pull request. It will be great if you can point out anything inappropriate. Thanks in advance!)

### DIFF
--- a/vncviewer/Surface_OSX.cxx
+++ b/vncviewer/Surface_OSX.cxx
@@ -87,10 +87,13 @@ static void render(CGContextRef gc, CGColorSpaceRef lut,
   CGContextSetBlendMode(gc, mode);
   CGContextSetAlpha(gc, alpha);
 
-  rect.origin.x = x;
-  rect.origin.y = y;
-  rect.size.width = w;
-  rect.size.height = h;
+  // adjust rect with cocoa scale factor to support HiDPI
+  const double scale_factor = cocoa_get_scale_factor();
+
+  rect.origin.x = x * scale_factor;
+  rect.origin.y = y * scale_factor;
+  rect.size.width = w * scale_factor;
+  rect.size.height = h * scale_factor;
 
   CGContextDrawImage(gc, rect, subimage);
 
--- a/release/Info.plist.in
+++ b/release/Info.plist.in
@@ -9,7 +9,7 @@
 	<key>CFBundleExecutable</key>
 	<string>TigerVNC Viewer</string>
 	<key>NSHighResolutionCapable</key>
-	<string>False</string>
+	<string>True</string>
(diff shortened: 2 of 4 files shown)


vncviewer/Surface_OSX.cxx
```diff
@@ -87,10 +87,13 @@ static void render(CGContextRef gc, CGColorSpaceRef lut,
   CGContextSetBlendMode(gc, mode);
   CGContextSetAlpha(gc, alpha);
 
-  rect.origin.x = x;
-  rect.origin.y = y;
-  rect.size.width = w;
-  rect.size.height = h;
+  // adjust rect with cocoa scale factor to support HiDPI
+  const double scale_factor = cocoa_get_scale_factor();
```
CendioOssman, line 91: Shouldn't this just be for when rendering to the screen? Not another Surface?

And I think it also depends on FLTK version. Perhaps we need to look at the transformation matrix?

CendioOssman (changes requested):
The issues you mentioned will also need to be resolved before we can merge this.

Have you done any performance profiling? Last time we had to back this out because it was slower than standard DPI mode.

prasskaly:
Hi, 
This is probably something I would like to try out. Is there a beta binary available with this update?

CendioOssman:
No response. Closing.

PR #1212, opened 2021-03-02
Support the VMware Cursor Position extension on vncviewer 

lhchavez:
This change makes it possible for re-synchronizing the remote cursor on
the vncviewer when in fullscreen mode. This is done by locally moving
the cursor position to what the server thinks it should be.

Now SDL games should work!

### DIFF
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -51,6 +51,7 @@
 
 #ifdef __APPLE__
 #include "cocoa.h"
+#include <Carbon/Carbon.h>
 #endif
 
 #define EDGE_SCROLL_SIZE 32
@@ -186,6 +187,14 @@ DesktopWindow::DesktopWindow(int w, int h, const char *name,
 
   // Show hint about menu key
   Fl::add_timeout(0.5, menuOverlay, this);
+
+  // By default we get a slight delay when we warp the pointer, something
+  // we don't want or we'll get jerky movement
+#ifdef __APPLE__
+  CGEventSourceRef event = CGEventSourceCreate(kCGEventSourceStateCombinedSessionState);
+  CGEventSourceSetLocalEventsSuppressionInterval(event, 0);
+  CFRelease(event);
+#endif
 }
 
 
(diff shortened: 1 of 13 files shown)


vncviewer/DesktopWindow.cxx
```diff
@@ -51,6 +51,8 @@
 
 #ifdef __APPLE__
 #include "cocoa.h"
+#include <CoreGraphics/CGGeometry.h>
+#include <CoreGraphics/CGRemoteOperation.h>
```
lhchavez, line 55: @CendioOssman did you get this working on macOS? or are there any further tweaks needed?

vncviewer/DesktopWindow.cxx
```diff
@@ -322,6 +332,25 @@ void DesktopWindow::setCursor(int width, int height,
 }
 
 
+void DesktopWindow::setCursorPos(const rfb::Point& pos)
+{
```
lhchavez, line 336: i take the mouse-grabbing comment back, this worked!

vncviewer/DesktopWindow.cxx
```diff
@@ -51,6 +51,7 @@
 
 #ifdef __APPLE__
 #include "cocoa.h"
+#include <CoreGraphics/CoreGraphics.h>
```
lhchavez, line 54: gave up trying to support individual headers and am now importing the whole thing. _maybe_ this will alleviate the compile errors.

vncviewer/DesktopWindow.cxx
CendioOssman: Still gettin `fatal error: CoreGraphics/CoreGraphics.h: No such file or directory`. So I would still suggest the top Carbon header.

vncviewer/DesktopWindow.cxx
CendioOssman: The grab check was fine and much welcome. My point was to **also** check focus. But then again, I guess grab implies focus so I guess that isn't needed...

CendioOssman:
X11 seems to work fine at least. So it's Wayland and Windows that need more work.

vncviewer/DesktopWindow.cxx
lhchavez: how do you do that? i don't have a way to test these changes, so i'm flying blind.

lhchavez:
> X11 seems to work fine at least. So it's Wayland and Windows that need more work.

is there a way to check for wayland to disable it there? i'll also drop windows support since i also don't have a windows machine.

vncviewer/DesktopWindow.cxx
lhchavez: Ah! got it, i didn't quite grok the meaning of the other comment. Restored the grab check.

vncviewer/DesktopWindow.cxx
CendioOssman: Just change it to `#include <Carbon/Carbon.h>` :)

CendioOssman:
> is there a way to check for wayland to disable it there? i'll also drop windows support since i also don't have a windows machine.

Having support for all our platforms is a requirement unfortunately, so we need to get these resolved. I'll see if I can dig a bit more this week.

lhchavez:
> > is there a way to check for wayland to disable it there? i'll also drop windows support since i also don't have a windows machine.
> 
> Having support for all our platforms is a requirement unfortunately, so we need to get these resolved. I'll see if I can dig a bit more this week.

got it, thanks! re-enabled that bit of code in the meantime.

CendioOssman:
Hmm... Odd. I retested and it works just fine on Windows and Wayland now. Not sure what I did wrong before.

So please put the grab check back and this is okay to merge.

vncviewer/DesktopWindow.cxx
lhchavez: _really_ restored the grab check.

PR #1242, opened 2021-04-30
made full-screen edge scrolling smoother and faster

ToyKeeper:
TigerVNC's slow viewport scrolling speed was bothering me, so I fixed it.  This makes it a lot more pleasant to use my notebook (1920x1080) as a dumb terminal for my desktop PC (2560x1440).  It gives faster and more precise control over the viewport position, so I don't have to spend so much time waiting for it to scroll when I move the cursor around.

-- commit message below --

The old method used a very slow frame rate with small steps,
which made it obnoxious during use.  The new method has a few
improvements:

- Calculates "edge" region width automatically, as a ratio of
  the viewport size.  Default is 1/16th.

- Uses a different edge width for x and y, which works better
  on very wide or very tall screens.

- Increased default frame rate from 10 fps to 60 fps.

- Replaced hardcoded frame rate with a #define.

- Reduced maximum movement per frame slightly, because the
  frame rate is so much faster.  The overall result is still much
  faster, but also easier to do small adjustments with.

- Fixed off-by-one error in scroll rate calculation formula.  Left/up
  scrolls were faster than down/right, but this is fixed.

This makes it more comfortable to use TigerVNC on a small screen
to work on a larger screen for hours at a time.

### DIFF
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -54,8 +54,18 @@
 #include <Carbon/Carbon.h>
 #endif
 
-#define EDGE_SCROLL_SIZE 32
-#define EDGE_SCROLL_SPEED 20
+// width of each "edge" region where scrolling happens,
+// as a ratio compared to the viewport size
+// default: 1/16th of the viewport size
+#define EDGE_SCROLL_SIZE 16
+// edge width is calculated at runtime; these values are just examples
+static int edge_scroll_size_x = 128;
+static int edge_scroll_size_y = 96;
+// maximum pixels to scroll per frame
+#define EDGE_SCROLL_SPEED 16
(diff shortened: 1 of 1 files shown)


ToyKeeper:
Not sure if I've got this PR set up how the maintainers like...  but please let me know if there's anything I could do to make the process easier or fit better within project standards.

CendioOssman:
Sorry for the lack of attention. This is on my todo, but I have not been able to put a lot of time towards TigerVNC as of late. I will get here eventually though so hopefully you can wait a bit longer.

If anyone else wants to test it out and have a look then that would also be welcome.

CendioOssman:
Finally got around to testing. This seems to work a lot nicer than the previous system. Thanks!

PR #1249, opened 2021-05-18
SELinux improvements

grulja:
Fedora guidelines for reference: https://fedoraproject.org/wiki/SELinux/IndependentPolicy.

### DIFF
--- a/contrib/packages/rpm/el7/SPECS/tigervnc.spec
+++ b/contrib/packages/rpm/el7/SPECS/tigervnc.spec
@@ -1,3 +1,8 @@
+
+#defining macros needed by SELinux
+%global selinuxtype targeted
+%global modulename vncsession
+
 Name:           tigervnc
 Version:        @VERSION@
 Release:        1%{?snap:.%{snap}}%{?dist}
@@ -121,10 +126,11 @@ This package contains icons for TigerVNC viewer
 %package selinux
 Summary:        SELinux module for TigerVNC
 BuildArch:      noarch
-Requires(pre):  libselinux-utils
-Requires(post): selinux-policy-base >= %{_selinux_policy_version}
-Requires(post): policycoreutils policycoreutils-python
-Requires(post): libselinux-utils
+BuildRequires:  selinux-policy-devel
+Requires:       selinux-policy-%{selinuxtype}
+Requires(post): selinux-policy-%{selinuxtype}
+BuildRequires:  selinux-policy-devel
+%{?selinux_requires}
 
 %description selinux
 This package provides the SELinux policy module to ensure TigerVNC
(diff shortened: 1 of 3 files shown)


CendioOssman:
Looks reasonable. Any compatibility changes with this? E.g. RHEL 7?

And not .xz?

grulja:
> Looks reasonable. Any compatibility changes with this? E.g. RHEL 7?

What compatibility changes you have in mind? 

> And not .xz?

I can see all SELinux policies being compressed with bzip2 or not compressed at all.

grulja:
Looking at the CI failure I can see what you mean by compatibility. I will update your SPEC file with the same changes I'm proposing for RHEL.

CendioOssman:
I was actually thinking more along the lines of "Does everyone support compressed policy modules?" and "Does this directory work everywhere?". :)

grulja:
> I was actually thinking more along the lines of "Does everyone support compressed policy modules?" and "Does this directory work everywhere?". :)

Let's find out: CC: @zpytela

contrib/packages/rpm/el7/SPECS/tigervnc.spec
```diff
@@ -234,20 +240,25 @@ if [ -x %{_bindir}/gtk-update-icon-cache ]; then
...
-%selinux_modules_install %{_datadir}/selinux/packages/vncsession.pp
-%selinux_relabel_post
+%selinux_modules_install -s %{selinuxtype} %{_datadir}/selinux/packages/%{selinuxtype}/%{modulename}.pp.bz2
+%selinux_relabel_post -s %{selinuxtype}
 
-%posttrans selinux
-%selinux_relabel_post
+if [ "$1" -le "1" ]; then # First install
+   # the daemon needs to be restarted for the custom label to be applied
+   %systemd_postun_with_restart vncserver.service
```
CendioOssman, line 253: We don't have a general service. Just the parametrized one.

vmojzis:
> I was actually thinking more along the lines of "Does everyone support compressed policy modules?" and "Does this directory work everywhere?". :)

Almost all the changes should work properly on RHEL 7 and really anywhere where you can get selinux-policy-targeted. The only exception is the conditional requires (Requires: (%{name}-selinux if selinux-policy-%{selinuxtype})), which was first introduced in rhel 8. If this is a concern for you, you can selectively build the selinux sub-package as described here https://fedoraproject.org/wiki/SELinux/IndependentPolicy#Example_spec_file_changes_to_incorporate_-selinux_subpackage and only set with_selinux on newer systems e.g.
```
%if 0%{?fedora} >= 30 || 0%{?rhel} >= 8
    %global with_selinux 1
%endif
```

CendioOssman:
We have separate spec files, so that should not be an issue.

I'll just wait for Travis to check the builds and then this should be good to merge.

Issue #1189, opened 2021-01-20
Tigervnc server not starting up in centos 8


imonmondal:
**Describe the bug**
I am unable to start vncserver in centos 8:

**Procedure** 
Started vncserver using the following procedure

Added user in 
**/etc/tigervnc/vncserver.users**
as 
1:user1
modified 
**/etc/tigervnc/vncserver-config-defaults**
as

_session=gnome
securitytypes=vncauth,tlsvnc
desktop=sandbox
#geometry=2000x1200
localhost
alwaysshared_

Started vncserver using the following command
**sudo systemctl start vncserver@:1**

**vncviewer :1** 
yields the following error
unable connect to socket : Connection refused (111).

Note: netstat -tpln confirm that vncserver is not listening to port 5901.
Have opened ports 590*

MihaiSandor:
Please provide full log of:

sudo systemctl status vncserver@:1

imonmondal:
**Status of vncserver@:1**

_Jan 20 13:17:34 vlsi183 systemd[1]: vncserver@:1.service: Succeeded.
Jan 20 13:20:19 vlsi183 systemd[1]: Starting Remote desktop service (VNC)...
Jan 20 13:20:19 vlsi183 systemd[1]: Started Remote desktop service (VNC).
Jan 20 13:20:19 vlsi183 systemd[1]: vncserver@:1.service: Succeeded.
Jan 21 16:36:20 vlsi183 systemd[1]: Starting Remote desktop service (VNC)...
Jan 21 16:36:20 vlsi183 systemd[1]: Started Remote desktop service (VNC).
Jan 21 16:36:20 vlsi183 systemd[1]: vncserver@:1.service: Succeeded.
Jan 21 16:36:28 vlsi183 systemd[1]: Starting Remote desktop service (VNC)...
Jan 21 16:36:28 vlsi183 systemd[1]: Started Remote desktop service (VNC).
Jan 21 16:36:28 vlsi183 systemd[1]: vncserver@:1.service: Succeeded._

The following are the entries in the file **/etc/tigervnc/vncserver-config-defaults**

_session=gnome
securitytypes=vncauth,tlsvnc
desktop=sandbox
#geometry=2000x1200
vlsi183
alwaysshared_



The following is the result of **netstat -tpln**

_Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name    
[...]

PR #1250, opened 2021-05-18
SELinux iImprovements for vnc session

zpytela:


grulja:
I cannot say whether these changes are correct, but I did a scratch build of Tigervnc for RHEL 8 and tested it and it seems to work.

When I run `ls -lZd .vnc` I get:
`drwxrwxr-x. 2 test3 test3 unconfined_u:object_r:vnc_home_t:s0 40 may 19 08:46`

So at least it looks it gets correct context (vnc_home_t). Let me know if there is anything else to check or try.

CendioOssman:
This is a fix for #1189, I assume?

unix/vncserver/selinux/vncsession.te
```diff
@@ -32,17 +27,20 @@ can_exec(vnc_session_t, vnc_session_exec_t)
...
 files_pid_filetrans(vnc_session_t, vnc_session_var_run_t, file)
 
-manage_files_pattern(vnc_session_t, xdm_home_t, xdm_home_t)
-manage_fifo_files_pattern(vnc_session_t, xdm_home_t, xdm_home_t)
-manage_sock_files_pattern(vnc_session_t, xdm_home_t, xdm_home_t)
-manage_lnk_files_pattern(vnc_session_t, xdm_home_t, xdm_home_t)
+manage_files_pattern(vnc_session_t, vnc_home_t, vnc_home_t)
+manage_fifo_files_pattern(vnc_session_t, vnc_home_t, vnc_home_t)
+manage_sock_files_pattern(vnc_session_t, vnc_home_t, vnc_home_t)
+manage_lnk_files_pattern(vnc_session_t, vnc_home_t, vnc_home_t)
```
CendioOssman, line 43: Does this need a migration strategy? These files might already have the type `xdg_home_t` (or `user_home_t`).

unix/vncserver/selinux/vncsession.te
```diff
@@ -20,48 +20,61 @@
...
+	userdom_spec_domtrans_all_users(vnc_session_t)
+	userdom_signal_all_users(vnc_session_t)
+
+	userdom_user_home_dir_filetrans(vnc_session_t, xdm_home_t, dir, ".vnc")
+	userdom_admin_home_dir_filetrans(vnc_session_t, xdm_home_t, dir, ".vnc")
+
+	# This also affects other tools, e.g. vncpasswd
+	gen_require(`
+		attribute userdomain;
+	')
```
CendioOssman, line 80: This is already in the top level `gen_require()`, so isn't this redundant?

zpytela:
> This is a fix for #1189, I assume?

Yes it is, this issue is referred to in the commit which resolves it.

unix/vncserver/selinux/vncsession.te
zpytela: Right, @grulja is aware of it, we spoke about some transition HOWTO document. A command like this should do the task:
```
restorecon -FRvn /home/*/.vnc
```

unix/vncserver/selinux/vncsession.te
zpytela: It is deduplicated in another commit.

unix/vncserver/selinux/vncsession.te
CendioOssman: Most users will probably not be able to figure that out unfortunately. Should it be automated in some stage?

Any precedence for how this has been handled in the past?

zpytela: The only general advice we can give is run `fixfiles -F onboot` and reboot the system. We have a precedence in running restorecon from specfile, but only for individual files or 1 directory. Running restorecon for /home can take hours and should not be in specfile. Additionally, in cases home directories are elsewhere than in /home, we can hardly cover that efficiently.

grulja: I already mention/advice that users should run restorecon on their vnc config folder, but that was for migration from old Tigervnc to a new one with proper systemd support. I'm hesitated to pick this change into RHEL 8 as it might again break users configuration and I'm thinking about picking just the one fixing https://github.com/TigerVNC/tigervnc/issues/1189. However, I think that CentOS Stream/RHEL9 might be a good candidate for this change so we don't break it again for users with future updates.

CendioOssman: We here upstream don't have the luxury of multiple versions like that, so this would put us in an uncomfortable situation.

Would it be a big no-no to do this from vncsession or vncserver?

zpytela:
I have not much experience with systemd user units, but it can be on option, too: creating a oneshot service to restore ~/.vnc context, and possibly run some other checks or actions. Then it either needs to be enabled per-user, or enabled globally in rpm scriptlet.

grulja:
Besides the issue with migration from old SELinux context to the new one, is there any other issue blocking this change before merging? Asking whether this is already safe to backport or whether I should rather wait?

CendioOssman:
> I have not much experience with systemd user units, but it can be on option, too: creating a oneshot service to restore ~/.vnc context, and possibly run some other checks or actions. Then it either needs to be enabled per-user, or enabled globally in rpm scriptlet.

I was thinking more like putting it in the code before we access the home directory here:

https://github.com/TigerVNC/tigervnc/blob/2a042c913ccc2ebf114386d33eb1aa4484d7aca4/unix/vncserver/vncsession.c#L356-L362

grulja:
Hi, I decided to try and look how this could be solved and I'm unsure this could be done from a systemd service, because we need to restore SELinux context for a different user from the user who starts the vncserver systemd service.

I can see there is selinux library, which could be used to get the context and call restorecon on a folder we need to recover, this could be all called from the code as suggested by @CendioOssman. Do you think this is a solution that could work? @zpytela what do you think?

The API that I believe can be used is:

```
**From selinux.h:**
/* Get file context, and set *con to refer to it.
   Caller must free via freecon. */
extern int getfilecon(const char *path, char ** con);

**From restorecon.h:**
/**
 * selinux_restorecon - Relabel files.
 * @pathname: specifies file/directory to relabel.
 * @restorecon_flags: specifies the actions to be performed when relabeling.
 *
 * selinux_restorecon(3) will automatically call
 * selinux_restorecon_default_handle(3) and selinux_restorecon_set_sehandle(3)
 * first time through to set the selabel_open(3) parameters to use the
 * currently loaded policy file_contexts and request their computed digest.
 *
 * Should other selabel_open(3) parameters be required see
 * selinux_restorecon_set_sehandle(3).
 */
extern int selinux_restorecon(const char *pathname,
				    unsigned int restorecon_flags);
```

zpytela:
@vmojzis can you suggest the proper library functions?

grulja:
> @vmojzis can you suggest the proper library functions?

Actually, if you can answer here: https://github.com/TigerVNC/tigervnc/pull/1372. Thank you.

Issue #1239, opened 2021-04-26
Accented keys not working when using the Java viewer


binary-manu:
**Describe the bug**
Keys with accented letters, like `ò`, `à`, `è` (and other symbols that use the same keys, like `#` which is ALT-GR+`à`) on an Italian keyboard are not sent to the server when using the Java viewer. If verbose logging is enabled, it says `Viewport: Ignoring KeyEvent with unknown Java keycode`. They work correctly when using the native client.

**To Reproduce**
1. Set locale and keymap information to Italian:
```
System Locale: LANG=it_IT.UTF-8
VC Keymap: it
X11 Layout: it
X11 Model: pc105
X11 Options: terminate:ctrl_alt_bksp
```
2. Start a local Xvnc instance running an application that can accept text.
3. Start the Java TigerVNC viewer and connect to the test server.
4. Notice that while most keys work, accented letters are ignored (don't know how to do that with a different keyboard, but these keys are usually located to the left of RETURN). 

**Expected behavior**
I should be able to use accented letters and other symbols on those keys.

**Client (please complete the following information):**
 - OS: Arch Linux x86-64
 - VNC client:
```
TigerVNC Java Viewer v1.11.0 (20200908)
Built on 2020-09-08 at 23:40:58
Copyright (C) 1999-2020 TigerVNC Team and many others (see README.rst)
See https://www.tigervnc.org for information on TigerVNC.
````
 - Client downloaded from: https://bintray.com/tigervnc/stable/tigervnc

**Server (please complete the following information):**
 - OS: Arch Linux x86-64
 - VNC server: TigerVNC
 - VNC server version: 1.11.0
 - Server downloaded from: distro package `tigervnc 1.11.0-6`
 - Server was started using: Xvnc -PasswordFile ~/.vnc/passwd -SecurityTypes VncAuth :2

**Additional context**
While trying to semi-randomly fix the issue, I noticed that if calls to `java.awt.event.KeyEvent.getKeyCode` are replaced with `getExtendedKeyCode`, those characters work. Pressing `à` causes the log to say:

```
Viewport: Key pressed: 0x10000e0 => 0x00e0
Viewport: Key released: 0x10000e0 => 0x00e0
```

bphinz:
Although it might be reasonably safe to fall back to getExtendedKeyCode only when getKeyCode fails...

binary-manu:
> Sorry that I have not had a chance to work on this

[...]

PR #1253, opened 2021-05-19
Handle keys reported only by KeyEvent.getExtendedKeyCode

binary-manu:
I tried to come up with a patch for #1239.

Basically it replaces calls to `KeyEvent.getKeyCode` with calls to `KeyEvent.getExtendedKeyCode`.
However, this change breaks the assumption that keycodes can't be larger than 65535. At least when using my local JRE, keys corresponding to accented characters have keycodes much bigger. To work around this, the combined "fake" keycode has been extended to a `long`.

As far as my local testing is concerned, it seems to work well with my Italian layout. I also performed basic tests using a US keymap and it also seems to work.

### DIFF
--- a/java/com/tigervnc/vncviewer/KeyMap.java
+++ b/java/com/tigervnc/vncviewer/KeyMap.java
@@ -161,7 +161,7 @@ public class KeyMap {
 
   public static int vkey_to_keysym(KeyEvent ev)
   {
-    int keyCode = ev.getKeyCode();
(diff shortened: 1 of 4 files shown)


CendioOssman:
@bphinz, is this something you want to get merged before release?

CendioOssman:
@binary-manu, that last commit needs to be left out. We don't want any merges going the "wrong" way as it really confuses things. You can do a force push to update this PR.

Grimstrike23 (approved):

PR #1258, opened 2021-05-25
Fix heap-based buffer overflow in screen size handling

grulja:
Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1151307. This is a patch we have been carrying for a while and it would be probably better to get it upstreamed. Just note I'm not the original author.

### DIFF
--- a/unix/x0vncserver/Image.cxx
+++ b/unix/x0vncserver/Image.cxx
@@ -82,6 +82,14 @@ void Image::Init(int width, int height)
   xim = XCreateImage(dpy, vis, DefaultDepth(dpy, DefaultScreen(dpy)),
                      ZPixmap, 0, 0, width, height, BitmapPad(dpy), 0);
 
+  if (xim->bytes_per_line <= 0 ||
+      xim->height <= 0 ||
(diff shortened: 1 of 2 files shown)


CendioOssman:
Can this still be triggered with 996356b6c65ca165ee1ea46a571c32a1dc3c3821 in place?

grulja:
I don't know. I didn't write this patch, it's just we have been applying it since I took over Tigervnc and thought it might be good to upstream it if it's still relevant.

grulja:
I don't know the API at all, this change basically checks whether XCreateImage() returns a sane XImage and I don't know whether this can happen with your changes in place. On the other hand this additional check doesn't seem to hurt and cause any potential issue . If you think it's not necessary at this point then we can close this PR and I can drop this downstream patch from our builds.

CendioOssman:
I'd say drop it. It should not be possible to hit this path given the other protection. And I don't think we have any reason to mistrust XCreateImage(). The only relevant part is the overflow check, but we have lots of places that would need that check (hence the early check), so it's just confusing to have the check here but nowhere else.

PR #1259, opened 2021-05-25
Fix viewer crash when cursor has not been set

grulja:
This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1038701. This is a patch we have been carrying for a while and it would be probably better to get it upstreamed. Just note I'm not the original author.

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -277,7 +277,7 @@ void Viewport::setCursor(int width, int height, const Point& hotspot,
     }
   }
 
-  if (Fl::belowmouse() == this)
+  if (Fl::belowmouse() == this && cursor)
     window()->cursor(cursor, cursorHotspot.x, cursorHotspot.y);
 }
 


CendioOssman:
This looks wrong. Under what circumstances can this happen? The code just above always sets cursor.

grulja:
This is a patch which was added to Tigervnc 1.3.0 in Fedora by the previous maintainer. It's possible that the code changed since then and when I look at the code now, I also think shouldn't be necessary. However, is there a chance that Fl_RGB_Image() call returns a nullptr?

CendioOssman:
I'm fairly sure C++ requires it to return something valid.

Perhaps things got confused with de1a3b95837379b023aa24d8626392838959d437?

grulja:
> Perhaps things got confused with [de1a3b9](https://github.com/TigerVNC/tigervnc/commit/de1a3b95837379b023aa24d8626392838959d437)?

Indeed!! I just checked my git log for tigervnc and it seems that this change didn't apply so I tried to rebase it did it this way, while it actually didn't apply because it was already part of upstream. 

Good you found out. One less downstream patch to be dropped.

PR #1260, opened 2021-05-25
Let user know that a view-only password is not used

grulja:
This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1639169. We have been carrying this in Fedora and RHEL for a while.


CendioOssman:
That upstream bug is closed. Can it be opened up to see the reasoning behind the change?

grulja:
This is because of following scenario:
1. Start vncpasswd
2. Press 'enter' after 'Would you like to enter a view-only password (y/n)?'

After this step you might not know whether the password you entered is for view-only or not (in case you use it for the first time probably). I know you get prompted for an additional password if you press "y", but there are apparently people who don't know that and such informative message probably doesn't hurt.

PR #1261, opened 2021-05-25
CharArray: pre-fill empty array with zeroes

grulja:
This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1637086.

Steps to Reproduce:
1. export MALLOC_CHECK_=2 MALLOC_PERTURB_=204
2. echo "123456" | vncpasswd -f >/tmp/out

### DIFF
--- a/common/rfb/util.h
+++ b/common/rfb/util.h
@@ -52,14 +52,17 @@ namespace rfb {
     CharArray(char* str) : buf(str) {} // note: assumes ownership
     CharArray(size_t len) {
       buf = new char[len]();
+      memset(buf, 0, len);
     }
     ~CharArray() {
-      delete [] buf;
+      if (buf) {
+        delete [] buf;
+      }
     }
     void format(const char *fmt, ...) __printf_attr(2, 3);
     // Get the buffer pointer & clear it (i.e. caller takes ownership)
     char* takeBuf() {char* tmp = buf; buf = 0; return tmp;}
-    void replaceBuf(char* b) {delete [] buf; buf = b;}
+    void replaceBuf(char* b) {if (buf) delete [] buf; buf = b;}
     char* buf;
   private:
     CharArray(const CharArray&);


CendioOssman:
The information around this commit is unfortunately a bit light. I'd like some more details as to *why* this is the proper fix. Adding that `length` attribute is not really ideal as it will often not have a valid value. A more robust fix would be preferred.

grulja:
The `length` attribute is initially set to zero so it will always have a valid value. It's been a while since I wrote this change so I need to think again why it was done this way, but my guess is:
1) `CharArray(size_t len)` is called and it creates a new char array (`new char[len]`), however, this array is not null terminated
2) `PlainPasswd::replaceBuf(char *b)` will call `memset()` on that buffer, however, if it was created with the constructor above, then `strlen(buf)` will not return what we want, therefore I use the `length` attribute if it's non-zero.
3) Other changes are just additional safety checks

common/rfb/util.h
```diff
@@ -51,16 +51,21 @@ namespace rfb {
...
+      if (buf) {
+        delete [] buf;
+        buf = nullptr;
+      }
     }
     void format(const char *fmt, ...) __printf_attr(2, 3);
     // Get the buffer pointer & clear it (i.e. caller takes ownership)
     char* takeBuf() {char* tmp = buf; buf = 0; return tmp;}
-    void replaceBuf(char* b) {delete [] buf; buf = b;}
+    void replaceBuf(char* b) {if (buf) delete [] buf; buf = b;}
```
johahauf, line 66: What about an update of `length` when `replaceBuf()` is called?

grulja: Good catch.

CendioOssman:
> The `length` attribute is initially set to zero so it will always have a valid value.

Right, but it doesn't always equal the length of the string. So sometimes it is null terminated, sometimes it has a clearly specified length. That kind of ambiguity is just a recipe for mistakes IMO.

These classes were clearly designed with null terminated strings in mind. So I think it would be better if we fix that up and make sure that holds true in all cases.

CendioOssman:
Hmm... If I'm reading the spec correctly, this is breakage caused by c++03. It seems like the code we have should initialize things to zero under c++98, but from c++03 the changed the behaviour. So this code likely worked when it was written.

Anyway, the `memset()` looks good. But we don't need the checks as it is perfectly safe to `delete [] NULL`. I've committed an adjusted patch in ba837fd283adbf462137be97ed6a1970a6c37b67.

PR #1262, opened 2021-05-25
Utilize system-wide crypto policies

grulja:
This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1179345. This is a fix we have been carrying for a while and it's a change according to Fedora guidelines: https://fedoraproject.org/wiki/Packaging:CryptoPolicies.

### DIFF
--- a/common/rfb/CSecurityTLS.cxx
+++ b/common/rfb/CSecurityTLS.cxx
@@ -205,26 +205,64 @@ void CSecurityTLS::setParam()
   static const char kx_anon_priority[] = ":+ANON-ECDH:+ANON-DH";
 
   int ret;
-  char *prio;
-  const char *err;
 
-  prio = (char*)malloc(strlen(Security::GnuTLSPriority) +
-                       strlen(kx_anon_priority) + 1);
-  if (prio == NULL)
-    throw AuthFailureException("Not enough memory for GnuTLS priority string");
+  // Custom priority string specified?
+  if (strcmp(Security::GnuTLSPriority, "") != 0) {
+    char *prio;
+    const char *err;
 
-  strcpy(prio, Security::GnuTLSPriority);
-  if (anon)
+    prio = (char*)malloc(strlen(Security::GnuTLSPriority) +
+                         strlen(kx_anon_priority) + 1);
+    if (prio == NULL)
+      throw AuthFailureException("Not enough memory for GnuTLS priority string");
+
+    strcpy(prio, Security::GnuTLSPriority);
+    if (anon)
+      strcat(prio, kx_anon_priority);
+
+    ret = gnutls_priority_set_direct(session, prio, &err);
+
+    free(prio);
+
+    if (ret != GNUTLS_E_SUCCESS) {
+      if (ret == GNUTLS_E_INVALID_REQUEST)
+        vlog.error("GnuTLS priority syntax error at: %s", err);
+      throw AuthFailureException("gnutls_set_priority_direct failed");
+    }
+  } else if (anon) {
+    const char *err;
+
+#if GNUTLS_VERSION_NUMBER >= 0x030603
+    ret = gnutls_set_default_priority_append(session, kx_anon_priority, &err, 0);
+    if (ret != GNUTLS_E_SUCCESS) {
+      if (ret == GNUTLS_E_INVALID_REQUEST)
+        vlog.error("GnuTLS priority syntax error at: %s", err);
+      throw AuthFailureException("gnutls_set_default_priority_append failed");
+    }
+#else
+    // We don't know what the system default priority is, so we guess
+    // it's what upstream GnuTLS has
+    static const char gnutls_default_priority[] = "NORMAL";
+    char *prio;
+
+    prio = (char*)malloc(strlen(gnutls_default_priority) +
+                         strlen(kx_anon_priority) + 1);
+    if (prio == NULL)
+      throw AuthFailureException("Not enough memory for GnuTLS priority string");
+
+    strcpy(prio, gnutls_default_priority);
     strcat(prio, kx_anon_priority);
 
-  ret = gnutls_priority_set_direct(session, prio, &err);
+    ret = gnutls_priority_set_direct(session, prio, &err);
 
-  free(prio);
+    free(prio);
 
-  if (ret != GNUTLS_E_SUCCESS) {
-    if (ret == GNUTLS_E_INVALID_REQUEST)
-      vlog.error("GnuTLS priority syntax error at: %s", err);
-    throw AuthFailureException("gnutls_set_priority_direct failed");
+    if (ret != GNUTLS_E_SUCCESS) {
+      if (ret == GNUTLS_E_INVALID_REQUEST)
+        vlog.error("GnuTLS priority syntax error at: %s", err);
+      throw AuthFailureException("gnutls_set_priority_direct failed");
+    }
+#endif
   }
 
   if (anon) {
--- a/common/rfb/SSecurityTLS.cxx
+++ b/common/rfb/SSecurityTLS.cxx
@@ -176,26 +176,64 @@ void SSecurityTLS::setParams(gnutls_session_t session)
   static const char kx_anon_priority[] = ":+ANON-ECDH:+ANON-DH";
 
   int ret;
-  char *prio;
-  const char *err;
 
-  prio = (char*)malloc(strlen(Security::GnuTLSPriority) +
-                       strlen(kx_anon_priority) + 1);
-  if (prio == NULL)
-    throw AuthFailureException("Not enough memory for GnuTLS priority string");
+  // Custom priority string specified?
+  if (strcmp(Security::GnuTLSPriority, "") != 0) {
+    char *prio;
+    const char *err;
 
-  strcpy(prio, Security::GnuTLSPriority);
-  if (anon)
+    prio = (char*)malloc(strlen(Security::GnuTLSPriority) +
+                         strlen(kx_anon_priority) + 1);
+    if (prio == NULL)
+      throw AuthFailureException("Not enough memory for GnuTLS priority string");
+
+    strcpy(prio, Security::GnuTLSPriority);
+    if (anon)
+      strcat(prio, kx_anon_priority);
+
+    ret = gnutls_priority_set_direct(session, prio, &err);
+
+    free(prio);
+
+    if (ret != GNUTLS_E_SUCCESS) {
+      if (ret == GNUTLS_E_INVALID_REQUEST)
+        vlog.error("GnuTLS priority syntax error at: %s", err);
+      throw AuthFailureException("gnutls_set_priority_direct failed");
+    }
+  } else if (anon) {
+    const char *err;
+
+#if GNUTLS_VERSION_NUMBER >= 0x030603
+    ret = gnutls_set_default_priority_append(session, kx_anon_priority, &err, 0);
+    if (ret != GNUTLS_E_SUCCESS) {
+      if (ret == GNUTLS_E_INVALID_REQUEST)
+        vlog.error("GnuTLS priority syntax error at: %s", err);
+      throw AuthFailureException("gnutls_set_default_priority_append failed");
+    }
+#else
+    // We don't know what the system default priority is, so we guess
+    // it's what upstream GnuTLS has
+    static const char gnutls_default_priority[] = "NORMAL";
+    char *prio;
+
+    prio = (char*)malloc(strlen(gnutls_default_priority) +
+                         strlen(kx_anon_priority) + 1);
+    if (prio == NULL)
+      throw AuthFailureException("Not enough memory for GnuTLS priority string");
+
+    strcpy(prio, gnutls_default_priority);
     strcat(prio, kx_anon_priority);
 
-  ret = gnutls_priority_set_direct(session, prio, &err);
+    ret = gnutls_priority_set_direct(session, prio, &err);
 
-  free(prio);
+    free(prio);
 
-  if (ret != GNUTLS_E_SUCCESS) {
-    if (ret == GNUTLS_E_INVALID_REQUEST)
-      vlog.error("GnuTLS priority syntax error at: %s", err);
-    throw AuthFailureException("gnutls_set_priority_direct failed");
+    if (ret != GNUTLS_E_SUCCESS) {
+      if (ret == GNUTLS_E_INVALID_REQUEST)
+        vlog.error("GnuTLS priority syntax error at: %s", err);
+      throw AuthFailureException("gnutls_set_priority_direct failed");
+    }
+#endif
   }
 
   if (gnutls_dh_params_init(&dh_params) != GNUTLS_E_SUCCESS)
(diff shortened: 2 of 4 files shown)


CendioOssman:
Following the system configuration is very desirable, but I think there are some issues with compatibility here. Upstream GnuTLS doesn't seem to guarantee that `@SYSTEM` exists. But how do things look in practice? Do all major distributions configure things this way?

And I think this also requires a configuration file. Which is probably fine if the system GnuTLS is used. But what about other builds, e.g. our general ones. Is the config file in the same place on all distributions?

And what happens if anything of the above fails? Can we fall back to something sensible?

CendioOssman:
How do you do this for other packages? Perhaps there are existing solutions to how upstream packages such as us can handle this?

grulja:
Hi Nikos, do you know answer to questions above? This is a change I've been requested to do in Fedora package (by you) and we carry it as a downstream patch and I thought it's something that can go upstream.

CC: @nmav

Looking at GnuTLS documentation, it says:
`Since version 3.5.1 it is allowed to specify fallback keywords such as @KEYWORD1,@KEYWORD2, and the first valid keyword will be used.`

So perhaps we can use **@SYSTEM** and fallback to **NORMAL**?

nmav:
You can use gnutls_set_default_priority() instead of using priority strings. This function will allow fedora to use SYSTEM, whereas upstream on different distros will use the upstream defaults whatever they are.

grulja:
I will, I just have some questions:
1) How to deal with `Xvnc -GnuTLSPriority` option? I would have to probably check if that option is different from Tigervnc default (`@NORMAL`) and use API you use currently, otherwise use `gnutls_set_default_priority()` instead.
2) Do you want to keep `:+ANON-ECDH:+ANON-DH` string added to priority? If so, then I don't know how to do that if you want to keep things limited to old API. I see there is `gnutls_set_default_priority_append()`, but it was added to GnuTLS 3.6.3 and this version is not in RHEL 7 for example.

CendioOssman:
@nmav, do you have any guidance here?

nmav:
> I will, I just have some questions:
> 
>     1. How to deal with `Xvnc -GnuTLSPriority` option? I would have to probably check if that option is different from Tigervnc default (`@NORMAL`) and use API you use currently, otherwise use `gnutls_set_default_priority()` instead.

Makes sense, but I'd not complicate things, if the override (GnuTLSPriority) option is specified, I'd let the user to set what they have set; otherwise  `gnutls_set_default_priority()`.

>     2. Do you want to keep `:+ANON-ECDH:+ANON-DH` string added to priority? If so, then I don't know how to do that if you want to keep things limited to old API. I see there is `gnutls_set_default_priority_append()`, but it was added to GnuTLS 3.6.3 and this version is not in RHEL 7 for example.

`:+ANON-ECDH:+ANON-DH` do not exist in TLS1.3, they have been removed from the protocol; given letsencrypt adoption today, I'd remove them completely. If you really want to allow non-certificate authentication I'd suggest [tofu](https://www.gnutls.org/manual/html_node/Verifying-a-certificate-using-trust-on-first-use-authentication.html).

grulja:
Updated PR:
1) I removed` :+ANON-ECDH:+ANON-DH` as @nmav suggested.
2) I noticed Tigervnc already calls gnutls_set_default_priority(), but overrides it later. I now do the override only if the user priority string is different from the default priority string, which I renamed to `@SYSTEM`.

CendioOssman:
> `:+ANON-ECDH:+ANON-DH` do not exist in TLS1.3, they have been removed from the protocol; given letsencrypt adoption today, I'd remove them completely. If you really want to allow non-certificate authentication I'd suggest [tofu](https://www.gnutls.org/manual/html_node/Verifying-a-certificate-using-trust-on-first-use-authentication.html).

We do have TOFU as well. However I'm unsure how many are relying on TLS without certificates, so I'm not ready to drop support for that quite yet.

> 1. I removed` :+ANON-ECDH:+ANON-DH` as @nmav suggested.
> 

That breaks all the non-certificate methods ("TLS*"), so we cannot do that. Or have I misunderstood the purpose of those settings, @nmav?

> 2. I noticed Tigervnc already calls gnutls_set_default_priority(), but overrides it later. I now do the override only if the user priority string is different from the default priority string, which I renamed to `@SYSTEM`.

I thought we concluded that `@SYSTEM` is not reliable across systems? So an empty strings is probably a better marker.

grulja:
> > `:+ANON-ECDH:+ANON-DH` do not exist in TLS1.3, they have been removed from the protocol; given letsencrypt adoption today, I'd remove them completely. If you really want to allow non-certificate authentication I'd suggest [tofu](https://www.gnutls.org/manual/html_node/Verifying-a-certificate-using-trust-on-first-use-authentication.html).
> 
> We do have TOFU as well. However I'm unsure how many are relying on TLS without certificates, so I'm not ready to drop support for that quite yet.
> 
> > 1. I removed` :+ANON-ECDH:+ANON-DH` as @nmav suggested.
> 
> That breaks all the non-certificate methods ("TLS*"), so we cannot do that. Or have I misunderstood the purpose of those settings, @nmav?
> 

Waiting for answer on this one.

> > 1. I noticed Tigervnc already calls gnutls_set_default_priority(), but overrides it later. I now do the override only if the user priority string is different from the default priority string, which I renamed to `@SYSTEM`.
> 
> I thought we concluded that `@SYSTEM` is not reliable across systems? So an empty strings is probably a better marker.

Fixed.

CendioOssman:
Oh? So GnuTLS doesn't fall back to older TLS in those cases? I guess I need to set up some tests.

Do you know which version of GnuTLS got this behaviour?

As for the intention, this is an extension we got off gtk-vnc/QEMU (VeNCrypt). We just implemented it the same way to be compatible.

CendioOssman:
It seems to work fine. When connecting without certificates it handshakes `(TLS1.2)-(ANON-ECDH)-(AES-256-CBC)-(SHA1)`. And with a certificate it handshakes `(TLS1.3)-(ECDHE-SECP256R1)-(RSA-PSS-RSAE-SHA256)-(AES-256-GCM)`. And there is nothing in our code modifying TLS versions directly.

This is with GnuTLS 3.6.14.

grulja:
> It seems to work fine. When connecting without certificates it handshakes `(TLS1.2)-(ANON-ECDH)-(AES-256-CBC)-(SHA1)`. And with a certificate it handshakes `(TLS1.3)-(ECDHE-SECP256R1)-(RSA-PSS-RSAE-SHA256)-(AES-256-GCM)`. And there is nothing in our code modifying TLS versions directly.
> 
> This is with GnuTLS 3.6.14.

So do you want me to add those ciphersuites back? Or shall we wait for confirmation from @nmav?

CendioOssman:
Yes, as removing those breaks all TLS* authentication methods.

However I'm not sure how they should be properly added if `gnutls_set_default_priority_append()` isn't available. @nmav is the expert here. Ideas?

CendioOssman:
Ping @nmav

nmav:
I'm no longer the expert, I have to refer to documentation myself to figure it out. Since you're modifying the code my suggestion is to get rid of anything related to anonymous as you'll have to do it eventually. Preshared keys is the only thing that is common between TLS 1.2 and 1.3 and doesn't require certificates. If that's not possible use `gnutls_set_default_priority_append()` when available to add the anonymous ciphersuites you need and gnutls as I see from [my post from the tls1.3 days](https://nikmav.blogspot.com/2018/05/gnutls-and-tls-13.html) it will try to negotiate tls 1.3 when it can.

CendioOssman:
Alright, so I propose the following then:

If `gnutls_set_default_priority_append()` is available, then use that. Otherwise use the upstream default of `"NORMAL"` and append what we need to that.

This will mean that things will work unmodified for anything recent, but you'll have to continue patching TigerVNC for older distributions such as RHEL 7.

Does that work for you @grulja? And can you update the PR with those changes?

grulja:
> Alright, so I propose the following then:
> 
> If `gnutls_set_default_priority_append()` is available, then use that. Otherwise use the upstream default of `"NORMAL"` and append what we need to that.
> 
> This will mean that things will work unmodified for anything recent, but you'll have to continue patching TigerVNC for older distributions such as RHEL 7.
> 
> Does that work for you @grulja? And can you update the PR with those changes?

I will update the PR the next week, I just returned from vacation so I'm sorry for late response.

common/rfb/CSecurityTLS.cxx
```diff
@@ -217,7 +217,11 @@ void CSecurityTLS::setParam()
   if (anon)
     strcat(prio, kx_anon_priority);
 
+#if GNUTLS_VERSION_NUMBER >= 0x030603
+  ret = gnutls_set_default_priority_append(session, prio, &err, 0);
+#else
   ret = gnutls_priority_set_direct(session, prio, &err);
+#endif
```
CendioOssman, line 224: This ended up a bit too far back. We still need to change the default to `""`, and call `gnutls_priority_set_direct()` in every case where it is not the empty string.

It is the specific case of anonymous mode and empty priority string that needs the latest round of special handling.

CendioOssman (changes requested):

common/rfb/CSecurityTLS.cxx
grulja: Ok, I hope I understood it correctly and didn't lost in all those possible scenarios.

common/rfb/CSecurityTLS.cxx
```diff
@@ -203,21 +203,46 @@ bool CSecurityTLS::processMsg()
...
-  strcpy(prio, Security::GnuTLSPriority);
+  if (strcmp(Security::GnuTLSPriority, "") != 0)
+    strcpy(prio, Security::GnuTLSPriority);
+#if GNUTLS_VERSION_NUMBER < 0x030603
+  else
+    strcpy(prio, Security::kx_gnutls_priority_fallback);
+#endif
+
   if (anon)
     strcat(prio, kx_anon_priority);
```
CendioOssman, line 236: `prio` might be uninitialized here (when new GnuTLS and default prio)

common/rfb/CSecurityTLS.cxx
```diff
@@ -203,21 +203,46 @@ bool CSecurityTLS::processMsg()
...
 
+#if GNUTLS_VERSION_NUMBER >= 0x030603
+  if (strcmp(Security::GnuTLSPriority, "") == 0)
+    ret = gnutls_set_default_priority_append(session, prio, &err, 0);
+  else
+    ret = gnutls_priority_set_direct(session, prio, &err);
+#else
   ret = gnutls_priority_set_direct(session, prio, &err);
```
CendioOssman, line 244: Same thing here.

common/rfb/CSecurityTLS.cxx
```diff
@@ -203,21 +203,46 @@ bool CSecurityTLS::processMsg()
...
+#else
+    prio = (char*)malloc(strlen(kx_gnutls_priority_fallback) +
+                         strlen(kx_anon_priority) + 1);
+#endif
+
   if (prio == NULL)
     throw AuthFailureException("Not enough memory for GnuTLS priority string");
 
-  strcpy(prio, Security::GnuTLSPriority);
+  if (strcmp(Security::GnuTLSPriority, "") != 0)
```
CendioOssman, line 228: For readability it might be worth adding a bool for this. It keeps repeating a few times.

CendioOssman (changes requested):
I think this should cover it. It's just one corner case that needs to be handled and then I can merge this.

common/rfb/CSecurityTLS.cxx
grulja: I don't see it. In case of default prio, it initializes `prio` for both new and old GnuTLS.

grulja: Ok, I finally understand what you mean. This one should be fixed.

common/rfb/CSecurityTLS.cxx
grulja: I think this one is correct, but what is not correct is combination of default prio and new GnuTLS, where `prio` will be only initialized when `anon` is true.

CendioOssman (changes requested):
I'm afraid this broke two other scenarios; new GnuTLS with custom priority string and anonymous authentication, and old GnuTLS with standard priority string and proper authentication.

There are three variables to consider here and I think trying to handle them all at once is just too difficult and confusing. I suggest the following code instead:

```c++
void CSecurityTLS::setParam()
{
  static const char kx_anon_priority[] = ":+ANON-ECDH:+ANON-DH";

  int ret;

  // Custom priority string specified?
  if (strcmp(Security::GnuTLSPriority, "") != 0) {
    char *prio;
    const char *err;

    prio = (char*)malloc(strlen(Security::GnuTLSPriority) +
                         strlen(kx_anon_priority) + 1);
    if (prio == NULL)
      throw AuthFailureException("Not enough memory for GnuTLS priority string");

    strcpy(prio, Security::GnuTLSPriority);
    if (anon)
      strcat(prio, kx_anon_priority);

    ret = gnutls_priority_set_direct(session, prio, &err);

    free(prio);

    if (ret != GNUTLS_E_SUCCESS) {
      if (ret == GNUTLS_E_INVALID_REQUEST)
        vlog.error("GnuTLS priority syntax error at: %s", err);
      throw AuthFailureException("gnutls_set_priority_direct failed");
    }
  } else if (anon) {
    const char *err;

#if GNUTLS_VERSION_NUMBER >= 0x030603
    ret = gnutls_set_default_priority_append(session, kx_anon_priority, &err, 0);
    if (ret != GNUTLS_E_SUCCESS)
      throw AuthFailureException("gnutls_set_default_priority_append failed");
#else
    // We don't know what the system default priority is, so we guess
    // it's what upstream GnuTLS has
    static const char gnutls_default_priority[] = "NORMAL";
    char *prio;

    prio = (char*)malloc(strlen(gnutls_default_priority) +
                         strlen(kx_anon_priority) + 1);
    if (prio == NULL)
      throw AuthFailureException("Not enough memory for GnuTLS priority string");

    strcpy(prio, gnutls_default_priority);
    strcat(prio, kx_anon_priority);

    ret = gnutls_priority_set_direct(session, prio, &err);

    free(prio);

    if (ret != GNUTLS_E_SUCCESS)
      throw AuthFailureException("gnutls_set_priority_direct failed");
#endif
  }
...
```

It has some redundant parts, but I think it is easier to follow and make sure everything is correct.

Also, don't forget to update the man pages.

CendioOssman (approved):

CendioOssman:
@grulja, I'm getting reports that this code isn't working. Could you test things in your end? I don't want to have a non-functional `master`.

grulja:
> @grulja, I'm getting reports that this code isn't working. Could you test things in your end? I don't want to have a non-functional `master`.

Sure. I will check on both RHEL 8 (new GnuTLS) and RHEL 7 (old GnuTLS).

grulja:
I can reproduce on RHEL8 and I think I know why.

From the documentation:
`The add_prio string should start as a normal priority string, e.g., ’-VERS-TLS-ALL:+VERS-TLS1.3:%COMPAT’ or ’%FORCE_ETM’. That is, it must not start with ’:’.`

However, the `kx_anon_priority` starts with `:` as it's appended to `NORMAL` in case of old GnuTLS. I will verify that's the case and submit a patch.

PR #1263, opened 2021-05-25
Make TLS work on FIPS systems

grulja:
This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1492107.

Steps to Reproduce:
1. set fips=1 on the kernel cmdline of the system hosting the VNC server 
2. boot with this setting and attempt to use vncviewer to connect to a system running a vnc daemon with FIPS turned on
3. Observe error:

### DIFF
--- a/common/rfb/SSecurityTLS.cxx
+++ b/common/rfb/SSecurityTLS.cxx
@@ -37,8 +37,6 @@
 #include <rdr/TLSOutStream.h>
 #include <gnutls/x509.h>
 
-#define DH_BITS 1024 /* XXX This should be configurable! */
-
 using namespace rfb;
 
 StringParameter SSecurityTLS::X509_CertFile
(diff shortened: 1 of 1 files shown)


CendioOssman:
Seems reasonable. But please also remove the define as it is now no longer used.

And perhaps also break things up a bit. That line ended up being very long. :)

CendioOssman:
Did you see this:

https://groups.google.com/g/tigervnc-devel/c/GHL650DS1II

Are you guys targeting different editions of FIPS?

grulja:
> Are you guys targeting different editions of FIPS?

CC: @nmav

This change was needed for RHEL 7, things might have changed since then.

nmav:
> > Are you guys targeting different editions of FIPS?
> 
> CC: @nmav
> 
> This change was needed for RHEL 7, things might have changed since then.

What Razbe writes makes sense, not only for FIPS, but as a good security practice. Including hard-coded rfc7919 parameters on  pre-3.6.0 versions makes sense as well.

jasonsikes:
> Did you see this:
> 
> https://groups.google.com/g/tigervnc-devel/c/GHL650DS1II
> 
> Are you guys targeting different editions of FIPS?

Sort of, yes. 

A few years ago FIPS required parameters to be at least 2048 bits. Newer FIPS requirements say parameters have to be selected from RFC 7919 or RFC 3526 (and also be at least 2048 bits).

CendioOssman:
So close this in favour of #1273?

grulja:
> So close this in favour of #1273?

Yes.

PR #1269, opened 2021-06-09
Add rfc7919 for FIPS.

jasonsikes:
This is a fix to call gnutls using the 2048-bit DH parameters from RFC-7919. This allows TigerVNC to be used in FIPS 140-3-compliant systems.

### DIFF
--- a/common/rfb/SSecurityTLS.cxx
+++ b/common/rfb/SSecurityTLS.cxx
@@ -37,7 +37,21 @@
 #include <rdr/TLSOutStream.h>
 #include <gnutls/x509.h>
 
-#define DH_BITS 1024 /* XXX This should be configurable! */
+/* FFDHE (RFC-7919) 2048-bit parameters, PEM-encoded */
+static unsigned char ffdhe2048[] =
+  "-----BEGIN DH PARAMETERS-----\n"
+  "MIIBDAKCAQEA//////////+t+FRYortKmq/cViAnPTzx2LnFg84tNpWp4TZBFGQz\n"
+  "+8yTnc4kmz75fS/jY2MMddj2gbICrsRhetPfHtXV/WVhJDP1H18GbtCFY2VVPe0a\n"
+  "87VXE15/V8k1mE8McODmi3fipona8+/och3xWKE2rec1MKzKT0g6eXq8CrGCsyT7\n"
+  "YdEIqUuyyOP7uWrat2DX9GgdT0Kj3jlN9K5W7edjcrsZCwenyO4KbXCeAvzhzffi\n"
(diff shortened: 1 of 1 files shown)


CendioOssman:
Looks good. Thanks.

Should this be conditional on just old versions of GnuTLS though? If the default behaviour changes in the future then this code will force an older behaviour.

jasonsikes:
I don't see an easy answer to this. It wouldn't be as simple as:
```
if GnuTLS version is at least 3.6.0:
    use GnuTLS's new "default" behavior
else
    use hardcoded parameters
```
If we tried then the code would look ugly. Switching behavior involves changing code in several places. I think it's best we decide one way or another before we commit a merge.

My arguments for this pull request:
* Even though we are using GnuTLS deprecated functions, I believe they will be supported for a few more years.
* TigerVNC will not require a newer version of GnuTLS.

My arguments for using the changes in https://groups.google.com/g/tigervnc-devel/c/GHL650DS1II
* GnuTLS 3.6.0 was introduced in August 2017, almost four years ago.
* Removes deprecated function calls.
* I think anyone who is using GnuTLS older that 3.6.0 is also using an older version of TigerVNC.

Wow. Now that I wrote all this I'm inclined to suggest this pull request be rejected.

CendioOssman:
> I don't see an easy answer to this. It wouldn't be as simple as:

Oh? How come? Can't we just put an `if` around the call to `gnutls_dh_params_import_pkcs3()`? What else would be needed?

> I think anyone who is using GnuTLS older that 3.6.0 is also using an older version of TigerVNC.

You'd think, but we regularly see people wanting the latest and greatest TigerVNC on their ancient systems. :)

For reference, neither Ubuntu 18.04, RHEL 7 or SLES 12 have GnuTLS 3.6.0.

jasonsikes:
> > I don't see an easy answer to this. It wouldn't be as simple as:
> 
> Oh? How come? Can't we just put an `if` around the call to `gnutls_dh_params_import_pkcs3()`? What else would be needed?
> 
> > I think anyone who is using GnuTLS older that 3.6.0 is also using an older version of TigerVNC.
> 
> You'd think, but we regularly see people wanting the latest and greatest TigerVNC on their ancient systems. :)
> 
> For reference, neither Ubuntu 18.04, RHEL 7 or SLES 12 have GnuTLS 3.6.0.

I stand corrected.

1. It is my personal preference to avoid C "if" macros where I can, but sometimes they are necessary.
2. I see a similar C macro in another file in TigerVNC (CSecurityTLS.cxx).
3. It's really not that ugly.

I'm closing this and creating a new pull request.

CendioOssman:
Continued in #1273.

PR #1273, opened 2021-06-13
Use RFC7919-2048 group in GnuTLS for FIPS compliance.

jasonsikes:
FIPS 140-3 only allows DH parameters from RFC3526 or RFC7919. Generating parameters randomly is no longer allowed.

GnuTLS starting from version 3.6.0 defaults to using 2048-bit RFC7919 parameters; this is FIPS-compliant. To use it we simply don't specify parameters.

For GnuTLS older than 3.6.0 we need to specify parameters. The same parameters are hard-coded.

The GnuTLS version is checked at compile time to decide which path to take.

### DIFF
--- a/common/rfb/SSecurityTLS.cxx
+++ b/common/rfb/SSecurityTLS.cxx
@@ -38,7 +38,23 @@
 #include <rdr/TLSOutStream.h>
 #include <gnutls/x509.h>
 
-#define DH_BITS 1024 /* XXX This should be configurable! */
+#if defined (SSECURITYTLS__USE_DEPRECATED_DH)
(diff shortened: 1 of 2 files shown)


CendioOssman:
Thanks. Any reason for a new define and not just check GNUTLS_VERSION_NUMBER where needed?

(And FYI, you can force push to an existing PR instead of creating a new one)

jasonsikes:
Thanks for the hint about continuing an existing pull request. This is my first time contributing to someone else's project on GiHub.

As for creating a new define: I think it helps readability. If a programmer has a question about why the 'ifdef' is there then they can see the comment where it is defined.

PR #1282, opened 2021-06-30
Add support for fullscreen over multiple, but not all monitors

hugolundin:
This pull request adds support to TigerVNC’s vncviewer for managing and using fullscreen over multiple, but not all monitors, as requested by issue #91. I would like to thank @ercanal for their work on this feature, giving me hints for how to implement this. 

# Overview

The implementation adds support for configuring what monitors to use in fullscreen, either via a configuration file, command line or the graphical options menu (`Options > Screen`). The aim is to make a robust solution that will be understandable for both regular and technical users. I have researched how other applications solve this problem and will present what I have done below. I will also describe parts of the implementation and what is left to be done before this can be considered ready. 

# Configuration

No matter the experience or technical knowledge, it should be convenient and easy for a user to configure what monitors they want to use. We also want to maintain the previously existing settings behaviour to not neglect any existing use-cases (full screen over the current monitor and full screen over all monitors). 

## Monitor identification

With the motivation that it should be easy to configure what monitors that are selected, we want to have a simple and robust way to identify monitors. In the GUI I have chosen to do it visually - the user simply clicks on the monitor they want to use.

In the configuration file and on the command line, I do not think we want to use the index that is provided by FLTK. Its stability relies on the implementation details of window managers and the library itself and might break a user configuration if it were to change. Instead, I propose that we have our own mapping that corresponds to the coordinates of the system monitor arrangement. 

The solution I have chosen is that the monitors should always be enumerated (starting at 1) from left to right, and if there is a conflict, from top to bottom. This index is entered into the configuration file or on the command line and is also the representation stored from what is configured in the GUI. 

## Graphical User Interface

A new widget (`MonitorArrangementWidget`) has been added to the bottom of `Options > Screen`. It allows the user to select which monitors to use. Additional information about a particular monitor can be obtained by hovering over the monitor. 

Because TigerVNC is a cross-platform application, I wanted to make sure that the user reliably can identify a display no matter what API:s are available. Position and size can be obtained from FLTK on all platforms, which is the primary way the user can identify a monitor graphically. The information displayed when hovering, varies depending on what platform the user is on. 

Both Windows, macOS and Gnome use blue color to indicate a selected state. Therefore, I think it might be reasonable to argue that it should be the case here as well. 

The position and size of a monitor *should* always correspond to what has been configured on the system. See for example my configuration in GNOME Settings and the screenshots of my implementation below: 

---

gnome

**Image 1**: My GNOME Settings display arrangement. 

---

selected_monitors

**Image 2**: An example of how the implemented GUI looks. Two monitors are selected. 

---

required_monitors

**Image 3**: The frame buffer is required to be *a* rectangle. Therefore, if the two monitors to the left and right are selected, the one in the middle will be implicitly included in fullscreen. In order to communicate to the user that this isn't a bug, but a requirement, we add a checkered pattern to that monitor. 

---

hover_mouse

**Image 4**: Further information that can be used to identify a monitor is shown when hovering over it. The information varies depending on the operating system and whether XRANDR is available (Linux). The resolution and coordinates should always be there, because it comes directly from FLTK. 

---

all_monitors

**Image 5**: When another mode is used,  the monitor selection configuration is disabled. 

---

disabled_state

**Image 6**: When fullscreen is not  used, the configuration panel also disabled. It could be argued that this shouldn't be the case, because the user might want to configure what settings should be used when they enter full screen mode later on. 

## CLI and config file 

Two new flags have been added; `FullScreenSelectedMonitors` and `FullScreenMode`. 

`FullScreenSelectedMonitors` can be set to a list of monitors that should be selected. Indices that are invalid or non-existent will be ignored. If the same index is repeated, the duplicates will be ignored. 

`FullScreenMode` can either be `Current`, `All` or `Selected`. The default mode is `Current`, corresponding to the old default fullscreen mode. `All` deprecates the existing flag `FullScreenAllMonitors=1`, to avoid invalid states, such as having enabled `FullScreenAllMonitors` and fullscreen on selected monitors simultaneously. 

Examples: 

```bash
$ vncviewer --FullScreenMode=Selected --FullScreenSelectedMonitors=1,3
$ vncviewer --FullScreenMode=Current
$ vncviewer --FullScreenMode=All
```

# TODO

- [x] Maintain the behaviour of existing configuration options. 
    - [x] `fullScreen = 1`
    - [x] `fullScreenAllMonitors`
- [x] Additional information on hover
    - [x] Linux 
    - [x] macOS
    - [x] Windows
- [x] Test the sorting of monitor indices, to make sure that it is correct with different configurations and operating systems.
- [x] Add documentation of new flags to the man pages.
- [x] Add default value for `fullScreenSelectedMonitors`.
- [x] Verify that all strings are localized. 
...

## Known issues

- [x] Grabbing will probably not work on macOS (due to `vncviewer/DesktopWindow.cxx:908`).
- [x] The “Press F8 to open the context menu”-overlay, displayed after connecting, is off-center (due to `vncviewer/DesktopWindow.cxx:467-501`).
- [x] `DesktopWindow: Failure grabbing keyboard`
- [x] ~~`XRequest.131: BadWindow (invalid Window parameter) 0x4a00079`~~ (also happens on the master branch, triggered by exiting vncviewer by pressing F8 and then enter with the mouse on a different screen than the active session).
- [x] ~~`Touch:       Unable to get X Input 2 event mask for window 0x04a00079`~~
- [x] macOS have negative monitor coordinates, which breaks the GUI

# Testing and verification

Acceptance criterias that can be useful for testing and verification can be found here:  [Bug 7006 - full screen over multiple, but not all monitors](https://www.cendio.com/bugzilla/show_bug.cgi?id=7006).

- [x] Linux
    - [x] Gnome
    - [x] KDE (probably not supported due to missing fullscreen features?)
    - [x] X11
    - [x] Wayland
- [x] Windows
- [x] macOS
…

---

The lists will be continuously updated as I continue my work. I gladly accept thoughts, feedback and code review of the implementation! :)

### DIFF
--- a/vncviewer/CMakeLists.txt
+++ b/vncviewer/CMakeLists.txt
@@ -63,6 +63,11 @@ elseif(APPLE)
   target_link_libraries(vncviewer "-framework IOKit")
 else()
   target_link_libraries(vncviewer ${X11_Xi_LIB})
+
+  if(X11_Xrandr_LIB)
+    add_definitions(-DHAVE_XRANDR)
+    target_link_libraries(vncviewer ${X11_Xrandr_LIB})
+  endif()
 endif()
 
 install(TARGETS vncviewer DESTINATION ${CMAKE_INSTALL_FULL_BINDIR})
--- a/vncviewer/CMakeLists.txt
+++ b/vncviewer/CMakeLists.txt
@@ -17,6 +17,8 @@ set(VNCVIEWER_SOURCES
   parameters.cxx
   keysym2ucs.c
   touch.cxx
+  MonitorArrangement.cxx
+  MonitorIndicesParameter.cxx
   vncviewer.cxx)
 
 if(WIN32)
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -465,9 +465,34 @@ void DesktopWindow::draw()
 
     // Make sure it's properly seen by adjusting it relative to the
     // primary screen rather than the entire window
-    if (fullscreen_active() && fullScreenAllMonitors) {
+    if (fullscreen_active()) {
       assert(Fl::screen_count() >= 1);
-      Fl::screen_xywh(sx, sy, sw, sh, 0);
+
+      rfb::Rect window_rect, screen_rect;
+      window_rect.setXYWH(x(), y(), w(), h());
+
+      bool found_enclosed_screen = false;
+      for (int i = 0; i < Fl::screen_count(); i++) {
+        Fl::screen_xywh(sx, sy, sw, sh, i);
+
+        // The screen with the smallest index that are enclosed by
+        // the viewport will be used for showing the overlay.
+        screen_rect.setXYWH(sx, sy, sw, sh);
+        if (screen_rect.enclosed_by(window_rect)) {
+          found_enclosed_screen = true;
+          break;
+        }
+      }
+
+      // If no monitor inside the viewport was found,
+      // use the one primary instead.
+      if (!found_enclosed_screen)
+        Fl::screen_xywh(sx, sy, sw, sh, 0);
+
+      // Adjust the coordinates so they are relative to the viewport.
+      sx -= x();
+      sy -= y();
+
     } else {
       sx = 0;
       sy = 0;
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -932,7 +932,7 @@ void DesktopWindow::grabKeyboard()
 #elif defined(__APPLE__)
   int ret;
   
-  ret = cocoa_capture_display(this, fullScreenAllMonitors);
+  ret = cocoa_capture_displays(this);
   if (ret != 0) {
     vlog.error(_("Failure grabbing keyboard"));
     return;
@@ -988,7 +988,7 @@ void DesktopWindow::ungrabKeyboard()
 #if defined(WIN32)
   win32_disable_lowlevel_keyboard(fl_xid(this));
 #elif defined(__APPLE__)
-  cocoa_release_display(this);
+  cocoa_release_displays(this);
 #else
   // FLTK has a grab so lets not mess with it
   if (Fl::grab())
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -885,9 +885,10 @@ void DesktopWindow::fullscreen_on()
   bool all_monitors = !strcasecmp(fullScreenMode, "all");
   bool selected_monitors = !strcasecmp(fullScreenMode, "selected");
 
-  if (not selected_monitors and not all_monitors)
-    fullscreen_screens(-1, -1, -1, -1);
-  else {
+  if (not selected_monitors and not all_monitors) {
+    int n = Fl::screen_num(x(), y(), w(), h());
+    fullscreen_screens(n, n, n, n);
+  } else {
     int top, bottom, left, right;
     int top_y, bottom_y, left_x, right_x;
 
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -880,10 +880,11 @@ int DesktopWindow::fltkHandle(int event, Fl_Window *win)
   return ret;
 }
 
-
 void DesktopWindow::fullscreen_on()
 {
-  if (not fullScreenAllMonitors)
+  bool all_monitors = !strcasecmp(fullScreenMode, "all");
+
+  if (not all_monitors)
     fullscreen_screens(-1, -1, -1, -1);
   else {
     int top, bottom, left, right;
@@ -1377,7 +1378,7 @@ void DesktopWindow::handleOptions(void *data)
     self->ungrabKeyboard();
 
   // Call fullscreen_on even if active since it handles
-  // fullScreenAllMonitors
+  // fullScreenMode
   if (fullScreen)
     self->fullscreen_on();
   else if (!fullScreen && self->fullscreen_active())
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -883,8 +883,9 @@ int DesktopWindow::fltkHandle(int event, Fl_Window *win)
 void DesktopWindow::fullscreen_on()
 {
   bool all_monitors = !strcasecmp(fullScreenMode, "all");
+  bool selected_monitors = !strcasecmp(fullScreenMode, "selected");
 
-  if (not all_monitors)
+  if (not selected_monitors and not all_monitors)
     fullscreen_screens(-1, -1, -1, -1);
   else {
     int top, bottom, left, right;
@@ -892,30 +893,57 @@ void DesktopWindow::fullscreen_on()
 
     int sx, sy, sw, sh;
 
-    top = bottom = left = right = 0;
+    std::set<int> monitors;
 
-    Fl::screen_xywh(sx, sy, sw, sh, 0);
+    if (selected_monitors and not all_monitors) {
+      std::set<int> selected = fullScreenSelectedMonitors.getParam();
+      monitors.insert(selected.begin(), selected.end());
+    } else {
+      for (int i = 0; i < Fl::screen_count(); i++) {
+        monitors.insert(i);
+      }
+    }
+
+    // If no monitors were found in the selected monitors case, we want
+    // to explicitly use the window's current monitor.
+    if (monitors.size() == 0) {
+      monitors.insert(Fl::screen_num(x(), y(), w(), h()));
+    }
+
+    // If there are monitors selected, calculate the dimensions
+    // of the frame buffer, expressed in the monitor indices that
+    // limits it.
+    std::set<int>::iterator it = monitors.begin();
+
+    // Get first monitor dimensions.
+    Fl::screen_xywh(sx, sy, sw, sh, *it);
+    top = bottom = left = right = *it;
     top_y = sy;
     bottom_y = sy + sh;
     left_x = sx;
     right_x = sx + sw;
 
-    for (int i = 1;i < Fl::screen_count();i++) {
-      Fl::screen_xywh(sx, sy, sw, sh, i);
+    // Keep going through the rest of the monitors.
+    for (; it != monitors.end(); it++) {
+      Fl::screen_xywh(sx, sy, sw, sh, *it);
+
       if (sy < top_y) {
-        top = i;
+        top = *it;
         top_y = sy;
       }
+
       if ((sy + sh) > bottom_y) {
-        bottom = i;
+        bottom = *it;
         bottom_y = sy + sh;
       }
+
       if (sx < left_x) {
-        left = i;
+        left = *it;
         left_x = sx;
       }
+
       if ((sx + sw) > right_x) {
-        right = i;
+        right = *it;
         right_x = sx + sw;
       }
     }
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -608,6 +608,11 @@ void DesktopWindow::resize(int x, int y, int w, int h)
 
     repositionWidgets();
   }
+
+  // Some systems require a grab after the window size has been changed.
+  // Otherwise they might hold on to displays, resulting in them being unusable.
+  if (fullscreen_active() && fullscreenSystemKeys)
+    grabKeyboard();
 }
 
 
--- a/vncviewer/MonitorArrangement.cxx
+++ b/vncviewer/MonitorArrangement.cxx
@@ -1,0 +1,303 @@
+/* Copyright 2021 Hugo Lundin <huglu@cendio.se> for Cendio AB.
+ * 
+ * This is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ * 
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License
+ * along with this software; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ * USA.
+ */
+
+#include <set>
+#include <vector>
+#include <string>
+#include <utility>
+#include <sstream>
+#include <assert.h>
+#include <algorithm>
+
+#include <FL/Fl.H>
+#include <FL/fl_draw.H>
+#include <FL/Fl_Button.H>
+
+#include <rfb/Rect.h>
+
+#include "MonitorArrangement.h"
+
+static const Fl_Boxtype FL_CHECKERED_BOX = FL_FREE_BOXTYPE;
+
+MonitorArrangement::MonitorArrangement(
+   int x, int y, int w, int h)
+:  Fl_Group(x, y, w, h),
+   SELECTION_COLOR(fl_lighter(FL_BLUE)),
+   AVAILABLE_COLOR(fl_lighter(fl_lighter(fl_lighter(FL_BACKGROUND_COLOR)))),
+   m_monitors()
+{
+  // Used for required monitors.
+  Fl::set_boxtype(FL_CHECKERED_BOX, checkered_pattern_draw, 0, 0, 0, 0);
+
+  box(FL_DOWN_BOX);
+  color(fl_lighter(FL_BACKGROUND_COLOR));
+  layout();
+  end();
+}
+
+MonitorArrangement::~MonitorArrangement()
+{
+
+}
+
+std::set<int> MonitorArrangement::get()
+{
+  std::set<int> indices;
+
+  for (int i = 0; i < (int) m_monitors.size(); i++) {
+    if (m_monitors[i]->value() == 1)
+      indices.insert(i);
+  }
+
+  return indices;
+}
+
+void MonitorArrangement::set(std::set<int> indices)
+{
+  for (int i = 0; i < (int) m_monitors.size(); i++) {
+    bool selected = std::find(indices.begin(), indices.end(), i) != indices.end();
+    m_monitors[i]->value(selected ? 1 : 0);
+  }
+}
+
+void MonitorArrangement::draw()
+{
+  for (int i = 0; i < (int) m_monitors.size(); i++) {
+    Fl_Button * monitor = m_monitors[i];
+
+    if (is_required(i)) {
+      monitor->box(FL_CHECKERED_BOX);
+      monitor->color(SELECTION_COLOR);
+    } else {
+      monitor->box(FL_BORDER_BOX);
+      monitor->color(AVAILABLE_COLOR);
+      monitor->selection_color(SELECTION_COLOR);
+    }
+  }
+
+  Fl_Group::draw();
+}
+
+void MonitorArrangement::layout()
+{
+  int x, y, w, h;
+  double scale = this->scale();
+  const double MARGIN_SCALE_FACTOR = 0.99;
+  std::pair<int, int> offset = this->offset();
+
+  for (int i = 0; i < Fl::screen_count(); i++) {
+    Fl::screen_xywh(x, y, w, h, i);
+
+    Fl_Button *monitor = new Fl_Button(
+      /* x = */ this->x() + offset.first + x*scale + (1 - MARGIN_SCALE_FACTOR)*x*scale,
+      /* y = */ this->y() + offset.second +  y*scale + (1 - MARGIN_SCALE_FACTOR)*y*scale,
+      /* w = */ w*scale*MARGIN_SCALE_FACTOR,
+      /* h = */ h*scale*MARGIN_SCALE_FACTOR
+    );
+
+    monitor->clear_visible_focus();
+    monitor->callback(monitor_pressed, this);
+    monitor->type(FL_TOGGLE_BUTTON);
+    monitor->when(FL_WHEN_CHANGED);
+    m_monitors.push_back(monitor);
+  }
+}
+
+bool MonitorArrangement::is_required(int m)
+{
+  // A selected monitor is never required.
+  if (m_monitors[m]->value() == 1)
+    return false;
+
+  // If no monitors are selected, none are required.
+  std::set<int> selected = get();
+  if (selected.size() <= 0)
+    return false;
+
+
+  // Go through all selected monitors and find the monitor
+  // indices that bounds the fullscreen frame buffer. If
+  // the given monitor's coordinates are inside the bounds,
+  // while not being selected, it is instead required.
+
+  int x, y, w, h;
+  int top_y, bottom_y, left_x, right_x;
+  std::set<int>::iterator it = selected.begin();
+
+  // Base the rest of the calculations on the dimensions
+  // obtained for the first monitor.
+  Fl::screen_xywh(x, y, w, h, *it);
+  top_y = y;
+  bottom_y = y + h;
+  left_x = x;
+  right_x = x + w;
+
+  // Go through the rest of the monitors,
+  // exhausting the rest of the iterator.
+  for (; it != selected.end(); it++) {
+    Fl::screen_xywh(x, y, w, h, *it);
+
+    if (y < top_y) {
+      top_y = y;
+    }
+
+    if ((y + h) > bottom_y) {
+      bottom_y = y + h;
+    }
+
+    if (x < left_x) {
+      left_x = x;
+    }
+
+    if ((x + w) > right_x) {
+      right_x = x + w;
+    }
+  }
+
+  rfb::Rect viewport, monitor;
+  viewport.setXYWH(left_x, top_y, right_x - left_x, bottom_y - top_y);
+
+  Fl::screen_xywh(x, y, w, h, m);
+  monitor.setXYWH(x, y, w, h);
+
+  return monitor.enclosed_by(viewport);
+}
+
+double MonitorArrangement::scale()
+{
+  const int MARGIN = 20;
+  std::pair<int, int> size = this->size();
+
+  double s_w = static_cast<double>(this->w()-MARGIN) / static_cast<double>(size.first);
+  double s_h = static_cast<double>(this->h()-MARGIN) / static_cast<double>(size.second);
+
+  // Choose the one that scales the least, in order to
+  // maximize our use of the given bounding area.
+  if (s_w > s_h)
+    return s_h;
+  else
+    return s_w;
+}
+
+std::pair<int, int> MonitorArrangement::size()
+{
+  int x, y, w, h;
+  int top, bottom, left, right;
+  int x_min, x_max, y_min, y_max;
+  x_min = x_max = y_min = y_max = 0;
+
+  for (int i = 0; i < Fl::screen_count(); i++) {
+    Fl::screen_xywh(x, y, w, h, i);
+
+    top = y;
+    bottom = y + h;
+    left = x;
+    right = x + w;
+
+    if (top < y_min)
+      y_min = top;
+
+    if (bottom > y_max)
+      y_max = bottom;
+
+    if (left < x_min)
+      x_min = left;
+
+    if (right > x_max)
+      x_max = right;
+  }
+
+  return std::make_pair(x_max - x_min, y_max - y_min);
+}
+
+std::pair<int, int> MonitorArrangement::offset()
+{
+  double scale = this->scale();
+  std::pair<int, int> size = this->size();
+  std::pair<int, int> origin = this->origin();
+
+  int offset_x = (this->w()/2) - (size.first/2 * scale);
+  int offset_y = (this->h()/2) - (size.second/2 * scale);
+
+  return std::make_pair(offset_x + abs(origin.first)*scale, offset_y + abs(origin.second)*scale);
+}
+
+std::pair<int, int> MonitorArrangement::origin()
+{
+  int x, y, w, h, ox, oy;
+  ox = oy = 0;
+
+  for (int i = 0; i < Fl::screen_count(); i++) {
+    Fl::screen_xywh(x, y, w, h, i);
+
+    if (x < ox)
+      ox = x;
+
+    if (y < oy)
+      oy = y;
+  }
+
+  return std::make_pair(ox, oy);
+}
+
+void MonitorArrangement::monitor_pressed(Fl_Widget *widget, void *user_data)
+{
+  MonitorArrangement *self = (MonitorArrangement *) user_data;
+
+  // When a monitor is selected, FLTK changes the state of it for us.
+  // However, selecting a monitor might implicitly change the state of
+  // others (if they become required). FLTK only redraws the selected
+  // monitor. Therefore, we must trigger a redraw of the whole widget
+  // manually.
+  self->redraw();
+}
+
+void MonitorArrangement::checkered_pattern_draw(
+  int x, int y, int width, int height, Fl_Color color)
+{
+  bool draw_checker = false;
+  const int CHECKER_SIZE = 8;
+
+  fl_color(fl_lighter(fl_lighter(fl_lighter(color))));
+  fl_rectf(x, y, width, height);
+
+  fl_color(Fl::draw_box_active() ? color : fl_inactive(color));
+
+  // Round up the square count. Later on, we remove square area that are
+  // outside the given bounding area.
+  const int count = (width + CHECKER_SIZE - 1) / CHECKER_SIZE;
+
+  for (int i = 0; i < count; i++) {
+    for (int j = 0; j < count; j++) {
+
+      draw_checker = (i + j) % 2 == 0;
+
+      if (draw_checker) {
+        fl_rectf(
+          /* x = */ x + i * CHECKER_SIZE,
+          /* y = */ y + j * CHECKER_SIZE,
+          /* w = */ CHECKER_SIZE - std::max(0, ((i + 1) * CHECKER_SIZE) - width),
+          /* h = */ CHECKER_SIZE - std::max(0, ((j + 1) * CHECKER_SIZE) - height)
+        );
+      }
+    }
+  }
+
+  fl_color(Fl::draw_box_active() ? FL_BLACK : fl_inactive(FL_BLACK));
+  fl_rect(x, y, width, height);
+}
--- a/vncviewer/MonitorArrangement.cxx
+++ b/vncviewer/MonitorArrangement.cxx
@@ -28,10 +28,31 @@
 #include <FL/fl_draw.H>
 #include <FL/Fl_Button.H>
 
-#include <rfb/Rect.h>
+#if defined(HAVE_XRANDR) && !defined(__APPLE__)
+#include <X11/extensions/Xrandr.h>
+#include <X11/X.h>
+#include <X11/Xlib.h>
+#include <X11/Xutil.h>
+#endif
 
+#ifdef __APPLE__
+#include <Carbon/Carbon.h>
+#include <IOKit/graphics/IOGraphicsLib.h>
+#include <IOKit/hidsystem/IOHIDLib.h>
+#include <IOKit/hidsystem/IOHIDParameter.h>
+#endif
+
+#include <rfb/Rect.h>
+#include <rfb/LogWriter.h>
+
+#ifdef WIN32
+#include "win32.h"
+#endif
+
+#include "i18n.h"
 #include "MonitorArrangement.h"
 
+static rfb::LogWriter vlog("MonitorArrangement");
 static const Fl_Boxtype FL_CHECKERED_BOX = FL_FREE_BOXTYPE;
 
 MonitorArrangement::MonitorArrangement(
@@ -116,6 +137,9 @@ void MonitorArrangement::layout()
     monitor->when(FL_WHEN_CHANGED);
     m_monitors.push_back(monitor);
   }
+
+  for (int i = 0; i < (int) m_monitors.size(); i++)
+    m_monitors[i]->copy_tooltip(description(i).c_str());
 }
 
 bool MonitorArrangement::is_required(int m)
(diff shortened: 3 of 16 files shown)


vncviewer/CMakeLists.txt
```diff
@@ -53,6 +55,13 @@ endif()
 
 target_link_libraries(vncviewer rfb network rdr os ${FLTK_LIBRARIES} ${GETTEXT_LIBRARIES})
 
+if(X11_FOUND AND X11_Xrandr_LIB)
+  add_definitions(-DHAVE_XRANDR)
+  target_link_libraries(vncviewer ${X11_Xrandr_LIB})
+else()
+  message(WARNING "No Xrandr extension.  vncviewer will not be able to identify monitors.")
+endif()
```
CendioOssman, line 63: This is just an X11 thing so should be conditional on platform.

vncviewer/DesktopWindow.cxx
```diff
@@ -465,9 +470,9 @@ void DesktopWindow::draw()
 
     // Make sure it's properly seen by adjusting it relative to the
     // primary screen rather than the entire window
-    if (fullscreen_active() && fullScreenAllMonitors) {
+    if (fullscreen_active() && (!strcmp(fullScreenMode, "Selected") || !strcmp(fullScreenMode, "All") || fullScreenAllMonitors)) {
```
CendioOssman, line 473: This is getting a bit long and difficult to read. Can we split things up better?

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,220 @@
...
+
+#include "MonitorArrangement.h"
+
+#define MONITOR_ARRANGEMENT_MARGIN 20
+#define MONITOR_MARGIN_SCALE_FACTOR 0.99
+#define MONITOR_AVAILABLE_COLOR fl_lighter(fl_lighter(fl_lighter(FL_BACKGROUND_COLOR)))
+#define MONITOR_SELECTED_COLOR fl_rgb_color(53, 132, 228)
+#define MONITOR_REQUIRED_COLOR fl_lighter(fl_lighter(fl_rgb_color(53, 132, 228)))
+
+#define FL_CHECKERED_BOX (Fl_Boxtype) FL_FREE_BOXTYPE
```
CendioOssman, line 41: `static const` is preferred so we get some type checking.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,220 @@
...
+#include <FL/Fl_Group.H>
+#include <FL/fl_draw.H>
+#include <FL/Fl_PNG_Image.H>
+#include <FL/Fl_Check_Browser.H>
+
+#include "MonitorArrangement.h"
+
+#define MONITOR_ARRANGEMENT_MARGIN 20
+#define MONITOR_MARGIN_SCALE_FACTOR 0.99
+#define MONITOR_AVAILABLE_COLOR fl_lighter(fl_lighter(fl_lighter(FL_BACKGROUND_COLOR)))
```
CendioOssman, line 37: This won't work as a `static const` though as it will be evaluated too early. I would prefer some other construction that keeps type checking though rather than using a `#define`.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,220 @@
...
+
+MonitorArrangement::MonitorArrangement(
+    int x, int y, int w, int h, MonitorArrangementDelegate *delegate
+): Fl_Group(x, y, w, h), m_delegate(delegate)
+{
+  box(FL_DOWN_BOX);
+  color(fl_lighter(FL_BACKGROUND_COLOR));
+
+  // Register a custom boxtype for the required monitor appearance.
+  Fl::set_boxtype(FL_CHECKERED_BOX, checkered_pattern_draw, 0, 0, 0, 0);
```
CendioOssman, line 56: Can this be called multiple times? I.e. if we would have multiple instances of this widget.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,220 @@
...
+    style(i);
+  }
+
+  Fl_Group::draw();
+}
+
+void MonitorArrangement::show()
+{
+    Fl_Group::show();
+}
```
CendioOssman, line 127: Why is this needed?

vncviewer/MonitorArrangement.h
```diff
@@ -0,0 +1,75 @@
...
+#include <FL/Fl.H>
+#include <FL/Fl_Window.H>
+#include <FL/Fl_Box.H>
+#include <FL/Fl_Button.H>
+#include <FL/Fl_Check_Button.H>
+#include <FL/Fl_Timer.H>
+#include <FL/Fl_Group.H>
+#include <FL/fl_draw.H>
+#include <FL/Fl_PNG_Image.H>
+#include <FL/Fl_Check_Browser.H>
```
CendioOssman, line 31: This looks very excessive. Only include the absolute minimum headers needed to avoid massive dependency chains.

vncviewer/Monitors.h
```diff
@@ -0,0 +1,140 @@
...
+
+#define DESCRIPTION_MAX_LEN 256
+
+enum ConfigurationMode {
+  CURRENT,
+  ALL,
+  SELECTED
+};
+
+class Monitors: public MonitorArrangementDelegate {
```
CendioOssman, line 37: Not a huge fan of this class. It seems to duplicate state, and take on multiple roles for various sections. It makes it difficult to follow how things are connected.

Have you looked at trying to split out what it does to the other components and try to make them more independent? E.g. avoid having a callback mechanism in the widget and instead have a classical `set()`/`get()` style API? And avoiding duplicating state in here by just giving the current string parameters to a helper function when it is time to calculate relevant monitors?

vncviewer/OptionsDialog.cxx
```diff
@@ -296,10 +302,20 @@ void OptionsDialog::loadOptions(void)
...
+
   fullScreenCheckbox->value(fullScreen);
-  fullScreenAllMonitorsCheckbox->value(fullScreenAllMonitors);
+
+  if (!strcmp(fullScreenMode, "All")) {
+    allButton->setonly();
+  } else if (!strcmp(fullScreenMode, "Selected")) {
+    selectedButton->setonly();
+  } else {
+    defaultButton->setonly();
```
CendioOssman, line 313: Try to be more specific than "default" as that is very ambiguous.

vncviewer/OptionsDialog.cxx
```diff
@@ -407,7 +423,18 @@ void OptionsDialog::storeOptions(void)
...
   remoteResize.setParam(remoteResizeCheckbox->value());
   fullScreen.setParam(fullScreenCheckbox->value());
-  fullScreenAllMonitors.setParam(fullScreenAllMonitorsCheckbox->value());
+
+  if (allButton->value()) {
+    fullScreenMode.setParam("All");
+  } else if (selectedButton->value()) {
+    fullScreenMode.setParam("Selected");
+  } else {
+    fullScreenMode.setParam("");
```
CendioOssman, line 432: I think every valid mode should have a proper name.

vncviewer/OptionsDialog.cxx
```diff
@@ -296,10 +302,20 @@ void OptionsDialog::loadOptions(void)
     desktopHeightInput->value(buf);
   }
   remoteResizeCheckbox->value(remoteResize);
+
   fullScreenCheckbox->value(fullScreen);
-  fullScreenAllMonitorsCheckbox->value(fullScreenAllMonitors);
+
+  if (!strcmp(fullScreenMode, "All")) {
```
CendioOssman, line 308: We should probably not be case sensitive.

vncviewer/OptionsDialog.cxx
```diff
@@ -902,6 +975,28 @@ void OptionsDialog::handleClipboard(Fl_Widget *widget, void *data)
...
 
+void OptionsDialog::handleFullScreen(Fl_Widget *widget, void *data)
+{
+  OptionsDialog *dialog = (OptionsDialog*)data;
+
+  if (dialog->fullScreenCheckbox->value()) {
+    dialog->fullScreenModeGroup->activate();
+  } else {
+    dialog->fullScreenModeGroup->deactivate();
+  }
```
CendioOssman, line 986: Hmm... so we didn't do this before? Perhaps we shouldn't do it now either. If someone wants to configure things in preparation for toggling fullscreen later?

vncviewer/OptionsDialog.h
```diff
@@ -22,6 +22,8 @@
 #include <map>
 
 #include <FL/Fl_Window.H>
+#include "Monitors.h"
+#include "MonitorArrangement.h"
```
CendioOssman, line 26: These are not necessary and should be avoided. Look at how we handle the FLTK widget classes.

CendioOssman (changes requested):

vncviewer/MonitorArrangement.cxx
hugolundin: `FL_CHECKERED_BOX` is an alias for `FL_FREE_BOXTYPE`  which is a pointer to the first free boxtype slot. Calling this function will set the callback for that type, so calling it repeatedly should not be a problem.   

However, another potential issue here is that if someone were to add another custom boxtype in the future it must be an alias to `FL_FREE_BOXTYPE+1`. Otherwise, this boxtype will break, when another callback is registered to the same free slot. I think that argues for that we might benefit from defining and registering all boxtypes in the same place.

vncviewer/MonitorArrangement.cxx
hugolundin: It isn't, now removed.

vncviewer/MonitorArrangement.h
hugolundin: I agree, now fixed.

vncviewer/MonitorArrangement.cxx
hugolundin: Moved into `MonitorArrangement` as const members.

vncviewer/MonitorArrangement.cxx
hugolundin: After discussing this with @CendioOssman, we decided to keep `FL_CHECKERED_BOX` defined here for now.

vncviewer/Monitors.h
hugolundin: The monitors class has been removed. It has been replaced with a more classical `set()`/`get()` style API on `MonitorArrangement`. Additionally, `MonitorIndicesParameter` (a subclass of `rfb::StringParameter` ) has  been added, which is responsible for mapping between our external index representation and the FLTK indices that is used internally. Overall, I would like to state that the structure is much more clear now.

vncviewer/CMakeLists.txt
```diff
@@ -53,6 +55,8 @@ endif()
 
 target_link_libraries(vncviewer rfb network rdr os ${FLTK_LIBRARIES} ${GETTEXT_LIBRARIES})
 
+
+
```
CendioOssman, line 59: Oops? :)

vncviewer/CMakeLists.txt
```diff
@@ -61,6 +65,11 @@ elseif(APPLE)
   target_link_libraries(vncviewer "-framework IOKit")
 else()
   target_link_libraries(vncviewer ${X11_Xi_LIB})
+
+  if(X11_FOUND AND X11_Xrandr_LIB)
```
CendioOssman, line 69: Can we get here without `X11_FOUND`?

vncviewer/DesktopWindow.cxx
```diff
@@ -850,50 +856,60 @@ int DesktopWindow::fltkHandle(int event, Fl_Window *win)
...
 
-
 void DesktopWindow::fullscreen_on()
 {
-  if (not fullScreenAllMonitors)
-    fullscreen_screens(-1, -1, -1, -1);
-  else {
-    int top, bottom, left, right;
+  int top, bottom, left, right;
+  std::set<int> selected = fullScreenSelectedMonitors.get();
```
CendioOssman, line 862: Shouldn't you check the mode first?

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,414 @@
...
+
+#include <set>
+#include <vector>
+#include <string>
+#include <utility>
+#include <sstream>
+#include <assert.h>
+#include <algorithm>
+
+#ifdef HAVE_XRANDR
```
CendioOssman, line 27: Does this need a macOS guard as well? X11 headers could be available there.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,414 @@
...
+{
+
+}
+
+std::set<int> MonitorArrangement::get()
+{
+  std::set<int> indices;
+
+  for (int i = 0; i < Fl::screen_count(); i++) {
+    if (m_monitors[i]->value() == 1) {
```
CendioOssman, line 70: What if there are now more monitors than when you created the widget?

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,414 @@
...
+    m_monitors[i]->copy_tooltip(description(i).c_str());
+  }
+}
+
+bool MonitorArrangement::is_required(int m)
+{
+  // A selected monitor is never required. 
+  if (m_monitors[m]->value() == 1) {
+    return false;
+  }
```
CendioOssman, line 138: Minor style issue, but in TigerVNC we usually leave out the curlies for a single statement.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,414 @@
...
+
+  // Get the platform specific name of the monitor. 
+  ss << name(m);
+
+  if (ss.str().empty()) {
+    int x, y, w, h;
+    Fl::screen_xywh(x, y, w, h, m);
+
+    // Fallback to showing resolution and position of monitor.
+    ss << w << "x" << h << "+" << x << "+" << y;
```
CendioOssman, line 298: The X11 geometry syntax is probably unfamiliar to most so I don't think it helps much here. Perhaps just resolution is sufficient?

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    for (int j = 0; j < crtc->noutput; j++) {
+      bool monitor_found = (crtc->x == x) &&
+          (crtc->y == y) &&
+          (crtc->width == ((unsigned int) w)) &&
+          (crtc->height == ((unsigned int) h));
+
+      if (monitor_found) {
+        XRROutputInfo *output = XRRGetOutputInfo(fl_display, res, crtc->outputs[j]);
```
CendioOssman, line 345: Do we need to check monitors rather than outputs to support the latest RandR?

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,414 @@
...
+
+std::string MonitorArrangement::name(int m)
+{
+  assert(m < (int) m_monitors.size());
+
+  int x, y, w, h;
+  std::stringstream ss;
+  Fl::screen_xywh(x, y, w, h, m);
+
+#if !defined(WIN32) && !defined(__APPLE__)
```
CendioOssman, line 312: This looks redundant with the checks below.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,414 @@
...
+
+#ifdef HAVE_XRANDR
+#include <X11/extensions/Xrandr.h>
+#include <X11/X.h>
+#include <X11/Xlib.h>
+#include <X11/Xutil.h>
+#endif
+
+#include <rfb/LogWriter.h>
+#include <rfb/CMsgWriter.h>
```
CendioOssman, line 35: This seems odd. What is it needed for?

vncviewer/MonitorIndicesParameter.cxx
```diff
@@ -0,0 +1,170 @@
...
+        vlog.error("No monitors found.");
+        return indices;
+    }
+
+    // sscanf will modify the parsed string. Therefore
+    // we make a copy of it before parsing.
+    const char* config = getValueStr();
+
+    while (*config) {
+        if (1 == sscanf(config, "%d%n", &value, &count)) {
```
CendioOssman, line 52: Syntax errors?

vncviewer/MonitorIndicesParameter.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <set>
+#include <vector>
+#include <rfb/Configuration.h>
+
+class MonitorIndicesParameter: public rfb::StringParameter {
+public:
+    MonitorIndicesParameter(const char* name_, const char* desc_, const char* v);
+    std::set<int> get();
+    bool set(std::set<int> indices);
```
CendioOssman, line 30: Probably best to follow the existing style of `getParam()`/`setParam()`.

vncviewer/OptionsDialog.cxx
```diff
@@ -46,6 +48,7 @@
 #include <FL/Fl_Round_Button.H>
 #include <FL/Fl_Int_Input.H>
 #include <FL/Fl_Choice.H>
+#include <FL/Fl_Check_Browser.H>
```
CendioOssman, line 51: What is this used for?

vncviewer/OptionsDialog.h
```diff
@@ -125,7 +129,12 @@ class OptionsDialog : public Fl_Window {
   Fl_Int_Input *desktopHeightInput;
   Fl_Check_Button *remoteResizeCheckbox;
   Fl_Check_Button *fullScreenCheckbox;
-  Fl_Check_Button *fullScreenAllMonitorsCheckbox;
+
+  Fl_Group *fullScreenModeGroup;
```
CendioOssman, line 133: Do we need to keep track of this group?

CendioOssman (changes requested):

vncviewer/MonitorArrangement.cxx
hugolundin: Nothing, removed.

vncviewer/OptionsDialog.cxx
hugolundin: Nothing, removed.

vncviewer/OptionsDialog.h
hugolundin: Previously, it was needed because I disabled that group when another option was chosen for fullscreen. This has been changed, so it can be removed.

vncviewer/MonitorArrangement.cxx
hugolundin: Future improvement, added to the acceptance criterias for now.

vncviewer/DesktopWindow.cxx
hugolundin: Indeed, fixed.

vncviewer/MonitorIndicesParameter.cxx
hugolundin: The parsing is now more robust and more validation has been added.

a1exh:
Looking forward to a Windows binary to test. Good work this is a feature that TigerVNC has needed for many years.

vncviewer/DesktopWindow.cxx
```diff
@@ -462,18 +461,29 @@ void DesktopWindow::draw()
...
+
+    for (int i = 0; i < Fl::screen_count(); i++) {
+      Fl::screen_xywh(sx, sy, sw, sh, i);
+
+      // The screen with the smallest index that are enclosed by
+      // the viewport will be used for showing the overlay.
+      screen_rect.setXYWH(sx, sy, sw, sh);
+      if (screen_rect.enclosed_by(viewport_rect)) {
+        break;
+      }
```
CendioOssman, line 476: Code style thing again. :)

vncviewer/DesktopWindow.cxx
```diff
@@ -462,18 +461,29 @@ void DesktopWindow::draw()
...
 
-    // Make sure it's properly seen by adjusting it relative to the
-    // primary screen rather than the entire window
-    if (fullscreen_active() && fullScreenAllMonitors) {
-      assert(Fl::screen_count() >= 1);
+      // If no monitor inside the viewport was found,
+      // use the one primary instead.
       Fl::screen_xywh(sx, sy, sw, sh, 0);
```
CendioOssman, line 480: Isn't this run on every loop now?

vncviewer/DesktopWindow.cxx
```diff
@@ -462,18 +461,29 @@ void DesktopWindow::draw()
...
+      // The screen with the smallest index that are enclosed by
+      // the viewport will be used for showing the overlay.
+      screen_rect.setXYWH(sx, sy, sw, sh);
+      if (screen_rect.enclosed_by(viewport_rect)) {
+        break;
+      }
 
-    // Make sure it's properly seen by adjusting it relative to the
-    // primary screen rather than the entire window
-    if (fullscreen_active() && fullScreenAllMonitors) {
```
CendioOssman, line 468: What about windowed mode?

vncviewer/DesktopWindow.cxx
```diff
@@ -850,50 +860,74 @@ int DesktopWindow::fltkHandle(int event, Fl_Window *win)
...
+          monitors.insert(i);
+        }
       }
-    }
 
-    fullscreen_screens(top, bottom, left, right);
+      // If there are monitors selected, calculate the dimensions
+      // of the frame buffer, expressed in the monitor indices that
+      // limits it. 
+      if (monitors.size() > 0) {
```
CendioOssman, line 885: Do we log selected but missing monitors at some point? That would be useful if users are getting problems.

vncviewer/DesktopWindow.h
```diff
@@ -26,7 +26,6 @@
 
 #include <rfb/Rect.h>
 #include <rfb/Pixel.h>
-
```
CendioOssman, line 29: Please keep this spacing. It provides some group of internal and external stuff.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,475 @@
...
+#endif
+
+#include <rfb/LogWriter.h>
+#include <FL/Fl.H>
+#include <FL/fl_draw.H>
+
+#ifdef __APPLE__
+#include <CoreFoundation/CoreFoundation.h>
+#include <objc/objc.h>
+#include <objc/objc-runtime.h>
```
CendioOssman, line 41: Aren't these implicitly pulled in by `CoreFoundation.h`? I don't think I've ever seen these explicitly included.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,475 @@
...
+#include "i18n.h"
+#include "MonitorArrangement.h"
+
+static rfb::LogWriter vlog("MonitorArrangement");
+static const Fl_Boxtype FL_CHECKERED_BOX = FL_FREE_BOXTYPE;
+
+MonitorArrangement::MonitorArrangement(
+   int x, int y, int w, int h)
+:  Fl_Group(x, y, w, h),
+   SELECTION_COLOR(fl_rgb_color(53, 132, 228)),
```
CendioOssman, line 62: Shouldn't this be some FL constant?

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,475 @@
...
+    /* options = */ kIODisplayOnlyPreferredName);
+
+  CFDictionaryRef dict = (CFDictionaryRef) CFDictionaryGetValue(info, CFSTR(kDisplayProductName));
+  CFIndex dict_len = CFDictionaryGetCount(dict);
+
+  if (dict_len > 0) {
+    CFTypeRef * names = new CFTypeRef[dict_len];
+    CFDictionaryGetKeysAndValues(dict, NULL, (const void **) names);
+
+    if ((sizeof(names) / sizeof(*names)) > 0) {
```
CendioOssman, line 344: Won't this always be true? We've already checked `dict_len`.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,475 @@
...
+  Fl::screen_xywh(x, y, w, h, m);
+
+  if (!XQueryExtension(fl_display, "RANDR", &xi_major, &ev, &err)) {
+    vlog.info(_("Failed to get monitor name because X11 RandR could not be found."));
+    return -1;
+  }
+
+  XRRScreenResources *res = XRRGetScreenResources(fl_display, DefaultRootWindow(fl_display));
+  if (!res) {
+    vlog.error(_("Failed to get XRRScreenResources for Window 0x%08lx"), DefaultRootWindow(fl_display));
```
CendioOssman, line 390: It's always just one root window, so you don't need to specify it.

vncviewer/MonitorArrangement.h
```diff
@@ -0,0 +1,86 @@
...
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ * USA.
+ */
+
+#ifndef __MONITOR_ARRANGEMENT_H__
+#define __MONITOR_ARRANGEMENT_H__
+
+#include <string>
+
+#include <FL/Fl_Button.H>
```
CendioOssman, line 24: I don't think you need to include this. You only have pointers to it.

vncviewer/MonitorIndicesParameter.cxx
```diff
@@ -0,0 +1,209 @@
...
+
+bool MonitorIndicesParameter::setParam(const char* value)
+{
+    std::set<int> indices;
+
+    if (strlen(value) <= 0)
+        return false;
+
+    if (!parse_indices(value, indices)) {
+        vlog.error(_("Parsing failed for FullScreenSelectedMonitors."));
```
CendioOssman, line 77: Let's assume it's not the code that's buggy and instead state that it's the value that is invalid. :)

Also, it would be helpful to specify _why_ it is invalid.

vncviewer/MonitorIndicesParameter.cxx
```diff
@@ -0,0 +1,209 @@
...
+    std::vector<MonitorIndicesParameter::Monitor> monitors = this->monitors();
+
+    if (monitors.size() <=  0) {
+        vlog.error(_("Failed to get monitors."));
+        // Don't return, store the configuration anyways.
+    }
+
+    for (int i = 0; i < ((int) monitors.size()); i++) {
+        if (std::find(indices.begin(), indices.end(), monitors[i].fltk_index) != indices.end())
+            config_indices.insert(i);
```
CendioOssman, line 99: If it isn't found then some warning should ideally be printed. Otherwise it is just magically lost.

vncviewer/MonitorIndicesParameter.cxx
```diff
@@ -0,0 +1,209 @@
...
+}
+
+static bool parse_number(std::string number, std::set<int>& indices)
+{
+    if (number.size() <= 0)
+        return false;
+
+    int v = strtol(number.c_str(), NULL, 0);
+
+    if (v < 0 || v > std::numeric_limits<int>::max()) {
```
CendioOssman, line 130: This may be more C++-y, but `INT_MAX` is a lot more readable. :)

vncviewer/MonitorIndicesParameter.cxx
```diff
@@ -0,0 +1,209 @@
...
+        vlog.error(_("The given index (%s) is too large to be valid."), number.c_str());
```
CendioOssman, line 131: The reader of this doesn't know the context of this message, so you need to specify that it is a screen index.

vncviewer/MonitorIndicesParameter.cxx
```diff
@@ -0,0 +1,209 @@
...
+            (*index)+1
+        );
+
+        separator = ",";
+    }
+
+    return setParam(buf);
+}
+
+static bool parse_number(std::string number, std::set<int>& indices)
```
CendioOssman, line 123: We've tried to avoid using `&` if the variable gets modified as you can't really tell from the call site.

vncviewer/OptionsDialog.h
```diff
@@ -66,6 +67,9 @@ class OptionsDialog : public Fl_Window {
 
   static void handleClipboard(Fl_Widget *widget, void *data);
 
+  static void handleFullScreen(Fl_Widget *widget, void *data);
```
CendioOssman, line 70: This is gone, right?

vncviewer/cocoa.mm
```diff
@@ -29,12 +29,14 @@
 
 #include <IOKit/hidsystem/IOHIDLib.h>
 #include <IOKit/hidsystem/IOHIDParameter.h>
+#import <IOKit/graphics/IOGraphicsLib.h>
```
CendioOssman, line 32: `#include` seems to work for the others, so probably best to be consistent.

vncviewer/parameters.cxx
```diff
@@ -101,10 +101,14 @@ IntParameter qualityLevel("QualityLevel",
                           8);
 
 BoolParameter maximize("Maximize", "Maximize viewer window", false);
-BoolParameter fullScreen("FullScreen", "Full screen mode", false);
+BoolParameter fullScreen("FullScreen", "Enable full screen", false);
+StringParameter fullScreenMode("FullScreenMode", "Full screen mode", "Current");
```
CendioOssman, line 105: This text is used with `--help`, so it probable needs to be more descriptive.

vncviewer/parameters.cxx
```diff
@@ -101,10 +101,14 @@ IntParameter qualityLevel("QualityLevel",
...
 BoolParameter fullScreenAllMonitors("FullScreenAllMonitors",
                                     "Enable full screen over all monitors",
-                                    true);
+                                    false);
+MonitorIndicesParameter fullScreenSelectedMonitors("FullScreenSelectedMonitors",
+                                         "Enable full screen over the given selected monitors",
```
CendioOssman, line 110: Same here.

vncviewer/parameters.cxx
```diff
@@ -101,10 +101,14 @@ IntParameter qualityLevel("QualityLevel",
                           8);
 
 BoolParameter maximize("Maximize", "Maximize viewer window", false);
-BoolParameter fullScreen("FullScreen", "Full screen mode", false);
+BoolParameter fullScreen("FullScreen", "Enable full screen", false);
+StringParameter fullScreenMode("FullScreenMode", "Full screen mode", "Current");
 BoolParameter fullScreenAllMonitors("FullScreenAllMonitors",
                                     "Enable full screen over all monitors",
-                                    true);
+                                    false);
```
CendioOssman, line 108: Should probably mention that this is deprecated. Or even hide it from `--help`.

vncviewer/parameters.cxx
```diff
@@ -705,39 +763,8 @@ char* loadViewerParameters(const char *filename) {
...
-            invalidParameterName = false;
-          }
-
-        } else {
-          vlog.error(_("Unknown parameter type for parameter %s"),
-                     parameterArray[i]->getName());
-        }
-      }
+      invalidParameterName = findAndSetViewerParameterFromValue(parameterArray, sizeof(parameterArray), value, line, lineNr, filepath)
+        && findAndSetViewerParameterFromValue(readOnlyParameterArray, sizeof(readOnlyParameterArray), value, line, lineNr, filepath);
```
CendioOssman, line 767: This line ended up very large and complex. Can we reformat things to make it more obvious what happens?

vncviewer/win32.c
```diff
@@ -418,3 +419,54 @@ int win32_has_altgr(void)
 
   return has_altgr;
 }
+
+typedef struct {
+	int x, y, w, h;
+  char* name;
```
CendioOssman, line 425: Some indentation bug here.

CendioOssman (changes requested):

vncviewer/DesktopWindow.cxx
hugolundin: Yep, definitely broken right now haha

vncviewer/MonitorArrangement.cxx
hugolundin: True, my thought behind this was to look whether `CFdictionaryGetKeysAndValues` actually filled the array with something. I realize that this check won't solve that and it has been replaced with another.

vncviewer/parameters.cxx
hugolundin: Hiding it from help might be good but requires structural changes to how parameters are listed. At the moment it might be easiest to annotate it as deprecated.

CendioOssman: I think `removeParam()` should do what you want.

vncviewer/MonitorIndicesParameter.cxx
hugolundin: I tried adding some more details about why the parsing went wrong.

williamsjoblom:
One minor thing that I would reconsider is the colors used to indicate which displays are selected and which are not. I find it somewhat confusing that the darker color (blue) is used to indicate a selected display when it is usually the other way around. One possible approach would be to replace the white color with dark grey.

I'm aware that `gnome-control-center` uses similar colors under the Adwaita light theme for their display selection. However, in their use case, only one display is selectable at a time, making this design choice less controversial.

hugolundin:
The commits have now been cleaned up (from 75 to 5... :sweat_smile:) and refined so that every commit performs independent changes, in an iterative manner.

vncviewer/DesktopWindow.cxx
```diff
@@ -467,7 +467,32 @@ void DesktopWindow::draw()
     // primary screen rather than the entire window
     if (fullscreen_active() && fullScreenAllMonitors) {
       assert(Fl::screen_count() >= 1);
-      Fl::screen_xywh(sx, sy, sw, sh, 0);
+
+      rfb::Rect viewport_rect, screen_rect;
+      viewport_rect.setXYWH(x(), y(), w(), h());
```
CendioOssman, line 471: Formally the "viewport" is the widget showing the VNC session, and you want the size of the window here. The viewport might be smaller or larger than the window depending on the session.

vncviewer/DesktopWindow.cxx
```diff
@@ -467,7 +467,32 @@ void DesktopWindow::draw()
     // primary screen rather than the entire window
     if (fullscreen_active() && fullScreenAllMonitors) {
```
CendioOssman: Can't this be simplified further now by just checking `fullscreen_active()`?

vncviewer/CMakeLists.txt
```diff
@@ -61,6 +61,11 @@ elseif(APPLE)
   target_link_libraries(vncviewer "-framework IOKit")
 else()
   target_link_libraries(vncviewer ${X11_Xi_LIB})
+
+  if(X11_Xrandr_LIB)
+    add_definitions(-DHAVE_XRANDR)
+    target_link_libraries(vncviewer ${X11_Xrandr_LIB})
+  endif()
```
CendioOssman, line 70: This seems to be in the wrong commit.

vncviewer/DesktopWindow.cxx
```diff
@@ -875,50 +873,62 @@ int DesktopWindow::fltkHandle(int event, Fl_Window *win)
...
-  else {
-    int top, bottom, left, right;
-    int top_y, bottom_y, left_x, right_x;
+  int top, bottom, left, right;
+  top = bottom = left = right = -1;
+  bool all_monitors = !strcasecmp(fullScreenMode, "all");
 
-    int sx, sy, sw, sh;
+  if (all_monitors) {
+      std::set<int> monitors;
```
CendioOssman, line 885: Incorrect indentation here, which is also making this diff more difficult to read then it has to be.

vncviewer/DesktopWindow.cxx
```diff
@@ -875,50 +873,62 @@ int DesktopWindow::fltkHandle(int event, Fl_Window *win)
...
+      for (int i = 0; i < Fl::screen_count(); i++) {
+        monitors.insert(i);
+      }
 
-    Fl::screen_xywh(sx, sy, sw, sh, 0);
-    top_y = sy;
-    bottom_y = sy + sh;
-    left_x = sx;
-    right_x = sx + sw;
+      int x, y, w, h;
```
CendioOssman: Could we keep these as `sx`, `sy`, ...? I think it makes the diff better so it's easier to see what is actually happening in this commit.

vncviewer/parameters.cxx
```diff
@@ -101,10 +101,13 @@ IntParameter qualityLevel("QualityLevel",
                           8);
 
 BoolParameter maximize("Maximize", "Maximize viewer window", false);
-BoolParameter fullScreen("FullScreen", "Full screen mode", false);
+BoolParameter fullScreen("FullScreen", "Enable full screen", false);
+StringParameter fullScreenMode("FullScreenMode", "Specify what mode to use when full screen is enabled. "
+                                                 "Should be either Current or All",
```
CendioOssman: It should probably be explicit what "Current" and "All" are referring to. So instead of "mode", perhaps write "Specify which monitors to use when...".

vncviewer/parameters.cxx
```diff
@@ -101,10 +101,13 @@ IntParameter qualityLevel("QualityLevel",
...
+                                                 "Current");
 BoolParameter fullScreenAllMonitors("FullScreenAllMonitors",
-                                    "Enable full screen over all monitors",
-                                    true);
+                                    "[DEPRECATED] Enable full screen over all monitors",
+                                    false);
```
CendioOssman, line 110: Did you test `removeParam()`?

vncviewer/vncviewer.man
```diff
@@ -205,9 +205,15 @@ Maximize viewer window.
 Start in full-screen mode.
 .
 .TP
-.B \-FullScreenAllMonitors
+.B \-FullScreenAllMonitors (DEPRECATED)
 Use all local monitors and not just the current one when switching to
-full-screen mode.
+full-screen mode. Replaced by \fB-FullScreenMode=all\fP
```
CendioOssman, line 210: If we can get rid of this from `--help` then it should also be removed from the man page IMO.

vncviewer/DesktopWindow.cxx
```diff
@@ -878,48 +879,60 @@ void DesktopWindow::fullscreen_on()
...
-      Fl::screen_xywh(x, y, w, h, *it);
-      top = bottom = left = right = *it;
-      top_y = y;
-      bottom_y = y + h;
-      left_x = x;
-      right_x = x + w;
+      // If there are monitors selected, calculate the dimensions
+      // of the frame buffer, expressed in the monitor indices that
+      // limits it. 
+      if (monitors.size() > 0) {
```
CendioOssman, line 899: Can we get rid of this extra indentation? It tends to obscure things.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,306 @@
...
+#include <FL/fl_draw.H>
+#include <FL/Fl_Button.H>
+
+#include <rfb/Rect.h>
+#include <rfb/LogWriter.h>
+
+#include "i18n.h"
+#include "MonitorArrangement.h"
+
+static rfb::LogWriter vlog("MonitorArrangement");
```
CendioOssman, line 55: This isn't used, so it and the include can be removed.

vncviewer/MonitorArrangement.cxx
```diff
@@ -0,0 +1,306 @@
...
+#include <algorithm>
+
+#include <FL/Fl.H>
+#include <FL/fl_draw.H>
+#include <FL/Fl_Button.H>
+
+#include <rfb/Rect.h>
+#include <rfb/LogWriter.h>
+
+#include "i18n.h"
```
CendioOssman, line 52: This isn't used as far as I can tell.

CendioOssman (changes requested):

vncviewer/CMakeLists.txt
hugolundin: Moved to the commit introducing tooltips.

vncviewer/MonitorArrangement.cxx
hugolundin: It is used, but in the commit introducing tooltips. Moved.

vncviewer/MonitorArrangement.cxx
hugolundin: Same here, moved.

vncviewer/parameters.cxx
hugolundin: `removeParam()` results in it being reject on command line too. We will keep it for now.

vncviewer/vncviewer.man
hugolundin: Same here. We will keep it for now.

vncviewer/DesktopWindow.cxx
hugolundin: Definitely. Fixed.

vncviewer/DesktopWindow.cxx
hugolundin: Renamed.

vncviewer/parameters.cxx
```diff
@@ -523,24 +553,8 @@ static char* loadFromReg() {
...
-    } else if (dynamic_cast<BoolParameter*>(parameterArray[i]) != NULL) {
-      if (getKeyInt(parameterArray[i]->getName(), &intValue, &hKey))
-        ((BoolParameter*)parameterArray[i])->setParam(intValue);
-    } else {      
-      vlog.error(_("Unknown parameter type for parameter %s"),
-                 parameterArray[i]->getName());
-    }
-  }
+  findAndSetViewerParametersFromReg(parameterArray, sizeof(parameterArray), &hKey);
+  findAndSetViewerParametersFromReg(readOnlyParameterArray, sizeof(readOnlyParameterArray), &hKey);
```
CendioOssman, line 562: Unfortunately this won't clean up the old variables, so they'll continue to be loaded.

vncviewer/cocoa.mm
```diff
@@ -49,31 +50,35 @@
...
+    CGDirectDisplayID displays[16];
 
-      if (CGGetActiveDisplayList(16, displays, &count) != kCGErrorSuccess)
-        return 1;
+    int sx, sy, sw, sh;
+    rfb::Rect viewport_rect, screen_rect;
 
-      if (count != (unsigned)Fl::screen_count())
-        return 1;
+    viewport_rect.setXYWH(win->x(), win->y(), win->w(), win->h());
```
CendioOssman, line 66: Another slightly incorrect viewport reference. :)

vncviewer/vncviewer.man
```diff
@@ -205,9 +205,14 @@ Maximize viewer window.
...
+.
+.TP
+.B \-FullScreenMode \fImode\fP
+Specify which monitors to use when in full screen. It should be either "current" or "all".
```
CendioOssman: The default value should be documented.

vncviewer/parameters.cxx
```diff
@@ -103,11 +103,15 @@ IntParameter qualityLevel("QualityLevel",
...
-                                                 "Should be either Current or All",
+                                                 "Should be either Current, Selected or All",
                                                  "Current");
 BoolParameter fullScreenAllMonitors("FullScreenAllMonitors",
                                     "[DEPRECATED] Enable full screen over all monitors",
                                     false);
+MonitorIndicesParameter fullScreenSelectedMonitors("FullScreenSelectedMonitors",
+                                         "Use the given list of monitors in full screen"
+                                         " when FullScreenMode = Selected.",
+                                         "");
```
CendioOssman, line 114: Should we have a default here, since empty is a bit undefined?

vncviewer/vncviewer.man
```diff
@@ -211,8 +211,15 @@ full-screen mode. Replaced by \fB-FullScreenMode=all\fP
...
-Specify which monitors to use when in full screen. It should be either "current" or "all". 
+Specify which monitors to use when in full screen. It should be either "current",
+"selected" (specified by \fB-FullScreenSelectedMonitors\fP) or "all".
 .
+.TP
+.B \-FullScreenSelectedMonitors \fImonitors\fP
+This option specifies the monitors to use with \fB-FullScreenMode=selected\fP.
+Monitors are ordered according to the system configuration from left to right,
+and in case of a conflict, from top to bottom. So, for example, "1,2,3" means
+that the first, second and third monitor counting from the left should be used.
```
CendioOssman, line 222: The default should be documented.

CendioOssman (changes requested):

CendioOssman (approved):

williamsjoblom:
I have tested a recent build on Fedora 33 against the acceptance criteria. This was done with two displays, so there are a few criteria that could not be tested. With these constraints considered, everything seems to work as intended!

CendioOssman (approved):

a1exh:
Great work getting this into master. Really looking forward to the next Windows beta to test this. Anywhere we can get nightly/weekly builds of Master for Windows x64?

bphinz:
On Fri, Jul 16, 2021 at 10:41 AM Alex Holland ***@***.***>
wrote:

> Great work getting this into master. Really looking forward to the next
> Windows beta to test this. Anywhere we can get nightly/weekly builds of
> Master for Windows x64?
>
The nightly for f1dfeecd has already been posted:

http://tigervnc.bphinz.com/nightly/

> --
Sent from Gmail Mobile

PR #1287, opened 2021-07-12
Fix inconsistent line endings in vncviewer configuration file

whenov:
This may be trivial, but it is a bit annoying seeing `^M` in first two lines of `default.tigervnc` in vim.


CendioOssman:
Thanks, that is indeed a bit odd. I suspect the idea was for it to be compatible between Windows and Linux. As such, we should probably check that this change doesn't break anything.

whenov:
@CendioOssman Thanks. Note that all other line endings used in the same source file (`parameters.cxx`) are `\n`.

CendioOssman:
This seems to be a safe change, and notepad on Windows is a lot more well behaved with Unix line endings now. So I'll go ahead and merge this. Thanks for your contribution!

PR #1289, opened 2021-07-14
Fix build on systems without HOST_NAME_MAX

arichardson:
Some operating systems such as FreeBSD don't define a HOST_NAME_MAX
macro. The portable approach to determine the real host name limit is
calling sysconf(_SC_HOST_NAME_MAX) so do that instead.

### DIFF
--- a/unix/vncserver/vncsession.c
+++ b/unix/vncserver/vncsession.c
@@ -339,7 +339,8 @@ static void
 redir_stdio(const char *homedir, const char *display)
 {
     int fd;
-    char hostname[HOST_NAME_MAX+1];
+    size_t hostlen;
+    char* hostname = NULL;
     char logfile[PATH_MAX];
 
     fd = open("/dev/null", O_RDONLY);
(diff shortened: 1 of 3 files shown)


CendioOssman:
Odd. I thought `HOST_NAME_MAX` was part of POSIX so that everyone implemented that...

That code is not in a global scope, so I guess calling `sysconf()` should be fine and we can avoid having fallbacks. Do you think you could update the patch to do that?

arichardson:
Updated to call sysconf(). To avoid using VLAs this also uses std::vector/malloc().

CendioOssman:
Sorry, things are a bit hectic on my end. I'll try to have a look at your PRs tomorrow and hopefully merge everything.

arichardson:
> Sorry, things are a bit hectic on my end. I'll try to have a look at your PRs tomorrow and hopefully merge everything.

Sure no rush, I can continue to build from my fork for now.

PR #1290, opened 2021-07-14
Allow building with -DBUILD_VIEWER=OFF

arichardson:
Without this change I get CMake errors due to use of ${FLTK_INCLUDE_DIR}
and references to the non-existent vncviewer target.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -229,31 +229,33 @@ if(BUILD_JAVA)
   add_subdirectory(java)
 endif()
 
-# Check for FLTK
-set(FLTK_SKIP_FLUID TRUE)
-set(FLTK_SKIP_OPENGL TRUE)
-set(FLTK_SKIP_FORMS TRUE)
-find_package(FLTK)
-
-if(UNIX AND NOT APPLE)
-  # No proper handling for extra X11 libs that FLTK might need...
-  if(X11_Xft_FOUND)
-    # Xft headers include references to fontconfig, so we need
-    # to link to that as well
-    find_library(FONTCONFIG_LIB fontconfig)
-    set(FLTK_LIBRARIES ${FLTK_LIBRARIES} ${X11_Xft_LIB} ${FONTCONFIG_LIB})
-  endif()
-  if(X11_Xinerama_FOUND)
-    set(FLTK_LIBRARIES ${FLTK_LIBRARIES} ${X11_Xinerama_LIB})
-  endif()
-  if(X11_Xfixes_FOUND)
-    set(FLTK_LIBRARIES ${FLTK_LIBRARIES} ${X11_Xfixes_LIB})
-  endif()
-  if(X11_Xcursor_FOUND)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -229,13 +229,18 @@ if(BUILD_JAVA)
...
+  set(FLTK_SKIP_FLUID TRUE)
+  set(FLTK_SKIP_OPENGL TRUE)
+  set(FLTK_SKIP_FORMS TRUE)
+  find_package(FLTK REQUIRED)
+else()
+  set(FLTK_FOUND FALSE)
+endif()
 
-if(UNIX AND NOT APPLE)
+if(FLTK_FOUND AND UNIX AND NOT APPLE)
```
CendioOssman, line 243: Can we move this up instead to simplify things?

release/CMakeLists.txt
```diff
@@ -15,10 +15,12 @@ endif()
 
 configure_file(tigervnc.iss.in tigervnc.iss)
 
-add_custom_target(installer
-  iscc -o. ${INST_DEFS} -F${CMAKE_PROJECT_NAME}${INST_SUFFIX}-${VERSION} tigervnc.iss
-  DEPENDS vncviewer
-  SOURCES tigervnc.iss)
+if(BUILD_VIEWER)
```
CendioOssman, line 18: Probably easier to make this entire directory conditional on the viewer being built.

CMakeLists.txt
```diff
@@ -229,13 +229,18 @@ if(BUILD_JAVA)
...
-find_package(FLTK)
+option(BUILD_VIEWER "Build TigerVNC viewer" ON)
+if(BUILD_VIEWER)
+  # Check for FLTK
+  set(FLTK_SKIP_FLUID TRUE)
+  set(FLTK_SKIP_OPENGL TRUE)
+  set(FLTK_SKIP_FORMS TRUE)
+  find_package(FLTK REQUIRED)
+else()
+  set(FLTK_FOUND FALSE)
```
CendioOssman, line 240: Can we avoid this? It was barely used before, and it is a bit odd compared to everything else we have.

We should be able to avoid it by using `BUILD_VIEWER` and `if(DEFINED FLTK_LIBRARIES)` instead.

CendioOssman (changes requested):
Seems reasonable enough. There are some minor tweaks I would appreciate if you could have a look at though.

May I ask what prompted this need? It's been like this for some time without anyone having issues.

arichardson:
> Seems reasonable enough. There are some minor tweaks I would appreciate if you could have a look at though.
> 
> May I ask what prompted this need? It's been like this for some time without anyone having issues.

I am trying to cross-compile Xvnc for RISC-V CheriBSD, a fork of FreeBSD with support for CHERI-extended architectures (using 128-bit capabilities for all pointers). I haven't got FLTK cross-compiled and building with `BUILD_VIEWER=False` gives me various CMake errors+warnings.

Will update the PR shortly.

arichardson:
ping? Does this look okay now?

PR #1291, opened 2021-07-14
Rename Input.h to vncInput.h to fix building on case-insensitive FS 

arichardson:
I am cross-compiling from macOS for a FreeBSD-derived system so my host
file system is case insensitive but the target isn't. Without this change
I get the following warnings which show that the vnc "Input.h" is being
included from mi/mi.h instead of the xserver "input.h":
```
In file included from /Users/alex/cheri/xvnc-server/hw/vnc/Input.c:33:
/Users/alex/cheri/xvnc-server/mi/mi.h:55:10: warning: non-portable path to file '"Input.h"'; specified path differs in case from file name on disk [-Wnonportable-include-path]
#include "input.h"
         ^~~~~~~~~
         "Input.h"
```

### DIFF
--- a/unix/xserver/hw/vnc/Makefile.am
+++ b/unix/xserver/hw/vnc/Makefile.am
@@ -13,12 +13,12 @@ noinst_LTLIBRARIES = libvnccommon.la
 HDRS = vncExtInit.h vncHooks.h \
 	vncBlockHandler.h vncSelection.h \
 	XorgGlue.h XserverDesktop.h xorg-version.h \
-	Input.h RFBGlue.h
+	vncInput.h RFBGlue.h
 
 libvnccommon_la_SOURCES = $(HDRS) \
 	vncExt.c vncExtInit.cc vncHooks.c vncSelection.c \
 	vncBlockHandler.c XorgGlue.c RandrGlue.c RFBGlue.cc XserverDesktop.cc \
-	Input.c InputXKB.c qnum_to_xorgevdev.c qnum_to_xorgkbd.c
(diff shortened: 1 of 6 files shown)


CendioOssman:
This shouldn't be an issue since we don't put that directory in the search path. Could you describe more how things go wrong?

(And what system are you on where Xvnc is relevant but the file system is case insensitive? :D)

arichardson:
> This shouldn't be an issue since we don't put that directory in the search path. Could you describe more how things go wrong?
> 
> (And what system are you on where Xvnc is relevant but the file system is case insensitive? :D)

I am cross-compiling from macOS for a FreeBSD-derived system so my host FS is case insensitive but the target isn't.
Without this change I get the following warnings which show that the vnc "Input.h" is being included instead of
 the xserver "input.h":
```
In file included from /Users/alex/cheri/xvnc-server/hw/vnc/Input.c:33:
/Users/alex/cheri/xvnc-server/mi/mi.h:55:10: warning: non-portable path to file '"Input.h"'; specified path differs in case from file name on disk [-Wnonportable-include-path]
#include "input.h"
         ^~~~~~~~~
         "Input.h"
```

arichardson:
force-pushed to update the commit message with more details.

CendioOssman:
Hmm... That's just wrong. But I can see the wrongness here as well. It seems to be a combination of automake silliness of adding `-I.`, and xserver using `#include "input.h"` where it should be using `#include <input.h>`. So I guess we have no choice but to do this...

Could you rename `InputXKB` as well as they are tightly connected. And remove that comment from the header as well as we now longer have a conflict with `input.h`.

arichardson:
ping? Does this look okay now?

PR #1293, opened 2021-07-15
Fix priority string when using newer GnuTLS

grulja:
The call of gnutls_set_default_priority_append() expects a normal priority string, which means it must not start with ':'.

Fixes issue introduced by https://github.com/TigerVNC/tigervnc/pull/1262.

### DIFF
--- a/common/rfb/CSecurityTLS.cxx
+++ b/common/rfb/CSecurityTLS.cxx
@@ -238,7 +238,9 @@ void CSecurityTLS::setParam()
     const char *err;
 
 #if GNUTLS_VERSION_NUMBER >= 0x030603
-    ret = gnutls_set_default_priority_append(session, kx_anon_priority, &err, 0);
+    // gnutls_set_default_priority_appends() expects a normal priority string that
+    // doesn't start with ":".
+    ret = gnutls_set_default_priority_append(session, kx_anon_priority + 1, &err, 0);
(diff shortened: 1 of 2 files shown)


grulja:
Or possibly I can have `kx_anon_priority[] = "+ANON-ECDH:+ANON-DH"` and just append `:` additionally in case the priority string is modified by the user or in case we have to use `NORMAL`.

The way I proposed it now is more understandable to me.

common/rfb/CSecurityTLS.cxx
```diff
@@ -238,7 +238,10 @@ void CSecurityTLS::setParam()
     const char *err;
 
 #if GNUTLS_VERSION_NUMBER >= 0x030603
-    ret = gnutls_set_default_priority_append(session, kx_anon_priority, &err, 0);
+    // gnutls_set_default_priority_appends() expects a normal priority string that
+    // doesn't start with ":".
+    static const char kx_anon_priority_normal[] = "+ANON-ECDH:+ANON-DH";
```
CendioOssman, line 243: I'll leave it up to you if you want to add or remove the `:`, but I'd like to avoid duplication if we can. So please try to use the same base string and modify things as needed.

CendioOssman (changes requested):

CendioOssman (approved):

CendioOssman:
Great. Thanks.

PR #1300, opened 2021-07-22
Add -interface option to x0vncserver

sotho:
This PR adds the "-interface" option found in the vncserver also to the x0vncserver.


CendioOssman:
Firewalls generally handle this much better than we can, but since we have it in Xvnc/libvnc.so I suppose it's okay to have the same behaviour in x0vncserver.

Could you update the man page though so this argument is documented?

sotho:
Done

CendioOssman:
Great, thanks.

PR #1301, opened 2021-07-27
fix name spelling error

phoebe-leong:


samhed:
> In American English, the noun is spelled the same as the verb - license. But in British English, the noun is spelled licence.

phoebe-leong:
oh right ty

PR #1302, opened 2021-07-27
make README markdown and add links to sections

phoebe-leong:


CendioOssman:
Github parses restructured text just fine, so I don't see what a conversion gives us. What is the issue you are trying to solve here?

phoebe-leong:
i think the links would be pretty good to add if not the markdown conversion

CendioOssman:
Okay, I see. However this document is meant to be included with the source, so references should ideally be local and not to an external page. The fact that GitHub also implicitly shows the README is just a bonus. Our primary web page is the one on tigervnc.org, not the README.

So I'm afraid I'm going to go ahead and close this PR, but I still want to thank you for your feedback.

PR #1330, opened 2021-09-12
When reconnecting don't ask for login data again

johahauf:


CendioOssman:
I'm a bit cautious about this. Hanging on to passwords longer than necessary has historically been a source of security issues. An opt-in would be nice. Perhaps a check box on the dialog asking for the password?

johahauf:
Yes, fully understand, I was also unsure about it, but just from personal use it is so convenient ;-)
Nice idea about the check box, would checking it by default be OK or should we save the last state in default.tigervnc/registry?

CendioOssman:
I think opt-in (i.e. unchecked) is the safer choice. I'm hoping we can live without the hassle of storing the choice. We don't really have any infrastructure for such things. The workflow should hopefully be `<password> <tab> <space> <enter>` to get the password remembered.

johahauf:
So like this?
image

Why not storing the last selection as part of the settings in the Windows registry or Linux default settings?

CendioOssman:
Yes, something like that. :)

The current settings storage system is based entirely around the configuration options. We currently don't have a way of storing anything other than configuration settings, and I'd like to avoid adding configuration options for "Don't ask me again" type of things as we can potentially have loads of those. So we need to build something new for those.

vncviewer/UserDialog.cxx
```diff
@@ -62,34 +63,54 @@ static void button_cb(Fl_Widget *widget, void *val) {
   widget->window()->hide();
 }
 
+static void keepPassword(Fl_Widget *widget, void *data) {
+  UserDialog *dialog = (UserDialog*) data;
+  Fl_Check_Button* flcb = (Fl_Check_Button*)widget;
+  dialog->setKeepPassword(flcb->value());
+}
```
CendioOssman, line 70: You could probably simplify this by simply checking the check box once the dialog is closed. We don't really need to know it's value until afterwards anyway.

vncviewer/UserDialog.cxx
```diff
@@ -62,34 +63,54 @@ static void button_cb(Fl_Widget *widget, void *val) {
...
+
 UserDialog::UserDialog()
 {
+  user_ = NULL;
+  password_ = NULL;
+  keepPasswd_ = true;
```
CendioOssman, line 76: Just a FYI, in C++ it's usually better to use initialisers for things like this. I.e.:

```c++
UserDialog::UserDialog() : user_(NULL), password_(NULL), keepPasswd_(true)
{
}
```

vncviewer/UserDialog.cxx
```diff
@@ -62,34 +63,54 @@ static void button_cb(Fl_Widget *widget, void *val) {
...
 
 UserDialog::~UserDialog()
 {
+  resetPassword();
+}
+
+void UserDialog::resetPassword()
+{
+  if(user_) { free(user_); user_ = NULL; }
+  if(password_) { free(password_); password_ = NULL; }
```
CendioOssman, line 87: We don't need to save lines, so I'd prefer if this is more readable by having things on their own lines. So the same style as the rest of the code. :)

vncviewer/UserDialog.cxx
```diff
@@ -62,34 +63,54 @@ static void button_cb(Fl_Widget *widget, void *val) {
...
+static void keepPassword(Fl_Widget *widget, void *data) {
+  UserDialog *dialog = (UserDialog*) data;
+  Fl_Check_Button* flcb = (Fl_Check_Button*)widget;
+  dialog->setKeepPassword(flcb->value());
+}
+
 UserDialog::UserDialog()
 {
+  user_ = NULL;
+  password_ = NULL;
```
CendioOssman, line 75: I would suggest more descriptive names on these, given that "user" and "password" occurs in multiple ways in this code.

vncviewer/UserDialog.cxx
```diff
@@ -62,34 +63,54 @@ static void button_cb(Fl_Widget *widget, void *val) {
...
+  if(password_) {
+    if(user_) *user = strdup(user_);
+    *password = strdup(password_);
     return;
   }
 
-  if (!user && envPassword) {
-    *password = strdup(envPassword);
+  user_ = getenv("VNC_USERNAME");
+  *password = getenv("VNC_PASSWORD");
```
CendioOssman, line 103: Any reason why not `password_` here? The lack of symmetry is a bit confusing.

vncviewer/UserDialog.cxx
```diff
@@ -62,34 +63,54 @@ static void button_cb(Fl_Widget *widget, void *val) {
...
 
   assert(password);
-  char *envUsername = getenv("VNC_USERNAME");
-  char *envPassword = getenv("VNC_PASSWORD");
 
-  if(user && envUsername && envPassword) {
-    *user = strdup(envUsername);
-    *password = strdup(envPassword);
+  if(password_) {
+    if(user_) *user = strdup(user_);
```
CendioOssman, line 97: This will crash. You need to check if `user` is valid.

(same thing a few other places)

vncviewer/UserDialog.cxx
```diff
@@ -164,6 +187,13 @@ void UserDialog::getUserPasswd(bool secure, char** user, char** password)
 
   y += 5;
 
+  keepPasswdCheckbox = new Fl_Check_Button(70, y, win->w()-70-10, 20,
+					   _("Keep password for reconnect"));
+  keepPasswdCheckbox->callback(keepPassword, this);
+  keepPasswdCheckbox->value(keepPasswd_);
```
CendioOssman, line 193: Reconnect is optional, so should we perhaps hide this if reconnect isn't enabled?

vncviewer/UserDialog.cxx
```diff
@@ -187,15 +217,19 @@ void UserDialog::getUserPasswd(bool secure, char** user, char** password)
   while (win->shown()) Fl::wait();
 
   if (ret_val == 0) {
-    if (user)
+    if (user) {
       *user = strDup(username->value());
+      user_ = strDup(username->value());
```
CendioOssman, line 222: You have to take care here with the difference between `strdup()` and `strDup()`. They use different allocators.

CendioOssman, line 222: This should also be if:ed with the value of the check box.

CendioOssman (changes requested):
Sorry for the delay. I've gone through it now and it's almost done. Just some minor tweaks left.

vncviewer/UserDialog.cxx
johahauf: not using `password_` because in `password_` we store the password only in case the checkbox is checked. The user name is always stored locally this is the reason for the lack of symmetry. Could change it in the way that `*user` is used for the `getenv` call.

vncviewer/UserDialog.cxx
johahauf: Thanks, did not realize this. Changed all occurrences to `strDup()` and `delete []`

johahauf: Why? any concerns about storing the user name in all cases?

CendioOssman: Should hopefully not be as sensitive as the password, but might still be some privacy aspect that people prefer this not being remembered.

I guess it's not a must to change this, but I think it might be easier to understand that it's all or nothing what the viewer remembers for credentials.

CendioOssman:
Sorry for dropping the ball on this. I've now merged an adjusted version of this as 633e75b5263cbde04a16a0f2c04165807db1bd91.

Issue #1065, opened 2020-07-21
vncviewer: The dialog for opening a configuration file could have a better default directory


erkkiruohtula:
Currently using the dialog to open a configuration file in vncviewer always sets the starting point at the root of the file system. This means the user may have to navigate several levels to find where the user prefers to keep the files.

The issue does not matter so much if the user always uses the same server, because Tigervnc remembers the latest server,  but someone like me who frequently connects to different servers finds this inconvenient.

It would be more usable if the starting point were either the user's home directory, or the latest directory used to load a configuration file. I think starting at home would be a trivial change to implement, from what I can see from the code.

(I'm now using Linux, but I recall Windows also started the file dialog at C:\).

samhed:
Agreed.

CendioOssman:
Fixed via #1331.

PR #1331, opened 2021-09-12
Adjust default directory for configuration file save and load

johahauf:
As suggested in Issue #1065 The dialog for opening a configuration file could have a better default directory

### DIFF
--- a/vncviewer/ServerDialog.cxx
+++ b/vncviewer/ServerDialog.cxx
@@ -23,6 +23,7 @@
 
 #include <errno.h>
 #include <algorithm>
+#include <libgen.h>
 
 #include <FL/Fl.H>
 #include <FL/Fl_Input.H>
@@ -67,6 +68,7 @@ ServerDialog::ServerDialog()
   y = margin;
   
   serverName = new Fl_Input_Choice(x, y, w() - margin*2 - server_label_width, INPUT_HEIGHT, _("VNC server:"));
+  usedDir = NULL;
 
   int adjust = (w() - 20) / 4;
   int button_width = adjust - margin/2;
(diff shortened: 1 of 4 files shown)


vncviewer/ServerDialog.cxx
```diff
@@ -53,6 +54,23 @@ static LogWriter vlog("ServerDialog");
...
+static void updateUsedDir(const char* filename)
+{
+#ifndef  _WIN32
+  char * name = strdup(filename);
+  usedDir = strdup(dirname(name));
+  free(name);
+#else
+  char dir[_MAX_DIR], drive[_MAX_DRIVE], path[_MAX_PATH];
+  if (_splitpath_s(filename, drive, _MAX_DRIVE, dir, _MAX_DIR, NULL, 0, NULL, 0) == 0) {
+    if (_makepath_s( path, _MAX_PATH, drive, dir, NULL, NULL) == 0) {
```
CendioOssman, line 67: `dirname()` seems to be provided by mingw, so I don't think we need this alternative for Windows?

vncviewer/ServerDialog.cxx
```diff
@@ -53,6 +54,23 @@ static LogWriter vlog("ServerDialog");
 
 const char* SERVER_HISTORY="tigervnc.history";
 
+static char* usedDir = NULL;
+static void updateUsedDir(const char* filename)
```
CendioOssman, line 58: Can't this be part of the class?

CendioOssman (changes requested):
This looks like a nice addition, thanks! I've got a couple of notes though.

vncviewer/ServerDialog.cxx
johahauf: Yes, you are right, it is working. Will remove it.

PR #1342, opened 2021-09-21
Add vncviewer -UseLocalCursor option to fix UltraVNC invisible cursor problem

ajschorr:
Please consider merging this patch to add a -UseLocalCursor option to vncviewer so that it will work again
with the UltraVNC server.

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -185,8 +185,9 @@ Viewport::Viewport(int w, int h, const rfb::PixelFormat& serverPF, CConn* cc_)
 
   OptionsDialog::addCallback(handleOptions, this);
 
-  // Make sure we have an initial blank cursor set
-  setCursor(0, 0, rfb::Point(0, 0), NULL);
+  if (!useLocalCursor)
+    // Make sure we have an initial blank cursor set
+    setCursor(0, 0, rfb::Point(0, 0), NULL);
 }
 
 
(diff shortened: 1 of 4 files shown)


CendioOssman:
I'm afraid this is not something I see having worth more options for. This is something that should be fixed in UltraVNC, and until then we already have the dot cursor as a workaround.

ajschorr:
I have gotten absolutely no response from the UltraVNC folks regarding this issue. Do you have any idea how to get their attention? I seem to be caught in the middle between the two projects, and nobody seems to care enough to fix this.

https://github.com/ultravnc/UltraVNC/issues/24

Do you have a relationship with anybody at UltraVNC that would allow you to raise this issue with them? I really don't think the dot cursor is a great workaround. The visibility is very poor.

CendioOssman:
It's just been two days. Unfortunately you'll have to have a bit more patience than that when it comes to open source projects. For many this is a hobby after all. :)

ajschorr:
I've been an open-source contributor and maintainer for decades, so am not new to this process.  My main concern  is that I don't actually know how to explain the problem properly to the UltraVNC folks. If you could please add a comment to that issue explaining what you think is wrong with their server, that could be very helpful. Thank you.

RudiDeVos:
Did a little research:
Seems TigerVNC send the pseudoEncoding before the actual encoding causing UltraVNC to ignore it.
Changed our code so that it's cached.
UltraVNC now send proper the cursor shapes.

TigerVNC support rfbEncodingXCursor/rfbEncodingRichCursor but not rfbEncodingPointerPos
The server can send the shapes, but can't send a position move, if the cursor moves on the server it isn't visible on the viewer, for that rfbEncodingPointerPos also need to be implemented.

https://www.uvnc.eu/download/1342/winvnc_TigerCursorTest.zip

PR #1344, opened 2021-09-23
Only grab keyboard when we have focus

### LINKED ISSUE #1332: Full screen issue with new beta on OsX
I tried out the new beta on my mac:
MacBook Pro (13-inch, 2017) running macOS Mojave 10.14.6
---
If i go full screen on my external monitor (independently of how i choose the monitor setting in Options menu) what happens is that i have no way to get back to other OsX apps (running on different displays) in any way.
With previous version i just had to click on any OsX app and focus would go to OsX and there i could switch to other apps (including full screen TigerVnc) clicking on Alt+Tab.
With this new version this is no longer possible as, apparently, once going to full screen, TigerVnc occupies all three monitors just for the active app in the top menu (independent of all possible combinations of settings for full screen in Option menu) and therefore if i click on an OsX app this one will gain focus for a very short period of time and will switch back to TigerVnc almost immediately (< 1 second).
Needless to say that this "feature" makes full screen mode (even on a single monitor) completely useless and therefore i cannot use TigerVnc at all.
Only way to switch between apps is to exit Full Screen mode and then everything works as expected.

With current stable version this does not happen, forced to stick to that one

samhed:
Closes #1332

### DIFF
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -1007,9 +1007,23 @@ Bool eventIsFocusWithSerial(Display *display, XEvent *event, XPointer arg)
 }
 #endif
 
+bool DesktopWindow::hasFocus()
+{
+  Fl_Widget* focus;
+
+  focus = Fl::grab();
+  if (!focus)
+    focus = Fl::focus();
+
+  if (!focus)
+    return false;
+
+  return focus->window() == this;
+}
+
 void DesktopWindow::maybeGrabKeyboard()
 {
-  if (fullscreenSystemKeys && fullscreen_active())
+  if (fullscreenSystemKeys && fullscreen_active() && hasFocus())
     grabKeyboard();
 }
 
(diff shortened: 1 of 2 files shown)


samhed:
I have tested any combination of fullscreen and send system keys config I could think of. I have tested switching focus manually and automatically (`sleep 3; xterm`). Things seem to behave better now, I have tested on these platforms:

- [ ] macOS 11.6
- [ ] Fedora 34, Gnome on Wayland
- [ ] Fedora 34, Gnome on Xorg
- [ ] Windows 10

vncviewer/DesktopWindow.cxx
```diff
@@ -1014,6 +1004,14 @@ Bool eventIsFocusWithSerial(Display *display, XEvent *event, XPointer arg)
 }
 #endif
 
+void DesktopWindow::maybeGrabKeyboard(bool ungrab)
+{
+  if (fullscreenSystemKeys && fullscreen_active())
+    grabKeyboard();
+  else if (ungrab)
+    ungrabKeyboard();
```
CendioOssman, line 1026: I'm cautious about moving `ungrabKeyboard()` in here as well. It will now be called a lot more often than it used to be. It also is a bit surprising given the method's name.

Can't we have just `grabKeyboard()` here?

CendioOssman (changes requested):

samhed:
Updated according to your comment @CendioOssman

CendioOssman (approved):

PR #1347, opened 2021-10-01
System certificate load returns number loaded not GNUTLS_E_SUCCESS

twhitehead:
Had noticed this in the output despite the fact that the system certificates seemed to be working fine
```
$ vncviewer 
...
 TLS:         Could not load system certificate trust store
...
```

From the [documentation for `gnutls_certificate_set_x509_system_trust`](https://www.gnutls.org/manual/html_node/Certificate-credentials.html), seems the error check should have been `< 0` not `!= GNUTLS_E_SUCCESS`. To quote
> Returns: the number of certificates processed or a negative error code on error.

Haven't actually tested this, but it seems pretty trivial and obvious, so submitting it anyway.


hildred:
wouldn't it be <1 as an empty certificate store is useless?

PR #1349, opened 2021-10-05
Workaround FLTK bug that causes OptionsDialog to appear behind DesktopWindow

### LINKED ISSUE #1345: Options window sometimes appears behind vncviewer on macOS
**Describe the bug**
When running vncviewer in fullscreen on macOS and you try to open the options window from the F8 menu, it appears behind the vncviewer remote viewport. This makes it impossible to use, and additionally you can no longer control the remote session either.

Note that when in the state of the options being behind vncviewer, you can press ESC to close the hidden options window.

**To Reproduce**
1. Start vncviewer on macOS
2. Go to settings and enable fullscreen mode
3. Connect
4. Press F8 and open Options
5. See error

**Expected behavior**
The options window should appear in front.

**Screencast**
(recorded on phone since macOS screencast didn't work on fullscreen apps)
https://user-images.githubusercontent.com/4117162/134532987-bdd66ccd-0343-4cb1-83d3-ab72359a095a.mp4

 - OS: macOS 11.6
 - VNC client: TigerVNC
 - VNC client version: master
 - Client downloaded from: http://tigervnc.bphinz.com/nightly/

 - OS: Fedora 34
 - VNC server: TigerVNC
 - VNC server version: 1.11.0
 - Server downloaded from: Used the TigerVNC server included in ThinLinc
 - Server was started using: `Xvnc :50 -SecurityTypes=None`

**Additional context**
Similarly as #1332, this seems to be a regression from f4423b410c4f83b5679b959096f15bd631c2318a. Using a vncviewer built from that commit shows the problem, using the commit before works fine.

And also, same as #1332, leaving the Mac idle for a while seems to "fix" the issue. It always reappears after a logout-login though.

samhed:
Fixes #1345, details can be found in that issue. Regression, mainly after #642.

Tested using macOS 11.6 and built using FLTK 1.3.5, both with and without my suggested FLTK fix.

### DIFF
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -918,12 +918,11 @@ void DesktopWindow::fullscreen_on()
 {
   bool allMonitors = !strcasecmp(fullScreenMode, "all");
   bool selectedMonitors = !strcasecmp(fullScreenMode, "selected");
+  int top, bottom, left, right;
 
   if (not selectedMonitors and not allMonitors) {
-    int n = Fl::screen_num(x(), y(), w(), h());
-    fullscreen_screens(n, n, n, n);
+    top = bottom = left = right = Fl::screen_num(x(), y(), w(), h());
   } else {
-    int top, bottom, left, right;
     int top_y, bottom_y, left_x, right_x;
 
     int sx, sy, sw, sh;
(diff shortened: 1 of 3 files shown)


vncviewer/cocoa.mm
```diff
@@ -88,28 +98,25 @@ int cocoa_capture_displays(Fl_Window *win)
 
   captured = true;
 
-  if ([nsw level] == CGShieldingWindowLevel())
+  if (cocoa_get_level(win) == CGShieldingWindowLevel())
```
CendioOssman, line 101: I don't think these are useful. This file is supposed to be able to use Cocoa directly. These indirection functions are only for non-ObjectiveC code.

vncviewer/DesktopWindow.cxx
```diff
@@ -921,7 +921,16 @@ void DesktopWindow::fullscreen_on()
 
   if (not selectedMonitors and not allMonitors) {
     int n = Fl::screen_num(x(), y(), w(), h());
+#ifdef __APPLE__
+    // This is a workaround for a bug in FLTK, see: https://github.com/fltk/fltk/pull/277
+    int savedLevel;
+    savedLevel = cocoa_get_level(this);
+#endif
     fullscreen_screens(n, n, n, n);
```
CendioOssman, line 929: Can't we pop this out one level to avoid the duplication? It's easy to miss this in future changes.

CendioOssman:
I'd also merge the two commits given that the new cocoa helper functions aren't really meaningful until the second commit. :)

vncviewer/DesktopWindow.cxx
samhed: Sure, good idea :+1:

PR #1357, opened 2021-10-15
update Chinese (simplified) language

wwj402:
only update Chinese (simplified) language zh_CN.po.


CendioOssman:
Thank you for your contribution. Unfortunately we cannot accept translations directly as our translations are handled by the translation project. Please submit your improvements to their team here and it should reach us the right way:

https://translationproject.org/team/zh_CN.html

PR #1372, opened 2021-11-11
SELinux: restore SELinux context in case of different policies

grulja:
Try to restore SELinux context when it doesn't match current SELinux policy, which can happen when upgrading from Tigervnc 1.11.0 to any newer  version.

I have some questions:
1) When testing this, I got following error: `setxattr failed: /home/jgrulich/.vnc/: Operation not permitted`. I don't know if it's harmless or whether it can be avoided, but it successfuly changed the context.
2) Is checking whether the SELinux context `has vnc_home_t` enough? Or should I rather check whether it's equal to `unconfined_u:object_r:vnc_home_t:s0`?

CC: @zpytela

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -268,6 +268,19 @@ if(UNIX AND NOT APPLE)
   endif()
 endif()
 
+# Check for SELinux library
+if(UNIX AND NOT APPLE)
+  check_include_files(selinux/selinux.h HAVE_SELINUX_H)
+  if(HAVE_SELINUX_H)
+    set(CMAKE_REQUIRED_LIBRARIES -lselinux)
+    set(CMAKE_REQUIRED_LIBRARIES)
+    set(SELINUX_LIBS selinux)
+    add_definitions("-DHAVE_SELINUX")
+  else()
+    message(WARNING "Could not find SELinux development files")
+  endif()
+endif()
+
 # Generate config.h and make sure the source finds it
 configure_file(config.h.in config.h)
 add_definitions(-DHAVE_CONFIG_H)
--- a/unix/vncserver/vncsession.c
+++ b/unix/vncserver/vncsession.c
@@ -37,6 +37,11 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 
+#ifdef HAVE_SELINUX
+#include <selinux/selinux.h>
+#include <selinux/restorecon.h>
+#endif
+
 extern char **environ;
 
 // PAM service name
@@ -360,6 +365,17 @@ redir_stdio(const char *homedir, const char *display)
             syslog(LOG_CRIT, "Failure creating \"%s\": %s", logfile, strerror(errno));
             _exit(EX_OSERR);
         }
+
+#ifdef HAVE_SELINUX
+        int result;
+        if (selinux_file_context_verify(logfile, 0) == 0) {
+            result = selinux_restorecon(logfile, SELINUX_RESTORECON_RECURSE);
+
+            if (result < 0) {
+                syslog(LOG_WARNING, "Failure restoring SELinux context for \"%s\": %s", logfile, strerror(errno));
+            }
+        }
+#endif
     }
 
     hostlen = sysconf(_SC_HOST_NAME_MAX);
--- a/unix/vncserver/CMakeLists.txt
+++ b/unix/vncserver/CMakeLists.txt
@@ -1,5 +1,5 @@
 add_executable(vncsession vncsession.c)
-target_link_libraries(vncsession ${PAM_LIBS})
+target_link_libraries(vncsession ${PAM_LIBS} ${SELINUX_LIBS})
 
 configure_file(vncserver@.service.in vncserver@.service @ONLY)
 configure_file(vncsession-start.in vncsession-start @ONLY)


CMakeLists.txt
```diff
@@ -268,6 +268,18 @@ if(UNIX AND NOT APPLE)
...
+# Check for SELinux library
+if(UNIX AND NOT APPLE)
+  check_include_files(selinux/selinux.h HAVE_SELINUX_H)
+  set(CMAKE_REQUIRED_LIBRARIES -lselinux)
+  set(CMAKE_REQUIRED_LIBRARIES)
+  if(HAVE_SELINUX_H)
+    set(SELINUX_LIBS selinux)
+  else()
+    message(FATAL_ERROR "Could not find SELinux development files")
+  endif()
```
CendioOssman, line 280: Is this a reasonable hard requirement? Most distributions are not using SELinux.

unix/vncserver/vncsession.c
```diff
@@ -360,6 +364,25 @@ redir_stdio(const char *homedir, const char *display)
...
         }
+
+        result = getfilecon(logfile, &context);
+
+        if (result < 0) {
+            syslog(LOG_CRIT, "Failure obtaining SELinux context for \"%s\"", logfile);
+            _exit(EX_OSERR);
+        }
+
+        if (strstr(context, "vnc_home_t") == NULL) {
```
CendioOssman, line 375: It's always dangerous to hard code a redundancy like this. Can we instead as selinux if the file is "wrong"?

Or if we hard code something, wouldn't it be safer to hard code the known wrong ones?

CendioOssman, line 375: What about NFS?

CendioOssman (changes requested):

CendioOssman:
Since we are waiting for some SELinux expert to weigh in here, I'll mark this as "draft" for now.

unix/vncserver/vncsession.c
grulja: Changed to:
` if (strstr(context, "xdm_home_t") || strstr(context, "user_home_t"))`

I think those should be the wrong ones, but I'll wait for @zpytela.

zpytela: I'd also use functions from libselinux, equivalents of restorecon binary. Entries in `unix/vncserver/selinux/vncsession.fc` read:
```
HOME_DIR/\.vnc(/.*)?      gen_context(system_u:object_r:vnc_home_t,s0)
HOME_ROOT/\.vnc(/.*)?      gen_context(system_u:object_r:vnc_home_t,s0)
```
so these are the correct ones for `~/.vnc/`.

zpytela:
> Try to restore SELinux context when it doesn't match current SELinux policy, which can happen when upgrading from Tigervnc 1.11.0 to any newer version.
> 
> I have some questions:
> 
>     1. When testing this, I got following error: `setxattr failed: /home/jgrulich/.vnc/: Operation not permitted`. I don't know if it's harmless or whether it can be avoided, but it successfuly changed the context.

Do you have more details? Were there some related entries in the audit log at that time? Are the directories on a local filesystem?

>     2. Is checking whether the SELinux context `has vnc_home_t` enough? Or should I rather check whether it's equal to `unconfined_u:object_r:vnc_home_t:s0`?

It is not that important, for type enforcement rules only the type-part is. The restorecon command does not change it by default, it needs the `-F` switch to change the user and role part.

unix/vncserver/vncsession.c
grulja: @vmojzis I found selinux_file_context_verify() function. Can this be used instead of getting the context of that folder and comparing strings?

vmojzis: > @vmojzis I found selinux_file_context_verify() function. Can this be used instead of getting the context of that folder and comparing strings?

Yes, selinux_file_context_verify should be ideal for this situation (it makes sure given file has appropriate context based on the loaded policy). If you still wanted to verify that the context is  "vnc_home_t" you could use getfilecon (gives you the context assigned to given file on disk) or selabel_lookup (gives you the context suggested by the system policy). But I would go with selinux_file_context_verify to avoid hard-coding anything.

grulja:
This solution works for me. I tested following scenario:
1) Have Fedora 35 with Tigervnc 1.11.0 and with a user who already used Tigervnc so the context of `$HOME/.vnc` was `xdm_home_t`.
2) Update to Tigervnc 1.12.0 where the new context is `vnc_home_t` and run `vncsession`.
3) Check that `$HOME/.vnc` has now `vnc_home_t` context and VNC session is correctly started and can be connected to.
4) Downgrade again to Tigervnc 1.11.0, now the context of `$HOME/.vnc` is `unlabeled_t`, probably because `vnc_home_t` is not defined. 
5) Run again our modified vncsession, you will again get the context updated to `xdm_home_t`.

unix/vncserver/vncsession.c
```diff
@@ -360,6 +365,18 @@ redir_stdio(const char *homedir, const char *display)
...
             _exit(EX_OSERR);
         }
+
+#ifdef HAVE_SELINUX
+        int result;
+        if (selinux_file_context_verify(logfile, 0) == 0) {
+            result = selinux_restorecon(logfile, SELINUX_RESTORECON_RECURSE);
+
+            if (result < 0) {
+                syslog(LOG_CRIT, "Failure restoring SELinux context for \"%s\"", logfile);
```
vmojzis, line 375: It might be useful to also log the error reported by restorecon `strerror(errno)`.

vmojzis:
Sorry for the late reply. I could have sworn I replied to the question about which library function to use (selinux_file_context_verify is the best choice, if you really wanted to make sure the right label is applied you could use getfilecon), but the reply is nowhere to be found.
Anyway, the patch looks good to me. You could add some more reporting as to why the functions failed in case they do (selinux_file_context_verify can return 0 in multiple scenarios), but that is just nice to have for future debugging and I'm fine with the code the way it is now.

unix/vncserver/vncsession.c
grulja: Updated. Thanks.

CendioOssman (changes requested):
Looks good to me, except I didn't see anything explicit about NFS. Did you test that? I.e. when labels are forced to `nfs_t`.

grulja:
> Looks good to me, except I didn't see anything explicit about NFS. Did you test that? I.e. when labels are forced to `nfs_t`.

Do we need a special handling for `nfs_t`? Won't this be automatically handled by SElinux internally?

@zpytela @vmojzis do you know?

zpytela:
> > Looks good to me, except I didn't see anything explicit about NFS. Did you test that? I.e. when labels are forced to `nfs_t`.
> 
> Do we need a special handling for `nfs_t`? Won't this be automatically handled by SElinux internally?
> 
> @zpytela @vmojzis do you know?

NFS 4.2 can export a directory with native labels, but when it's not, some support should be in the policy so that vnc can read/manage files on a volume with `nfs_t`. For other services such rules are allowed by turning on a boolean which is off by default, see example

https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/contrib/mta.te#L112

Similar for smb or fusefs if this is supported by vnc.

grulja:
> > > Looks good to me, except I didn't see anything explicit about NFS. Did you test that? I.e. when labels are forced to `nfs_t`.
> > 
> > 
> > Do we need a special handling for `nfs_t`? Won't this be automatically handled by SElinux internally?
> > @zpytela @vmojzis do you know?
> 
> NFS 4.2 can export a directory with native labels, but when it's not, some support should be in the policy so that vnc can read/manage files on a volume with `nfs_t`. For other services such rules are allowed by turning on a boolean which is off by default, see example
> 
> https://github.com/fedora-selinux/selinux-policy/blob/rawhide/policy/modules/contrib/mta.te#L112
> 
> Similar for smb or fusefs if this is supported by vnc.

Isn't this what you implemented with https://github.com/TigerVNC/tigervnc/commit/3c8622691abfb377b48bf3749dd629c5a7120cf4? It's still not clear to me whether a special check needs to be done in this PR for `nfs_t`, or maybe I misunderstood your answer. I would expect that when I check a directory against system policy, it would check correct policy even for NFS mounted directory.

zpytela:
> Isn't this what you implemented with [3c86226](https://github.com/TigerVNC/tigervnc/commit/3c8622691abfb377b48bf3749dd629c5a7120cf4)? It's still not clear to me whether a special check needs to be done in this PR for `nfs_t`, or maybe I misunderstood your answer. I would expect that when I check a directory against system policy, it would check correct policy even for NFS mounted directory.

Right. The boolean set on just adds a few rules, default context does not change, note correct types can be even on NFS mounted files.

@vmojzis, is there a good practice for selinux_restorecon() usage when the path is e. g. on NFS, i. e. with a context different to default one based on path, but still correct? In this case, the library call would end with `_exit(EX_OSERR)`.

grulja:
> Looks good to me, except I didn't see anything explicit about NFS. Did you test that? I.e. when labels are forced to `nfs_t`.

Does the response from @vmojzis answer your question?

CendioOssman:
Yes, but the answer seems to be that TigerVNC won't start with NFS, which isn't really acceptable. :)

grulja:
> Yes, but the answer seems to be that TigerVNC won't start with NFS, which isn't really acceptable. :)

Yes, but I don't know how to fix it in this case. It is broken already since we changed the SELinux policy and this patch makes it half-broken for non-NFS users.

zpytela:
> > Yes, but the answer seems to be that TigerVNC won't start with NFS, which isn't really acceptable. :)
> 
> Yes, but I don't know how to fix it in this case. It is broken already since we changed the SELinux policy and this patch makes it half-broken for non-NFS users.

Unfortunately, restorecon or selinux_file_context_verify() are unable to discern why the label does not match so you can e. g. change the error to warning, or somehow detect if the file is on cifs or nfs with nfs_t labels.

CendioOssman:
A warning is probably safer. We don't know if the wrong label will prevent things from starting, just that it is likely.

Issue #102, opened 2015-01-17
trapping of all input in window mode


specing:
I am connecting to a remote running the same tiling window manager that I am running on my machine and while keyboard control keys work fine in fullscreen, I would like to have an option for tigerVNC to trap all keys even when in window mode.

Something like what Qemu SDL window does, upon clicking on it with mouse it traps all input except a configurable escape code (e.g. right CTRL).

CendioOssman:
Overlaps with issue #2.

specing:
Hah, anniversary edition.

hackonteur:
I looked into this a little bit and got frustrated. I only have a build environment working in Linux at the moment, but I got the keyboard grabbing to working in windowed mode there. The problem was the ungrabbing. I think that I ran into some fundamental issues with FLTK, at least in X.
If I move the mouse from inside the window to the titlebar, I see FL_LEAVE. Then if I push down the mouse button on the titlebar and drag it around, when I release the mouse button I get an FL_FOCUS and FL_ENTER. If I continue moving the mouse above the titlebar, I never get an FL_LEAVE event, so there is no way to ungrab the keyboard in that case.

CendioOssman:
Does indeed sound like a FLTK bug. You could try asking on their mailing list if you don't want to dig around in FLTK yourself.

hackonteur:
For those interested in this feature, it is worth noting that TurboVNC already supports it (although it is not documented): https://github.com/TurboVNC/turbovnc/issues/26

Also remember that you can use the TurboVNC viewer with the TigerVNC server.

JoseTomasTocino:
Has there been any progress in this issue?

CendioOssman:
Not much unfortunately. I've started playing around with a different approach to the hotkeys in order to make something like this easier to handle.

djuarezg:
still no update?

CendioOssman:
Afraid not. Trying to focus on things like #838 right now.

vermgit:
It's 2021 now, so 6.5 years have passed by. Any chance we can have this please?

I mean how am I supposed to switch windows on the remote machine? Clicking the app in the task bar is cumbersome as I'm a keyboard person.

danielkrajnik:
It would be really helpful if the TigerVNC viewer could caputre keyboard like remmina does. Working without it is so much harder.

nesqi:
It would be great if this could be implemented in some way for two reasons.

[...]

Issue #1070, opened 2020-08-01
Shortcut (like F8) to minimize the vnc


Naff86:
I currently use tigervnc for work using full screen mode over multiple monitors. To minimize the screen you press F8 and then clicking on minimize. 20% of the time I click on the toggle full screen button that is right next to the minimize on the F8 menu. This jumbles all the vnc window locations and is a pain to reconfigure. 

I suggest adding a shortcut to minimize ghe tiger vnc screen. Or alternatively to move the toggle full screen button somewhere lower on the menu. 

Thanks!

samhed:
You should be able to use the shortcut characters in the menu. When you open the F8 menu, look for the underlined characters.

For example, minimizing should probably be something like F8->m

Naff86:
Works! Thanks! 
F8->z

PR #1375, opened 2021-11-16
Use generic hot key combo for client control

### LINKED ISSUE #2: fast access (hotkey) to grab_keyboard()
From the mail to tigervnc-devel@lists.sourceforge.net from the 02.11.2013

Dear Tigervnc-developers,

I would like to know if it would be possible to integrate the functionality
provided by the attached patch (I know it is a hack, so it would need to be
done correctly) to a future stock tigervnc-release.

It assigns the grab_keyboard() function directly on the key "CONTROL_R" or
"META_L" for fast access (as also implemented in remmina).
This function is so much needed for me and my colleagues that I can not
imagine that it is not useful for others.
If you use a window manager like awesome then you need it....

Thanks a lot,

Michael

As I don't know how to attach a file to the comment, I paste the patch as an example I did here:

``` diff
--- tigervnc-1.3.0/vncviewer/DesktopWindow.cxx  2013-07-01 14:42:06.000000000 +0200
+++ ../tigervnc-1.3.0/vncviewer/DesktopWindow.cxx   2013-11-02 04:51:44.447096532 +0100
@@ -59,6 +59,9 @@ DesktopWindow::DesktopWindow(int w, int
   : Fl_Window(w, h), cc(cc_), firstUpdate(true),
     delayedFullscreen(false), delayedDesktopSize(false)
 {
+  
+  grab_keyboard_state = 0;
+
   scroll = new Fl_Scroll(0, 0, w, h);
   scroll->color(FL_BLACK);

@@ -449,8 +452,10 @@ int DesktopWindow::fltkHandle(int event,
       //        a) Fl::grab(0) on X11 will release the keyboard grab for us.
       //        b) Gaining focus on the system level causes FLTK to switch
       //           window level on OS X.
-      if (dw->fullscreen_active())
+      if (dw->fullscreen_active()) {
         dw->grabKeyboard();
+   dw->grab_keyboard_state = 1;
+      }
       break;

     case FL_UNFOCUS:
@@ -458,6 +463,7 @@ int DesktopWindow::fltkHandle(int event,
       //        focus as it is very tied to this specific window on some
       //        platforms and we want to be able to open subwindows.
       dw->ungrabKeyboard();
+      dw->grab_keyboard_state = 0;
       break;
     }
   }
--- tigervnc-1.3.0/vncviewer/DesktopWindow.h    2013-07-01 14:42:05.000000000 +0200
+++ ../tigervnc-1.3.0/vncviewer/DesktopWindow.h 2013-11-02 04:53:30.975097085 +0100
@@ -81,12 +81,15 @@ public:

   void fullscreen_on();

-private:
-  static int fltkHandle(int event, Fl_Window *win);
+  int grab_keyboard_state;

   void grabKeyboard();
   void ungrabKeyboard();

+private:
+  static int fltkHandle(int event, Fl_Window *win);
+
+
   static void handleGrab(void *data);

   void maximizeWindow();
@@ -113,6 +116,9 @@ private:
   bool firstUpdate;
   bool delayedFullscreen;
   bool delayedDesktopSize;
+
+
+
 };

 #endif
--- tigervnc-1.3.0/vncviewer/Viewport.cxx   2013-07-01 14:42:05.000000000 +0200
+++ ../tigervnc-1.3.0/vncviewer/Viewport.cxx    2013-11-02 04:56:05.376802328 +0100
@@ -356,6 +356,8 @@ int Viewport::handle(int event)
   int buttonMask, wheelMask;
   DownMap::const_iterator iter;

+  DesktopWindow *dw = dynamic_cast<DesktopWindow*>(window());
+
   switch (event) {
   case FL_PASTE:
     buffer = new char[Fl::event_length() + 1];
@@ -434,9 +436,27 @@ int Viewport::handle(int event)
     while (!downKeySym.empty())
       handleKeyEvent(downKeySym.begin()->first, downKeySym.begin()->first,
                      "", false);
+    dw->ungrabKeyboard();
+    dw->fullscreen_off();
+
     return 1;

   case FL_KEYDOWN:
+    // Code to use Control_R as a grabKeyboard shortcut, as in remmina
+    if (menuKeyCode && (Fl::event_key() == (hostKeyMetaL ? FL_Meta_L : FL_Control_R)  ) ) {
+      if(dw->grab_keyboard_state == 0) {
+   dw->grab_keyboard_state = 1;
+   dw->grabKeyboard();
+      } 
+      else  {
+   dw->grab_keyboard_state = 0;
+   dw->fullscreen_off();
+   dw->ungrabKeyboard();
+      }
+      //window()->fullscreen_off();
+      vlog.debug("keyboard_grab key (ctrl-right) hit, keyboard grab state: %d\n", dw->grab_keyboard_state);
+    }
+    
     if (menuKeyCode && (Fl::event_key() == menuKeyCode)) {
       popupContextMenu();
       return 1;
--- tigervnc-1.3.0/vncviewer/parameters.cxx 2013-07-01 16:00:00.000000000 +0200
+++ ../tigervnc-1.3.0/vncviewer/parameters.cxx  2013-11-02 04:57:22.692802730 +0100
@@ -127,6 +127,10 @@ BoolParameter sendPrimary("SendPrimary",
                           "Send the primary selection and cut buffer to the "
                           "server as well as the clipboard selection",
                           true);
+BoolParameter hostKeyMetaL("HostKeyMetaL",
+              "Use the MetaL (Windows-Key) instead of "
+              "Control_R as the host key (grabKeyboard shortcut)",
+              false);

 StringParameter menuKey("MenuKey", "The key which brings up the popup menu",
                         "F8");
@@ -171,6 +175,7 @@ VoidParameter* parameterArray[] = {
   &acceptClipboard,
   &sendClipboard,
   &sendPrimary,
+  &hostKeyMetaL,
   &menuKey,
   &fullscreenSystemKeys
 };
--- tigervnc-1.3.0/vncviewer/parameters.h   2013-07-01 14:42:06.000000000 +0200
+++ ../tigervnc-1.3.0/vncviewer/parameters.h    2013-11-02 04:57:48.371802864 +0100
@@ -58,6 +58,8 @@ extern rfb::BoolParameter acceptClipboar
 extern rfb::BoolParameter sendClipboard;
 extern rfb::BoolParameter sendPrimary;

+extern rfb::BoolParameter hostKeyMetaL;
+
 extern rfb::StringParameter menuKey;

 extern rfb::BoolParameter fullscreenSystemKeys;
```
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/2925617-fast-access-hotkey-to-grab_keyboard?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github).
</bountysource-plugin>

CendioOssman:
We want to be able to define more keyboard combinations that allows the user to control the client instead of sending the keys to the server. Instead of adding a never ending stream of random keys that just confuses things, we'll define a set of modifiers that are the base for all client control commands.

As a start, this adds hot keys for the menu (replacing menu key), grabbing and ungrabbing the keyboard, and toggling fullscreen.

Fixes #2 and partially #102. We cannot grab the keyboard on macOS in windowed mode with the current mechanism unfortunately.


CendioOssman:
Annoying. GNOME has a bug that prevents moving the window whilst the keyboard is grabbed:

https://gitlab.gnome.org/GNOME/mutter/-/issues/2021

Unfortunately it's old enough that MATE also has it:

https://github.com/mate-desktop/marco/issues/704

I don't see a way around it, so those users will have to live with having to ungrab to move the window. :/

CendioOssman:
My first plan was to keep the menu and have it pop up using <Modifiers>+M. However I see that at least PCoIP uses that to minimize the window:

https://www.teradici.com/web-help/TER1706002/3.3/Content/Topics/Using/Changing_window_mode.htm

We'd like people to keep their experiences from other software, and a menu is more uncommon, so perhaps it should have some other hot key.

CendioOssman:
A major issue here is how to deal with non-Latin keyboards. E.g. Cyrillic or Greek keyboards. I've had a look what other system do:

### GTK on Linux

Checks all other XKB groups for matches. So it basically expects a Latin alternative to be configured at all times. Also not that this is on the XKB level, where only four alternatives can be present at all time, not the much larger list that might be configured in GNOME.

If that isn't the case then no keyboard shortcuts work in at least GNOME's applications. It doesn't matter what the display language is, shortcuts are always in Latin.

### macOS

Refuses to let you remove the last Latin keyboard from the configuration, meaning one Latin keyboard layout is always present. Keyboard shortcuts are always in Latin and use that layout during lookup (somehow).

### Windows

Gladly lets you remove all Latin languages and keyboard layouts. Shortcuts are still in Latin though and still work. Unknown exactly what algorithm it uses at this point.

michaeltraxler:
I'm so happy that you put in the work to add this feature!
Thanks a lot! 
Would it be possible to also add the "right-ctrl" key to the list of hot-keys?
So, to make them to two keys, what they are, the left and right ctrl-key.

CendioOssman:
This is already fairly complex, so that is not in the plans, no. :/

Is there some scenario where you think you'll still have conflicts with other hotkeys?

CendioOssman:
> ### macOS
> 
> Refuses to let you remove the last Latin keyboard from the configuration, meaning one Latin keyboard layout is always present. Keyboard shortcuts are always in Latin and use that layout during lookup (somehow).

The magic seems to be that Command+\<key\> always produces a latin character. This makes sense given that Command is pretty much always used as a modifier for shortcuts on macOS.

This is a bit obscured during normal use of VNC as we only respect Shift, CapsLock and Alt (really AltGr) when generating keys for the server.

Also note #1396 which currently prevents non-Latin keyboard layouts anyway.

michaeltraxler:
> Is there some scenario where you think you'll still have conflicts with other hotkeys?

Yes, so many key-combinations are already used. ctrl-alt is "quite" common, so used by by some applications. With Alt+Win I can live as this is not used up to now by any known application. Not so easy to press, but humans are adaptable. Even better would be also to support "AltGr", instead of only "Alt". The same argument as for left and right "ctrl". The more options, the better, as everyone has a different keyboard and wants to adapt the hotkey to the specific keyboard he/she is using. But still I would prefer to use the ctrl_right key, as remmina is using for the grab key. But sure, it might be that others really use this ctrl_right key for their normal work :-)  But it is really great that you give so many options to the user already with the current version! A great way to go!

CendioOssman:
> ### Windows
> 
> Gladly lets you remove all Latin languages and keyboard layouts. Shortcuts are still in Latin though and still work. Unknown exactly what algorithm it uses at this point.

My guess is that applications focus on virtual-key codes. If I use `MapVirtualKey()` to map a scan code to a virtual key, then that virtual key to a character, then I always get a Latin character, even if the keyboard layout is Greek or a Cyrillic one.

The behavior seem to match what I'm seeing from other applications, so I'll go with that.

CendioOssman:
> My first plan was to keep the menu and have it pop up using <Modifiers>+M. However I see that at least PCoIP uses that to minimize the window:
> 
> https://www.teradici.com/web-help/TER1706002/3.3/Content/Topics/Using/Changing_window_mode.htm
> 
> We'd like people to keep their experiences from other software, and a menu is more uncommon, so perhaps it should have some other hot key.

Counter point: Most, if not all, systems have a shortcut for this already. E.g. Win+H in GNOME, Win+Down in Windows, and Cmd+M in macOS. As such I don't think we need to bother with duplicating functionality.

We added minimize to the context menu to solve #1070. However in that case the scenario was full screen, and likely keyboard grab was also active. That meant the system shortcuts didn't work. With this PR there will be a keyboard shortcut to release the keyboard, and then you can use the system shortcut. E.g. Ctrl+Alt followed by Win+Down on Windows.

We could still move the menu key to something else to avoid confusion, but I don't think we should add a shortcut for minimize.

rudironsonijr:
@CendioOssman, this refactoring makes a lot of sense. Do you intend to move it forward?
I can help!

CendioOssman:
At some point. But at the moment there are other projects ahead of this on my todo.

Neustradamus:
To follow this PR.

CendioOssman:
This should be feature complete now. The only point I'm debating is the shortcut to grab the keyboard. "Grab" is a rather old and technical term that isn't too friendly for new users. So "G" might not be the best choice. But I don't have any good alternatives at the moment.

CendioOssman:
> Annoying. GNOME has a bug that prevents moving the window whilst the keyboard is grabbed:

This got fixed in GNOME 44!

CendioOssman:
> Since F8 has been the default for a long time, what are your thoughts on showing a popup message when someone presses F8? It could hint about the changed shortcut for opening the context menu.

The transition could be an issue, yes. But I'm sceptical about adding such a thing. I'm worried about annoying users needlessly. Extra so if they've previously configured the menu on a different key.

The toast when connecting has been changed to show the new shortcut for opening the menu. Hopefully that is enough to inform users.

PR #1377, opened 2021-11-18
Update Chinese (simplified) translation

UOS-liang:


CendioOssman:
Thank you for your contribution. Unfortunately we cannot accept translations directly as our translations are handled by the translation project. Please submit your improvements to their team here and it should reach us the right way:

https://translationproject.org/team/zh_CN.html

PR #1380, opened 2021-11-23
Deduplicate vncviewer error messages

williamsjoblom:
vncviewer previously had a the same localized error message duplicated
on a bunch of places. Pull these duplicates out into a single function.


samhed (approved):
Looks good

PR #1382, opened 2021-11-24
Be consistent in "full screen" and "full-screen" use

williamsjoblom:
Previously, there were a number of inconsistencies in the use of
"full-screen" and "full screen" in logs, documentation, and user
interfaces. The consensus seems to be that "full screen" is correct when
used as a noun while "full-screen" is correct when used as an
adjective. These inconsistencies have now been adjusted to follow
consensus.

### DIFF
--- a/java/com/tigervnc/vncviewer/Parameters.java
+++ b/java/com/tigervnc/vncviewer/Parameters.java
@@ -117,7 +117,7 @@ public class Parameters {
 
   public static BoolParameter fullScreen
   = new BoolParameter("FullScreen",
-    "Full Screen Mode",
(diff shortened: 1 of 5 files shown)


williamsjoblom:
Here are a few sources to back my consensus claims:
https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/f/full-screen
https://english.stackexchange.com/questions/162421/fullscreen-or-full-screen
https://ell.stackexchange.com/questions/157952/full-screen-full-screen-or-fullscreen

samhed (approved):
Looks good!

PR #1383, opened 2021-11-24
Build fixes to libvnc.so for Xorg-server-21.1.1

joakim-tjernlund:
This makes libvnc Xorg module against Xorg-server-21.1.1.
Note that now one must enable present support in Xorg to make
libvnc build.

Only build tested though.

### DIFF
--- a/unix/xserver/hw/vnc/xorg-version.h
+++ b/unix/xserver/hw/vnc/xorg-version.h
@@ -33,8 +33,8 @@
 #error "X.Org older than 1.16 is not supported"
 #endif
 
-#if XORG_AT_LEAST(1, 21, 0)
-#error "X.Org newer than 1.20 is not supported"
+#if XORG_AT_LEAST(1, 22, 0)
+#error "X.Org newer than 1.21 is not supported"
 #endif
 
 #endif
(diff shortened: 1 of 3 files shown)


CendioOssman (changes requested):
Thanks. Most of this looks fine. But I think it's safer to clone of the patch file, just like we've done previously. Avoids issues in case we need to do further tweaks in the future for 21.

PS. I thought 21 required Meson?

joakim-tjernlund:
> Thanks. Most of this looks fine. But I think it's safer to clone of the patch file, just like we've done previously. Avoids issues in case we need to do further tweaks in the future for 21.

Done.
> 
> PS. I thought 21 required Meson?

I think they just added meson but the old autotools system is still there.

joakim-tjernlund:
Just remembered, the patching miinitext.c look odd, is that really needed?
If not one could just build libvnc against an already installed Xorg and remove the patching competently.
Then tigervnc does not have to bother with meson later on either.

CendioOssman:
That's so that Xvnc is built correctly, so I'm afraid it is needed. We also use internal APIs, so we will always need to be built in the Xorg source tree.

joakim-tjernlund:
> That's so that Xvnc is built correctly, so I'm afraid it is needed. We also use internal APIs, so we will always need to be built in the Xorg source tree.

I see, does that extend to the libvnc.so Xorg module as well?

CendioOssman:
Yup. That's why it needs to be rebuilt when Xorg changes. I think the only somewhat stable API is the graphics driver API, and we need more than that.

PR #1402, opened 2021-12-27
Support Button2 emulation using Shift+ButtonOne

dl-g:
This works better on Macs with a one-button mouse or a track pad.

### DIFF
--- a/vncviewer/parameters.cxx
+++ b/vncviewer/parameters.cxx
@@ -185,9 +185,6 @@ static VoidParameter* parameterArray[] = {
 #endif // HAVE_GNUTLS
   &SecurityClient::secTypes,
   /* Misc. */
-  &emulateMiddleButton,
-  &emulateMiddleButton2,
-  &dotWhenNoCursor,
   &reconnectOnError,
   &shared,
   /* Compression */
@@ -206,6 +203,7 @@ static VoidParameter* parameterArray[] = {
   /* Input */
   &viewOnly,
   &emulateMiddleButton,
+  &emulateMiddleButton2,
   &dotWhenNoCursor,
   &acceptClipboard,
   &sendClipboard,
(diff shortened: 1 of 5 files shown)


vncviewer/parameters.cxx
```diff
@@ -181,6 +185,9 @@ static VoidParameter* parameterArray[] = {
 #endif // HAVE_GNUTLS
   &SecurityClient::secTypes,
   /* Misc. */
+  &emulateMiddleButton,
+  &emulateMiddleButton2,
+  &dotWhenNoCursor,
```
conqp, line 190: Why did you put this in here given the scope of this commit?
Not it's in the array twice.

dl-g: Thanks for the review.

This is a mistake from the last rebase conflict resolution. I have fixed it.  Thanks for catching it.

CendioOssman:
Please see #1192. To handle this properly we likely need more complexity than this simple check. E.g. with this code Shift will be kept down whilst the middle button is sent. And Shift+Middle might do different things in the application than just Middle.

dl-g:
Thank you for the review.  If I read the comments correctly, you raised two concerns,

1. LeftShift + LeftClick (or LeftShift + MiddleClick) might have an existing binding.  This is true for any other emulations, including Left+Right Click,   and LeftAlt+LeftClick.  This feature is only useful if this combination is currently unassigned.
That's why this feature should be a non-default option.  The user enables it if this does not conflict with his existing environment.   Ideally,  it would have been more flexible, if the user can choose a key (or another button) to be used with LeftClick for the emulation.

2. How much code would it require to implement this feature?  As proven by the L+R emulation,  it does not take much code change.  If you believe the L+R code is correct.

dl-g:
Here is the sequence of input events involved in the emulation:

Viewer                                               Server (no emulation)                       Server (emulation)
===================================================================
LeftShift down                                  LeftShift down                                   LeftShift down
LeftClick down                                  LeftClick down                                  MiddleClick down
LeftClick up                                       LeftClick up                                       MiddleClick up
LeftShift up                                       LeftShift up                                        LeftShift up

It's assumed that there is no existing binding for LeftShift + MiddleClick.

PR #1405, opened 2022-01-07
Update MonitorArrangement widget to match parameter settings after screen configuration change

samhed:
If the options dialog was open when a screen configuration happened the widget could get out of sync from the settings. A scenario when this happened was:

 1) 3 monitors, fullscreen selected on the two right-most screens
 2) disconnect the left-most screen (the one not selected)

In this case, using GNOME, vncviewer would appear in fullscreen on the right of the two remaining monitors, but the widget would show both monitors selected. The reason was that the MonitorArragement index doesn't work the same way as FLTK's screen index.

It's debatable how vncviewer should behave here, but the GUI should at least match the actual setting.

### DIFF
--- a/vncviewer/OptionsDialog.cxx
+++ b/vncviewer/OptionsDialog.cxx
@@ -20,8 +20,8 @@
 #include <config.h>
 #endif
 
+#include <assert.h>
 #include <stdlib.h>
-
 #include <list>
 
 #include <rdr/types.h>
@@ -40,6 +40,7 @@
 #include "parameters.h"
 #include "MonitorArrangement.h"
 
+#include <FL/Fl.H>
 #include <FL/Fl_Tabs.H>
 #include <FL/Fl_Button.H>
 #include <FL/Fl_Check_Button.H>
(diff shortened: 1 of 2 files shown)


samhed:
Tested on Fedora 35, macOS 12.1, and Windows 10.

CendioOssman:
Did you look at how this would race with the event handler in the monitor widget?

samhed:
I had considered it, but now that I look closer I see that we're not safe here. That event handler could cause problems if it fires after the handler in this PR since the map `monitors` isn't cleared in the new handler.

I see two options:
* making `refresh()` of the MonitorArrangement class public
* removing the handler from MonitorArrangement and making this stuff the responsibility of the implementor

CendioOssman:
* Use an idle handler/timeout to make sure this is run after every other handler is done

samhed:
Good point regarding the timeout. Updated the commit.

I had some trouble testing this, but looking at the code at https://github.com/fltk/fltk/blob/branch-1.3/src/Fl.cxx indicates that timeouts with `time` set to zero should work as expected.

I verified that things still work on Windows 11 and Fedora 35.

PR #1411, opened 2022-01-19
Fix session resize after mirroring on Linux vncviewer

williamsjoblom:
If monitor mirroring was enabled while in a session with vncviewer
running on Linux, the session would not be properly resized on the
server. This was a consequence of only looking at the size and
coordinates of each screen when matching against existing screens after
the screen layout was changed, when in fact we have two (or more)
monitors with the same coordinates and size (but differing ids). This
led to the same monitor being added twice to the layout which would
later fail layout validation, resulting in no resize request being sent
to the server.

When matching, we now also check if the existing screen is not already
present in the layout before considering it a match.

### DIFF
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -21,6 +21,8 @@
 #include <config.h>
 #endif
 
+#include <algorithm>
+
 #include <assert.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


williamsjoblom:
For d83cc9e, I also had a look around for other places where we might want unconditional logging of screen layouts that refuses to validate. Sadly, the other calls to `ScreenSet::validate` are either not assertive in nature or throw an `Exception`. We could theoretically throw a derived exception with the layout dump attached, but this would complicate exception handling quite a bit. So I'd suggest that we settle with the logging changes in d83cc9e for now.

PR #1412, opened 2022-01-19
Change Background color of FullScreen Window Borders to Black

mtatton:
Hello, I would like to propose a tiny patch. My motivation is having vnc client session screen of source computer smaller than actual screen size of local computer. In this case when I go fullscreen to prevent alt+f4 shortcut to work the border of the vnc session is grey. That's a bit disturbing for my work. This patch is a proposal to enter immersive mode by having the blank borders of smaller screen in full screen in black. Thank You

### DIFF
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -465,7 +465,7 @@ void DesktopWindow::draw()
   // Redraw background only on full redraws
   if (redraw) {
     if (offscreen)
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm afraid this was on purpose in order to see where the session borders are in case a session in smaller than the window. The dark gray was meant to be dark enough to not have a massive contrast against black, but still different enough that you could tell the difference.

PR #1426, opened 2022-02-07
SELinux: use /root/.vnc in file context specification

zpytela:
Instead of HOME_ROOT/.vnc, /root/.vnc should be used
for user root's home to specify default file context
as HOME_ROOT actually means base for home dirs (usually /home).

### DIFF
--- a/unix/vncserver/selinux/vncsession.fc
+++ b/unix/vncserver/selinux/vncsession.fc
@@ -18,7 +18,7 @@
 #
 
 HOME_DIR/\.vnc(/.*)?      gen_context(system_u:object_r:vnc_home_t,s0)
-HOME_ROOT/\.vnc(/.*)?      gen_context(system_u:object_r:vnc_home_t,s0)
+/root/\.vnc(/.*)?      gen_context(system_u:object_r:vnc_home_t,s0)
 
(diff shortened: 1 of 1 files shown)


grulja:
I wonder whether it shouldn't be `HOME_DIR` replaced by `/root` as from the documentation I see:
`HOME_ROOT - This keyword is replaced by the GNU / Linux users root home directory, normally '/home' is the default.`
which would mean `HOME_ROOT/\.vnc(/.*)?` expands to e.g.` /home/jgrulich/.vnc/.*`

BUT
`HOME_DIR - This keyword is replaced by the GNU / Linux users home directory, normally '/home/' is the default.`
which would mean `HOME_DIR/\.vnc(/.*)?` expands to e.g. `/home/jgrulich//.vnc/.*`

Point is one `/` vs two `//`.

grulja:
Btw. this change seems to work as it is.

Before:
```
[root@localhost ~]# matchpathcon /root/.vnc
/root/.vnc	system_u:object_r:admin_home_t:s0
```
After:
```
[root@localhost ~]# matchpathcon /root/.vnc/
/root/.vnc	system_u:object_r:xdm_home_t:s0
```
Ignore that it has `xdm_home_t` instead of `vnc_home_t`. I test this with Tigervnc 1.11.0.

PR #1451, opened 2022-04-09
Add view only button on context menu

norihiro:
This PR suggests to add a view-only item on the F8 menu.

When assisting someone's desktop through VNC, I frequently switch view-only. The menu item with `v` shortcut is useful to switch view-only.

Please feel free to reject if you don't want to add this feature.

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -104,6 +104,7 @@ static rfb::LogWriter vlog("Viewport");
 // Menu constants
 
 enum { ID_DISCONNECT, ID_FULLSCREEN, ID_MINIMIZE, ID_RESIZE,
+       ID_VIEWONLY,
        ID_CTRL, ID_ALT, ID_MENUKEY, ID_CTRLALTDEL,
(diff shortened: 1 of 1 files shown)


CendioOssman:
Your use case is reasonable. I'm concerned about clutter in the menu though as I don't think there is a one-size-fits all here. So with that in mind we try to be restrictive in adding things.

What could make things less restrictive is adding the ability for the user to choose what should be included. Then we would just be restrictive with the default settings, and users could enable more uncommon things as needed.

Do you think you could have a look at adding such a feature?

norihiro:
> I'm concerned about clutter in the menu though as I don't think there is a one-size-fits all here.

I agree that point.
Maybe I can add an option `-MenuViewOnly`

CendioOssman:
That unfortunately doesn't scale well. I was thinking more like `-MenuItems=minimize,viewonly`. That way we can easily have many items. Controlling this from the Options dialog would also be preferred.

CendioOssman:
Closing due to lack of activity.

PR #1452, opened 2022-04-09
Add button to send print screen key on context menu

norihiro:
This PR adds a menu item to send print screen key.
Since the print screen key is grabbed by the session manager at client side, the screenshot cannot be taken on the server side.


CendioOssman:
There are many keyboard combinations that are grabbed by the local system, not just PrintScreen. That's why we have the "Send system keys" feature in the viewer.

I assume this isn't sufficient in your scenario?

CendioOssman:
Ah. Then #102 is probably of interest to you.

PR #1464, opened 2022-05-03
Allow "PointerEventInterval" to be set in "~/.vnc/default.tigervnc".

ziyan:
```bash
# cat ~/.vnc/default.tigervnc
TigerVNC Configuration file Version 1.0
DotWhenNoCursor=1
PointerEventInterval=1
```

Fixes the issue where "PointerEventInterval" param is not recognized in the configuration:

```
 Parameters:  Failed to read line 3 in file
              /host/home/ziyan/.vnc/default.tigervnc: Unknown parameter
```

### DIFF
--- a/vncviewer/parameters.cxx
+++ b/vncviewer/parameters.cxx
@@ -207,7 +207,8 @@ static VoidParameter* parameterArray[] = {
   &setPrimary,
 #endif
   &menuKey,
-  &fullscreenSystemKeys
+  &fullscreenSystemKeys,
+  &pointerEventInterval,
 };
 
 static VoidParameter* readOnlyParameterArray[] = {


CendioOssman:
I'm afraid, we only save (and load) things that are accessible from the GUI. That is done, so users don't get confused by extra settings they cannot see.

We could consider loading extra settings, but not storing them. But that would be a different PR than this. And I'm not sure if that would solve your use case.

Why do you want to change this setting?

ziyan:
@CendioOssman Sorry for the late response. The main purpose of setting `PointerEventInterval` to something lower than the default is due to our usage of a 3D viewer over VNC connection on the local network. 3D viewer implemented using OpenSceneGraph relies on continuous mouse pointer events to calculate the camera rotation. The difference is very visible between the default 17ms and 1 or 0ms. At the 17ms interval, objects "feel" heavy to drag and spin. At a lower interval, it feels like the viewer is running locally. I do understand this is quite an advanced parameter and the user needs to understand the implication on network usage and how the VNC server will react.

CendioOssman:
That is unexpected. The default interval should be faster than your monitor, so the "feel" should be the same.

Do you think you could type up some instructions so we can test this ourselves? I'm hoping there is a bug somewhere that can be fixed.

PR #1478, opened 2022-06-04
QEMU Audio support in VNC Viewer (currently Windows only)

mkupchik:
This pull request implements a new feature in the VNC Viewer: support for VNC protocol extension called "QEMU audio". My testing indicates that it works quite well, including anti-jitter measures (such as audio output device starvation detection and adjusting audio playback delay accordingly). I've implemented QEMU audio support on the Windows platform only, because this is only platform I'm familiar with (and the only platform I use TigerVNC on). This is a byproduct of another project I currently work on, so I'm just sharing my work in hope that it will be eventually upstreamed and I won't have to rebase it to the subsequent versions of TigerVNC and recompile manually.


common/rfb/CConnection.cxx
```diff
@@ -49,7 +49,7 @@ static LogWriter vlog("CConnection");
 CConnection::CConnection()
   : csecurity(0),
     supportsLocalCursor(false), supportsCursorPosition(false),
-    supportsDesktopResize(false), supportsLEDState(false),
+    supportsDesktopResize(false), supportsLEDState(false), supportsQEMUAudio(false),
```
CendioOssman, line 52: We try to use generic phrasing for this, so future extensions can also be handled. So please rename this to something like `supportsAudio`.

common/rfb/CMsgWriter.cxx
```diff
@@ -204,6 +204,25 @@ void CMsgWriter::writeClientCutText(const char* str)
...
+  os->writeU8(qemuAudio);
+  os->writeU16(enable ? 0 : 1);
+  endMsg();
+}
+
+void CMsgWriter::writeQemuAudioSetFormat(rdr::U8 fmt, rdr::U8 channels, rdr::U32 frequency)
+{
+  startMsg(msgTypeQEMUClientMessage);
+  os->writeU8(qemuAudio);
+  os->writeU16(2);
```
CendioOssman, line 219: Please try to define constants for things such as this. Makes it easier to read this code.

common/rfb/CMsgReader.cxx
```diff
@@ -81,6 +82,23 @@ bool CMsgReader::readServerInit()
...
+      is->hasData(__rfbmin(maxBufferedAudioBytes, nAudioBytesLeft));  // request as much as possible
+      size_t available = __rfbmin(is->avail(), nAudioBytesLeft);      // see how many we've got
+      size_t consumed = available ? handler->audioAddSamples(is->getptr(available), available) : 0;
+      if (consumed == 0)
+        break;
+      is->skip(consumed);
+      nAudioBytesLeft -= consumed;
+    }
+    if (nAudioBytesLeft != 0)
+      return false;
```
CendioOssman, line 96: This complexity is something that doesn't really fit well  in `CMsgReader`. How large are these audio chunks? Can't we let the input stream handle the buffering?

vncviewer/CConn.cxx
```diff
@@ -373,6 +393,22 @@ void CConn::framebufferUpdateEnd()
...
+      writer()->writeQemuAudioSetFormat(win32AudioOutput->getSampleFormat(), 
+                                        win32AudioOutput->getNumberOfChannels(),
+                                        win32AudioOutput->getSamplingFreq());
+      writer()->writeQemuAudioEnableOrDisable(true /* enable */);
+    } else {
+      delete win32AudioOutput;
+      win32AudioOutput = NULL;
+    }
+  }
+#endif
```
CendioOssman, line 410: Can this be pushed to `CConnection` instead? Starting the audio on first detection of QEMU audio support should be generic, and ideally handled there.

vncviewer/Win32AudioOutput.cxx
```diff
@@ -0,0 +1,303 @@
...
+  : haveWO(false), openedWO(false), sampleFormat(sampleFormatU8), numberOfChannels(0),
+    samplingFreq(0), currentStreamId(0), handleWO(NULL), bufPtr(NULL), bufTotalSize(0),
+    bufFreeSize(0), bufUnsubmittedSize(0), bufSubmittedHead(0), bufUnsubmittedHead(0),
+    doneHdrsSlist(NULL), hdrsInFlight(0), extraDelayInMillisec(0)
+{
+  static const rdr::U32 freqTable[4] = {44100, 48000, 22050, 11025};
+
+  for (rdr::U8 bits_per_sample = 16; bits_per_sample != 0; bits_per_sample -= 8) {
+    for (rdr::U8 n_channels = 2; n_channels != 0; n_channels--) {
+      for (rdr::U8 freq_index = 0; freq_index < 4; freq_index++) {
```
CendioOssman, line 31: Doesn't Windows resample for us? Do we really need to try all of these? It would make things a lot simpler if we just ha a fixed format we always use.

CendioOssman (changes requested):
Thanks for your PR! This is definitely something we are interested in.

How is the easiest way for us to go about testing this?

common/rfb/CConnection.cxx
mkupchik: OK

common/rfb/CMsgWriter.cxx
mkupchik: OK

common/rfb/CMsgReader.cxx
mkupchik: Audio chunks may be large, there are no protocol limits besides UINT32_MAX. Bad-behaved VNC server may outgrow MAX_BUF_SIZE in common/rdr/BufferedInStream.cxx. That's why I prefer not to leave audio samples in the input buffer of BufferedInStream, and move them to circular playback buffer of Win32AudioOutput as soon as they arrive. Calls to memcpy happen in  Win32AudioOutput::addSamples(), usually once, or twice if wraparound in the target circular buffer happens, and caller of that interface just need to provide a pointer and length of continuous chunk of source data. This continuous chunk of source data is owned by BufferedInStream.

When audio frame is fully received, Win32AudioOutput::submitSamples() makes one or two heap allocations of struct Win32AudioOutput::HdrInSlist, then calls waveOutPrepareHeader() WinAPI function (which boils down to allocation of struct OVERLAPPED on heap, allocation of kernel event object and some pointer bookkeeping), then it calls waveOutWrite() WinAPI function (which boils down to submitting asyncronous I/O request to the NT kernel device driver, the audio mixer). There's no need to move bytes around in this context, asyncronous NT I/O request in flight refers to the audio samples in circular buffer owned by Win32AudioOutput which were added there earlier, by Win32AudioOutput::addSamples(). So while audio playback I/O request is still in flight, original socket buffer owned by BufferedInStream may contain different data, may be reallocated, may be empty etc.

When kernel completes asyncronous I/O request, it signals via kernel event object and awakes internal worker thread started by winmm.dll/wdmaud.drv. This thread deallocates struct OVERLAPPED on heap, disposes kernel event object and invokes our Win32AudioOutput::waveOutCallback(). There we link struct Win32AudioOutput::HdrInSlist into slist to dispose it later, during the next call to Win32AudioOutput::submitSamples() or in the destructor of Win32AudioOutput.

So, what we need from BufferedInStream in this context of CMsgReader is an API to fetch some data fresh from socket and provide raw pointer to the continuous chunk of memory along with number of available bytes. There's an upper limit on the number of bytes we need here (nAudioBytesLeft), and there's no lower limit. We're happy to consume less than nAudioBytesLeft bytes here (especially if nAudioBytesLeft is large, i.e. this is a large audio frame). If there's partial audio sample/frame (e.g. less than 4 bytes for default audio format, 16-bit stereo), then up to 3 bytes will be left unconsumed in the BufferedInStream, for later use when more data arrives.

vncviewer/CConn.cxx
mkupchik: Yes, I think it's possible. Testing of win32AudioOutput against NULL and invokation of win32AudioOutput->openAndAllocateBuffer() unless already opened should remain in CConn, probably in the new virtual method, and the rest of this logic (including calling methods of CMsgWriter) is generic and can be moved up to the base class.

vncviewer/Win32AudioOutput.cxx
CendioOssman: All of this adds a lot of complexity to the code. So if it isn't needed in practice, then it's just a cost.

Have you actually seen it failing to give you the requested format?

common/rfb/CMsgReader.cxx
mkupchik: I think this complexity can be reduced by adding a new method to rdr::InStream, which takes a number of bytes wanted by caller, similar to hasData(), but unlike hasData() does not require all of that bytes to be available. Instead it just returns a pointer and length of what is available. Length may be less than requested. Also this new method must call rdr::InStream::overrun() for actual I/O on socket to happen. Then code in CMsgReader will be shorter. What to do think?

mkupchik: Also, I think I need to provide justification for MSGSTATE_AUDIO_DATA. It's true that entire 1 MiB of audio samples can fit in socket I/O buffer and be submitted for playback in one go. However consider ill-behaved VNC server which sends 1 GiB of audio data, so 1 MiB of these data is going to be submitted for playback and the rest (1023 MiB) is going to be discarded. How we can model this logic in the terms of rdr::InStream: "keep N bytes at the beginning of the buffer, but discard everything above N?" I think we cannot do this, rdr::InStream is designed for FIFO-style consumption of incoming data, we can't cut away samples/bytes/whatever in the middle of that buffer. Hence the compexity in CMsgReader, the new state of the FSM, and the double buffering.

faust93:
Might be useful, there's a PoC with cross-platform audio support using miniaudio library based on original @mkupchik commit
https://github.com/faust93/tigervnc/commits/master (last 3 commits)
Haven't tested it with QEMU though but for me it works more or less ok under macOS and Linux with my TurboVNC fork patched for audio support
https://github.com/faust93/turbovnc/commit/092b233cdda6bf433d1911c800d1b488f8c933fd

CendioOssman:
This was finished up and resolved via #2138.

PR #1480, opened 2022-06-08
Workaround for MFT cropping in H264 decoding

mmozeiko:
It seems MFT h264 decoder does not support frame cropping in h264 bitstream. So in cases when frame width or height is not multiple of 16 then decoded output can potentially be used with wrong offset.

This code adds explicit parsing of SPS to extract cropping information to use, and will apply cropping if reported size differs from
expected.

Good example to test is with pikvm's vnc server on 1920x1080 resolution. rpi hardware h264 encoder encodes it as 1920x1088 resolution with offset from top and bottom set to 4. This means that without this patch there are 4 black rows on top, and 4 missing rows on bottom. See the following images:

original code:
bad

with this PR:
fixed


CendioOssman (approved):
Thanks for the fix!

PR #1489, opened 2022-07-04
Add missing FLTK include statements

Albrecht-S:
This PR is about adding include statements (and one other modification) to make TigerVNC compile with FLTK 1.4 (Git master).

Although FLTK 1.3 and 1.4 are API compatible, some missing headers have been unnecessarily included in other FLTK 1.3 headers such that not including these headers in source files worked by accident. I'm commenting the changes below in the form of pseudo diff statements:
```
+#include <FL/x.H>
```
This one is the header missing most often. It is needed for global FLTK variables and functions like fl_display and similar. Note that this header file has been renamed to <FL/platform.H> in FLTK 1.4.0 and a backwards compatibility header <FL/x.H> was kept (includes platform.H).

OTOH FLTK 1.3.5 and later got an additional header <FL/platform.H> so code can use either one if it targets FLTK 1.3.5 or later. The "safe" version for now is <FL/x.H> but the future proof version is <FL/platform.H> although FLTK 1.4 and later will retain <FL/x.H> for a while.

```
vncviewer/MonitorArrangement.h:
-class Fl_Group;
+#include <FL/Fl_Group.H>
```
It is not possible to derive a class from an incomplete type (forward declaration `class Fl_Group`). In FLTK 1.3 `<FL/Fl_Group.H>` must have been included elsewhere.

```
+#include <malloc.h>
```
Two occurrences show that `malloc.h` has been included elsewhere, supposedly in another FLTK 1.3 header, but not in FLTK 1.4.

Note that reducing of unnecessary FLTK and system header includes was a goal of changes in FLTK 1.4 for better compilation speed. Including these additional headers in the TigerVNC source files should not do any harm because all header files have the usual header guard statements `#ifndef ...`.

I'd appreciate if this PR would be accepted and merged because it makes testing TigerVNC with FLTK 1.4 much easier and this will be needed in the future anyway.

### DIFF
--- a/vncviewer/MonitorArrangement.h
+++ b/vncviewer/MonitorArrangement.h
@@ -29,7 +29,8 @@
 #include <map>
 #include <set>
 
-class Fl_Group;
+#include <FL/Fl_Group.H>
+
 class Fl_Button;
 
 class MonitorArrangement: public Fl_Group {
--- a/vncviewer/PlatformPixelBuffer.cxx
+++ b/vncviewer/PlatformPixelBuffer.cxx
@@ -21,6 +21,7 @@
 #endif
 
 #include <assert.h>
+#include <stdlib.h>
 
 #if !defined(WIN32) && !defined(__APPLE__)
 #include <sys/ipc.h>
--- a/vncviewer/Surface_X11.cxx
+++ b/vncviewer/Surface_X11.cxx
@@ -21,6 +21,7 @@
 #endif
 
 #include <assert.h>
+#include <stdlib.h>
 
 #include <FL/Fl_RGB_Image.H>
 #include <FL/x.H>
--- a/tests/perf/fbperf.cxx
+++ b/tests/perf/fbperf.cxx
@@ -26,6 +26,7 @@
 #include <FL/Fl.H>
 #include <FL/Fl_Window.H>
 #include <FL/fl_draw.H>
+#include <FL/x.H>
 
 #include <rdr/Exception.h>
 #include <rfb/util.h>
--- a/vncviewer/MonitorArrangement.cxx
+++ b/vncviewer/MonitorArrangement.cxx
@@ -37,6 +37,7 @@
 #include <FL/Fl.H>
 #include <FL/fl_draw.H>
 #include <FL/Fl_Button.H>
+#include <FL/x.H>
 
 #if defined(HAVE_XRANDR) && !defined(__APPLE__)
 #include <X11/extensions/Xrandr.h>
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -73,6 +73,7 @@
 
 #include <FL/Fl_Menu.H>
 #include <FL/Fl_Menu_Button.H>
+#include <FL/x.H>
 
 #if !defined(WIN32) && !defined(__APPLE__)
 #include <X11/XKBlib.h>


Albrecht-S:
This PR does not change any build files, e.g. `CMakeLists.txt`. I used CMake in my tests and had to make a small change for building TigerVNC with CMake on my Linux system. This change was intentionally not included because it may be platform specific, but here it is FYI anyway:
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index d88b177f..f0b973e9 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,6 +265,7 @@ if(BUILD_VIEWER)
     if(X11_Xrender_FOUND)
       set(FLTK_LIBRARIES ${FLTK_LIBRARIES} ${X11_Xrender_LIB})
     endif()
+    set(FLTK_LIBRARIES ${FLTK_LIBRARIES} png dl)
   endif()
 endif()
 
```

Note 1: this is not a fix, it's more like a hack.
Note 2: I ran the TigerVNC viewer (vncviewer) built this way with FLTK 1.4 (Git master) successfully but this does not guarantee any specific functionality. I also built and tested it only on Linux.

vncviewer/MonitorArrangement.h
```diff
@@ -29,7 +29,7 @@
 #include <map>
 #include <set>
 
-class Fl_Group;
+#include <FL/Fl_Group.H>
```
CendioOssman, line 32: A blank line would be nice here to separate includes from declarations.

vncviewer/PlatformPixelBuffer.cxx
```diff
@@ -34,6 +34,7 @@
 #include <rdr/Exception.h>
 
 #include "PlatformPixelBuffer.h"
+#include <stdlib.h>
```
CendioOssman, line 37: Please try to maintain the header sorting/grouping. In this case this should be up higher with `<assert.h>`.

vncviewer/Surface_X11.cxx
```diff
@@ -28,6 +28,7 @@
 #include <rdr/Exception.h>
 
 #include "Surface.h"
+#include <stdlib.h>
```
CendioOssman, line 31: Same here.

vncviewer/PlatformPixelBuffer.cxx
```diff
@@ -34,7 +34,7 @@
 #include <rdr/Exception.h>
 
 #include "PlatformPixelBuffer.h"
-#include <malloc.h>
+#include <stdlib.h>
```
CendioOssman, line 37: Please squash this in to the original commit.

CendioOssman (changes requested):
Thanks. Much appreciated!

How are things going for 1.4.0? Any beta out soon? :)

CendioOssman:
> This PR does not change any build files, e.g. `CMakeLists.txt`. I used CMake in my tests and had to make a small change for building TigerVNC with CMake on my Linux system. This change was intentionally not included because it may be platform specific, but here it is FYI anyway:

Are you building a static FLTK library here? If so, then that is always a bit magical. There is no decent way to figure out recursive dependencies in those cases.

We have the `BUILD_STATIC` CMake flag which tries to handle this. But it is very much best-effort and will break on many setups.

vncviewer/PlatformPixelBuffer.cxx
Albrecht-S: Will do.

vncviewer/Surface_X11.cxx
Albrecht-S: Done.

Albrecht-S:
Changed as requested, squashed, rebased, and force-pushed (07d4e29042409a0c2d4912d25af8fab71fa71f33).

CendioOssman (approved):
Great, thanks!

Albrecht-S:
Thanks for merging this PR.

PR #1491, opened 2022-07-06
Add vncviewer option to use large cursor instead of dot

### LINKED ISSUE #1483: Vncviewer - show mouse symbol rather than dot when no cursor .  [$100]
In vncviewer - there is an option for "show dot when no cursor " , though this is very helpful . 
its  unusable , while using VNC client , remote mouse lags and the dot which creates a perception of speed is pretty invisible . 

I think in vncviewer project : 

adding an option to show imaginary mouse pointer or an option to replace dot with icons similar to mouse cursor will be a good add . 

I know this may not be a valid request , but i plan to propose a bounty and bountysource expects a issue url :) 

https://app.bountysource.com/issues/109362423-vncviewer-show-mouse-symbol-rather-than-dot-when-no-cursor

Bounty of 100 USD - i am ok , even if i can get a working local copy . 

<bountysource-plugin>

---
There is a **[$100 open bounty](https://app.bountysource.com/issues/109362423-vncviewer-show-mouse-symbol-rather-than-dot-when-no-cursor?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github)** on this issue. Add to the bounty at [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F3557444&utm_medium=issues&utm_source=github).
</bountysource-plugin>

krystof1119:
This PR adds a large cursor bitmap to both versions of vncviewer
(Java and native) which can be used instead of the previously used dot
cursor, as well as the options and parameters to be able to enable or
disable its use at runtime.

Resolves #1483

### DIFF
--- a/java/com/tigervnc/vncviewer/OptionsDialog.java
+++ b/java/com/tigervnc/vncviewer/OptionsDialog.java
@@ -152,8 +152,10 @@ class OptionsDialog extends Dialog {
 
   /* Misc. */
   JCheckBox sharedCheckbox;
-  JCheckBox dotWhenNoCursorCheckbox;
+  JCheckBox alwaysCursorCheckbox;
   JCheckBox acceptBellCheckbox;
+
+  JComboBox cursorTypeChoice;
 
   /* SSH */
   JCheckBox tunnelCheckbox;
@@ -311,6 +313,7 @@ class OptionsDialog extends Dialog {
     handleAutoselect();
     handleCompression();
     handleJpeg();
+    handleAlwaysCursor();
 
     /* Security */
     Security security = new Security(SecurityClient.secTypes);
@@ -458,7 +461,9 @@ class OptionsDialog extends Dialog {
 
     /* Misc. */
     sharedCheckbox.setSelected(shared.getValue());
-    dotWhenNoCursorCheckbox.setSelected(dotWhenNoCursor.getValue());
+    alwaysCursorCheckbox.setSelected(alwaysCursor.getValue());
+    String cursorTypeStr = cursorType.getValueStr();
+    cursorTypeChoice.setSelectedItem(cursorTypeStr);
     acceptBellCheckbox.setSelected(acceptBell.getValue());
 
     /* SSH */
@@ -613,8 +618,9 @@ class OptionsDialog extends Dialog {
 
     /* Misc. */
     shared.setParam(sharedCheckbox.isSelected());
-    dotWhenNoCursor.setParam(dotWhenNoCursorCheckbox.isSelected());
+    alwaysCursor.setParam(alwaysCursorCheckbox.isSelected());
     acceptBell.setParam(acceptBellCheckbox.isSelected());
+    cursorType.setParam((String)cursorTypeChoice.getSelectedItem());
 
     /* SSH */
     tunnel.setParam(tunnelCheckbox.isSelected());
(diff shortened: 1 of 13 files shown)


krystof1119:
Okay, that's fair. I'll leave this PR open for when we (hopefully) get more info from the requester.

CendioOssman:
It doesn't seem like any real progress is being made on #1483, but the issue seems to be better support for VNC servers without fancy cursor handling.

With that in mind, we can probably look at progressing here. Are you still interested in working on this, @krystof1119?

krystof1119:
It depends on the scope of the project. If all we're talking about is adding an option to change the local cursor displayed, I could probably build on what I had made previously, but I don't have much experience with the VNC protocol itself, so if the changes were more involved than that, I might have to see whether or not it's beyond me.

CendioOssman:
No protocol changes, @krystof1119. Just better client-side options.

vncviewer/Viewport.cxx
```diff
@@ -243,6 +243,44 @@ static const char * dotcursor_xpm[] = {
   " ... ",
   "     "};
 
+static const char * largecursor_xpm[] = {
```
CendioOssman, line 246: If we're going to use a standard pointer, then I think we get a better result if we just use the standard system pointer.

vncviewer/parameters.cxx
```diff
@@ -64,6 +64,9 @@ BoolParameter emulateMiddleButton("EmulateMiddleButton",
 BoolParameter dotWhenNoCursor("DotWhenNoCursor",
                               "Show the dot cursor when the server sends an "
                               "invisible cursor", false);
+BoolParameter largeCursorInsteadOfDot("LargeCursorInsteadOfDot",
+                                      "Use a large cursor bitmap instead of a dot "
+                                      "when using the dot cursor", false);
```
CendioOssman, line 69: Having multiple boolean options feels confusing. I think this is now a multiple-choice question, where we pick between "no cursor", "dot cursor", or "pointer cursor".

CendioOssman (changes requested):

krystof1119:
Okay, I'll try and get a bit of work done on this - probably not going to have time this week, but I should be able to look at this after.

vncviewer/Viewport.cxx
krystof1119: I'm not entirely sure what you mean by "standard system pointer" - I was trying to follow the way this is done just a few lines earlier for the dot cursor, but if there's a better way of doing this, I'm open to using it.

CendioOssman: For FLTK that means calling `cursor(FL_CURSOR_DEFAULT);`.

krystof1119:
It took me longer than I thought to get around to this, but I've got an implementation done in a different branch to the one in this PR ([multipleCursorTypes](https://github.com/krystof1119/tigervnc/tree/multipleCursorTypes)). A few notes:

- It seems that the frameworks don't allow getting the bitmap of the system mouse pointer, so I needed to special-case the calls to set the mouse pointer (the code is a bit ugly, I might still change it).
- I renamed the config option to show the dot cursor since it's not necessarily a dot cursor anymore. I don't know if I should put in some code to handle setting the new config option if the old is detected, or if it would be better to keep the old name (even if it's not accurate anymore).
- This is my first time working with these frameworks, so I'd appreciate it if someone went over my code, just to check I haven't done something stupid. Tying into this: I implemented the radio buttons with a string configuration option (it looks like that's how they're implemented in the rest of the codebase), so I need to run strcmp every time the cursor is set. This only seems to happen whenever the cursor enters the window area, so it's not too expensive, but it still feels weird.

I can force-push the changes onto the branch associated with this PR after someone reviews the code, I just didn't want to do it before.

CendioOssman:
It's difficult to review it if it's not here in the PR. Please do a force push and we can have a look at the details.

krystof1119:
Okay, I've force pushed my code to this branch. I've also made one minor change to make the CI checks pass (I had used a NULL somewhere a nullptr is now expected).

CendioOssman (changes requested):
Sorry for the late reply. I've been away for a few week.s

Looks really promising! Just a few tweaks needed.

vncviewer/OptionsDialog.cxx
```diff
@@ -835,11 +847,46 @@ void OptionsDialog::createInputPage(int tx, int ty, int tw, int th)
...
+                                                       _("Show local cursor when not provided by server")));
+    localCursorCheckbox->callback(handleLocalCursor, this);
     ty += CHECK_HEIGHT + TIGHT_MARGIN;
+
+    /* Cursor type box */
+    ty += GROUP_LABEL_OFFSET;
+    cursorGroup = new Fl_Group(tx, ty, width, 0, _("Cursor type"));
+    cursorGroup->box(FL_FLAT_BOX);
+    cursorGroup->labelfont(FL_BOLD);
+    cursorGroup->align(FL_ALIGN_LEFT | FL_ALIGN_TOP);
```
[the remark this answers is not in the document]
krystof1119: This reminds me: I wanted to ask, should this be a checkbox to enable + multiple choice for type, or a multiple choice-only question for type with "None" as an option (and config options to match)? Examples below:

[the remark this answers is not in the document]
CendioOssman: I think the latter is too confusing, as it loses all information on when this setting is relevant. So I say go with the first variant.

krystof1119:
I've made the requested changes, including changes to the equivalent parts of the Java version.

krystof1119:
The previous version did not work correctly with the system cursor enabled after the VNC client lost and regained focus, this fixes it (my bad for pushing without testing that).

vncviewer/Viewport.cxx
```diff
@@ -122,7 +122,9 @@ Viewport::Viewport(int w, int h, const rfb::PixelFormat& /*serverPF*/, CConn* cc
     altGrArmed(false),
 #endif
     firstLEDState(true), pendingClientClipboard(false),
-    menuCtrlKey(false), menuAltKey(false), cursor(nullptr)
+    menuCtrlKey(false), menuAltKey(false), cursor(nullptr),
+    useSystemCursorIfBlank(alwaysCursor && !strcasecmp("system", cursorType)),
```
CendioOssman, line 126: These values can change, so they cannot be checked just once here.

CendioOssman, line 126: The fallback logic here is different from the options dialog. Here you get the dot cursor for invalid values, whilst in the options dialog you get the system cursor for invalid values.

For sanity, it might be best if the fallback is the same as the default value.

vncviewer/Viewport.cxx
```diff
@@ -279,8 +284,18 @@ void Viewport::setCursor(int width, int height, const Point& hotspot,
     }
   }
 
-  if (Fl::belowmouse() == this)
+  if (Fl::belowmouse() == this) {
+    showCursor();
+  }
```
CendioOssman, line 289: We don't generally use brackets around single statements.

vncviewer/Viewport.cxx
```diff
@@ -279,8 +284,18 @@ void Viewport::setCursor(int width, int height, const Point& hotspot,
...
+}
+
+void Viewport::showCursor()
```
CendioOssman, line 292: Nice cleanup!

vncviewer/Viewport.h
```diff
@@ -51,6 +51,9 @@ class Viewport : public Fl_Widget, public EmulateMB {
   void setCursor(int width, int height, const rfb::Point& hotspot,
                  const uint8_t* data);
 
+  // Show the currently set (or system) cursor
+  void showCursor();
```
CendioOssman, line 55: This is an internal helper, so it's best put in the private section.

CendioOssman (changes requested):

krystof1119:
Sorry about the multiple separate force pushes, I noticed some of your comments too late.

I was hoping to avoid calling strcasecmp every time the cursor changes, but I imagine it's not going to be noticeable performance-wise. Changing the cursor type while the Java VNC viewer is running is still not working entirely correctly, but I don't think that's solvable without more extensive changes (and it seems like it wasn't working perfectly before, so at least it's not a regression).

I also decided to change the fallback cursor type to "Dot" in the options dialog rather than changing it to "System" here, since that was the previous behavior.
Other than that, I've made the two remaining changes, thanks for catching that.

CendioOssman (approved):
Looks good to me. Nice work!

@bphinz, are you happy with the Java version?

java/com/tigervnc/vncviewer/OptionsDialog.java
```diff
@@ -1173,31 +1179,54 @@ private JPanel createMiscPanel() {
...
+    alwaysCursorCheckbox = new JCheckBox("Show local cursor when not provided by server");
+    alwaysCursorCheckbox.addItemListener(new ItemListener() {
+      public void itemStateChanged(ItemEvent e) {
+        handleAlwaysCursor();
+      }
+    });
+    JLabel cursorTypeLabel = new JLabel("Cursor type:");
+    String[] cursorTypes = {"Dot", "System"};
+    cursorTypeChoice = new MyJComboBox(cursorTypes);
+    cursorTypeChoice.setEditable(true);
```
bphinz, line 1191: Why is this editable?  Doing so allows for an arbitrary string to entered but then subsequently only checks for "Dot" or "System".

krystof1119: Because I got setEditable confused with setEnabled and didn't notice the difference in behavior when testing. I've corrected it now.

CendioOssman:
What's the status here? Are there more things that you want done, @bphinz?

bphinz:
Nope, let’s go ahead and merge it

PR #1506, opened 2022-08-01
Correctly recognize the upper left (ISO) key of European keyboards on macOS.

shysaur:
On macOS, the latest version of TigerVNC ignores the so-called ISO key present on some European keyboards following the ISO 9995 standard. This key is located at the upper left corner of the keyboard (see this image of the Italian keyboard on Wikipedia commons for reference: https://commons.wikimedia.org/wiki/File:Italian_Keyboard_layout.svg#/media/File:Italian_Keyboard_layout.svg) Affected keyboards, apart from the aforementioned Italian layout, include the UK and Spanish layouts.

The key is not recognized because the corresponding macOS scancode is missing in the map to qnum scancodes. This patch simply adds the missing scancode. I tested the patch on macOS 12.5 with an Italian-layout laptop keyboard.

### DIFF
--- a/vncviewer/osx_to_qnum.c
+++ b/vncviewer/osx_to_qnum.c
@@ -15,6 +15,7 @@ const unsigned short code_map_osx_to_qnum[256] = {
   [0x7] = 0x2d, /* osx:7 (ANSI_X) -> linux:45 (KEY_X) -> qnum:45 */
   [0x8] = 0x2e, /* osx:8 (ANSI_C) -> linux:46 (KEY_C) -> qnum:46 */
   [0x9] = 0x2f, /* osx:9 (ANSI_V) -> linux:47 (KEY_V) -> qnum:47 */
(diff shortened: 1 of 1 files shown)


CendioOssman:
Thanks for reporting. I can reproduce the issue here.

Unfortunately, the issue seems to be a bit more complex. That key should send a different key code. But macOS is also sending different key codes depending on the chosen layout. So, I think some more mapping might be needed.

Let me have a look and see if I can get something dynamic working.

CendioOssman:
Something more complete has been merged as 97f51df7013ac818ec3e3c94b6c07b1300de1acd. Please give it a go, and it should hopefully work well for you.

Thanks again for reporting this!

shysaur:
Your more complete fix works great for me. Thanks!

CendioOssman:
Glad to hear it. Thanks for testing!

Issue #1224, opened 2021-03-23
x0vncserver - Could not map key event to X11 key code -- Ubuntu Focal (20.04)


delphir-com:
I just upgraded from 18.04 to 20.04 and got an issue with the Cyrillic (Russian) locale when using with x0vncserver 1.10.1

each time when I'm expecting to type a cyrillic character I`m getting a console message:

```
XDesktop:    Could not map key event to X11 key code
```

I manually downgraded to 1.7.0 (which is the actual version in 18.04) and the issue disappeared. 
I also tested tigervnc-1.11.80-20210322git06e53809.x86_64 from nightly, but it still has the same issue.

Client:
Ubuntu 20.04 + remmina=1.4.11+ppa202102060756.r0a51356f.deec7640c~ubuntu20.04.1

Server:
Ubuntu 20.04 + x0vncserver above 1.7

let me know if you need any extra info, I can easily reproduce this issue

CendioOssman:
Does this happen with our client as well?

delphir-com:
yes, exactly the same behavior (tigervnc-viewer 1.10.1)

delphir-com:
a bit of an update:

the issue happens when the client's and server's current locale is not the same.

for example if client's current language is set to "EN"glish, but server's "RU"ssian - then no chars appear on the remote.
the same happens if client is RU but server is EN.

if both are RU or both are EN - then the correct characters are printed while typing.

CendioOssman:
I'm afraid then it sounds like things are working as expected. It was some kind of bug (likely ac73038232dd7d6e41006357de2120f580b9f05f and related fixes)  in earlier versions that happened to stumble in to something sane in your specific case.

If you want to ignore the client's keymap then I suggest trying `-RawKeyboard=1` and see if that is more to your liking. Otherwise I'm afraid there isn't much we can do as we can only work with the keys available in the server's keymap.

delphir-com:
then why is it working correctly in x0vncserver 1.7 ?

CendioOssman:
I don't think it worked the way it was supposed there either. It just happened that the buggy behaviour was something that was still useful in your case.

When you had Russian configured on the client and English on the server, I'm assuming you did not get Cyrillic characters in the VNC session?

delphir-com:
yes, I don't get any chars when client/host language is set to a different values. but in your example I would expect to get English chars as like you say "English on the server" so I would expect to get chars in server's locale.

[...]

PR #1510, opened 2022-08-09
x0vncserver: add new keysym in case we don't find a matching keycode

grulja:
We might often fail to find a matching X11 keycode when the client has a different keyboard layout and end up with no key event. To avoid a failure we add it as a new keysym/keycode pair so the next time a keysym from the client that is unknown to the server is send, we will find a match and proceed with key event. This is same behavior used in Xvnc or x11vnc, although Xvnc has more advanced mapping from keysym to keycode.

### DIFF
--- a/unix/x0vncserver/XDesktop.cxx
+++ b/unix/x0vncserver/XDesktop.cxx
@@ -31,6 +31,7 @@
 #include <x0vncserver/XDesktop.h>
 
 #include <X11/XKBlib.h>
+#include <X11/Xutil.h>
 #ifdef HAVE_XTEST
 #include <X11/extensions/XTest.h>
 #endif
@@ -50,6 +51,7 @@ void vncSetGlueContext(Display *dpy, void *res);
 #include <x0vncserver/Geometry.h>
 #include <x0vncserver/XPixelBuffer.h>
 
+using namespace std;
 using namespace rfb;
 
 extern const unsigned short code_map_qnum_to_xorgevdev[];
@@ -264,6 +266,9 @@ void XDesktop::start(VNCServer* vs) {
 void XDesktop::stop() {
   running = false;
 
+  // Delete added keycodes
+  deleteAddedKeysyms(dpy);
+
 #ifdef HAVE_XDAMAGE
   if (haveDamage)
     XDamageDestroy(dpy, damage);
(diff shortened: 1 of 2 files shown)


grulja:
This should fix issues like https://github.com/TigerVNC/tigervnc/issues/1224, however, from what I read from @CendioOssman this is not what it's supposed to do. Maybe we can hide this under an option that will be disabled by default and users can enable it if they want to?

CendioOssman:
Well, it's what it is currently designed to do. Doesn't mean we don't want to do better. :)

This code seems to use the old core API, though. That is handled through various magical and buggy compatibility code. Do you think you could look at handling this the XKB way instead?

Ideally, we also want to merge this with the code in Xvnc somehow. We want to avoid having multiple implementations of complex stuff like this.

grulja:
> Well, it's what it is currently designed to do. Doesn't mean we don't want to do better. :)
> 
> This code seems to use the old core API, though. That is handled through various magical and buggy compatibility code. Do you think you could look at handling this the XKB way instead?
> 
> Ideally, we also want to merge this with the code in Xvnc somehow. We want to avoid having multiple implementations of complex stuff like this.

I wanted to use XKB instead, but some of the API you use in Xvnc doesn't seem to be public and that's why you have to build Xvnc with xorg-xserver sources.

CendioOssman:
There should be equivalent API. In fact, I think the normal behaviour is that the XKB configuration is assembled outside of the X server (by xkbcomp).

grulja:
I updated this change to use XKB instead of XLib API. 

@CendioOssman can you please take a look? I tested it and it works for me.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -383,6 +391,92 @@ KeyCode XDesktop::XkbKeysymToKeycode(Display* dpy, KeySym keysym) {
...
+
+void XDesktop::deleteAddedKeycodes(Display* dpy) {
+  XkbDescPtr xkb;
+  xkb = XkbGetMap(dpy, XkbAllComponentsMask, XkbUseCoreKbd);
+
+  if (!xkb)
+    return;
+
+  for (int key = xkb->min_key_code; key < xkb->max_key_code; key++) {
+    if (added_keysyms[key] != NoSymbol && XkbKeyNumGroups(xkb, key) != 0) {
```
CendioOssman, line 443: This will wipe any changes the user might have made. We should probably try to confirm that it is our changes that we are deleting.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -383,6 +391,92 @@ KeyCode XDesktop::XkbKeysymToKeycode(Display* dpy, KeySym keysym) {
...
+      XkbMapChangesRec changes;
+      memset(&changes, 0, sizeof(changes));
+
+      XkbChangeTypesOfKey(xkb, key, 0, XkbGroup1Mask, NULL, &changes);
+
+      changes.changed |= XkbKeySymsMask;
+      changes.first_key_sym = key;
+      changes.num_key_syms = 1;
+
+      XkbChangeMap(dpy, xkb, &changes);
```
CendioOssman, line 453: Should this perhaps be accumulated so that a single change is done for the entire cleanup?

unix/x0vncserver/XDesktop.cxx
```diff
@@ -383,6 +391,92 @@ KeyCode XDesktop::XkbKeysymToKeycode(Display* dpy, KeySym keysym) {
...
+      break;
+  }
+
+  if (key < xkb->min_key_code)
+    return false;
+
+  memset(&changes, 0, sizeof(changes));
+
+  syms = XkbKeySymsPtr(xkb, key);
+  syms[0] = keysym;
```
CendioOssman, line 416: You sure this order is safe? I'm thinking that there might not be any space allocated before `XkbChangeTypesOfKey()` is called.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -383,6 +391,92 @@ KeyCode XDesktop::XkbKeysymToKeycode(Display* dpy, KeySym keysym) {
...
+
+  int types[XkbNumKbdGroups] = { XkbOneLevelIndex };
+  XkbChangeTypesOfKey(xkb, key, 1, XkbGroup1Mask, types, &changes);
+  XkbKeySymEntry(xkb, key, 0, 0) = keysym;
```
CendioOssman, line 420: This seems redundant?

And unsafe. That function is not supposed to return a reference according to the documentation.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -383,6 +391,92 @@ KeyCode XDesktop::XkbKeysymToKeycode(Display* dpy, KeySym keysym) {
...
+
+  if (key < xkb->min_key_code)
+    return false;
+
+  memset(&changes, 0, sizeof(changes));
+
+  syms = XkbKeySymsPtr(xkb, key);
+  syms[0] = keysym;
+
+  int types[XkbNumKbdGroups] = { XkbOneLevelIndex };
```
CendioOssman, line 418: Did you check the Xvnc code? It has a comment saying that this doesn't work properly with CapsLock.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -383,6 +391,92 @@ KeyCode XDesktop::XkbKeysymToKeycode(Display* dpy, KeySym keysym) {
...
+  XkbChangeTypesOfKey(xkb, key, 1, XkbGroup1Mask, types, &changes);
+  XkbKeySymEntry(xkb, key, 0, 0) = keysym;
+
+  changes.changed |= XkbKeySymsMask;
+  changes.first_key_sym = key;
+  changes.num_key_syms = 1;
+
+  if (XkbChangeMap(dpy, xkb, &changes)) {
+    vlog.info("Added unknown keysym 0x%lx to keycode %d", keysym, key);
```
CendioOssman, line 427: `XKeysymToString()` for clarity?

unix/x0vncserver/XDesktop.cxx
```diff
@@ -383,6 +391,92 @@ KeyCode XDesktop::XkbKeysymToKeycode(Display* dpy, KeySym keysym) {
...
+
+  // TODO: try to further guess keycode with all possible mods as Xvnc does
+
+  if (keycode == 0) {
+    keycode = addKeysym(dpy, keysym);
+    if (keycode == 0) {
+      vlog.error("Failure adding new keysym 0x%lx", keysym);
+      return keycode;
+    }
+  }
```
CendioOssman, line 474: Perhaps consider inverting this logic to avoid nesting. I.e.:

```C++
if (keycode != 0)
  return keycode;
```

unix/x0vncserver/XDesktop.cxx
```diff
@@ -383,6 +391,92 @@ KeyCode XDesktop::XkbKeysymToKeycode(Display* dpy, KeySym keysym) {
...
+  if (XkbChangeMap(dpy, xkb, &changes)) {
+    vlog.info("Added unknown keysym 0x%lx to keycode %d", keysym, key);
+    added_keysyms[key] = keysym;
+    return key;
+  }
+
+  return 0;
+}
+
+void XDesktop::deleteAddedKeycodes(Display* dpy) {
```
CendioOssman, line 435: `deleteAddedKeysyms()`, for consistency.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -113,6 +114,10 @@ XDesktop::XDesktop(Display* dpy_, Geometry *geometry_)
       ledState |= 1u << i;
   }
 
+  // Initialize our list for newly added keysyms
+  added_keysyms.resize(256);
+  fill(added_keysyms.begin(), added_keysyms.end(), NoSymbol);
```
CendioOssman, line 119: Why not a `std::map` like the pressed keys? I assume this vector will be mostly unused anyway as most keycodes are already taken.

CendioOssman (changes requested):
Looks promising, but there are some possible problems. Please have a look at the comments.

unix/x0vncserver/XDesktop.cxx
grulja: Shouldn't all added keysyms in this `vector` (now `map`) be ours? I always put it there only when we successfuly add a new keysym.

unix/x0vncserver/XDesktop.cxx
grulja: I tried that, but it doesn't work for some reason and always removes just the last one. I updated the change with it so you can see how I try and you hopefully you will see what I'm doing wrong.

grulja: Fixed. I expected it works a bit differently.

unix/x0vncserver/XDesktop.cxx
CendioOssman: No, because other tools might also update the keyboard layout. E.g. `setxkbmap`. x0vncserver doesn't have exclusive control of the layout.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -383,6 +388,118 @@ KeyCode XDesktop::XkbKeysymToKeycode(Display* dpy, KeySym keysym) {
...
+  unsigned int key;
+  XkbDescPtr xkb;
+  XkbMapChangesRec changes;
+  KeySym *syms;
+  KeySym upper, lower;
+
+  xkb = XkbGetMap(dpy, XkbAllComponentsMask, XkbUseCoreKbd);
+
+  if (!xkb)
+    return false;
```
CendioOssman, line 403: I guess this technically works, but it is confusing. Returning 0 would be preferred here as that is what the caller checks for.

CendioOssman (changes requested):
Looks nice!

There's just a teeny detail, and then I can go ahead and merge this.

unix/x0vncserver/XDesktop.cxx
grulja: Fixed, same wrong return was also few lines below.

CendioOssman (approved):
Thanks!

PR #1513, opened 2022-08-11
x0vncserver: update/display cursor only on correct screen in zaphod mode

grulja:
We have to check whether we update cursor position/shape only in case the cursor is on our display, otherwise in zaphod mode, ie. when having two instances of x0vncserver on screens :0.0 and :0.1 we would be having the cursor duplicated and actually not funcional (aka ghost cursor) as it would be actually not present. We also additionally watch EnterNotify and LeaveNotify events in order to show/hide cursor accordingly.

Change made with help from Olivier Fourdan (@ofourdan)

### DIFF
--- a/unix/x0vncserver/XDesktop.cxx
+++ b/unix/x0vncserver/XDesktop.cxx
@@ -192,7 +192,8 @@ XDesktop::XDesktop(Display* dpy_, Geometry *geometry_)
                    RRScreenChangeNotifyMask | RRCrtcChangeNotifyMask);
     /* Override TXWindow::init input mask */
     XSelectInput(dpy, DefaultRootWindow(dpy),
-                 PropertyChangeMask | StructureNotifyMask | ExposureMask);
+                 PropertyChangeMask | StructureNotifyMask |
+                 ExposureMask | EnterWindowMask | LeaveWindowMask);
   } else {
 #endif
     vlog.info("RANDR extension not present");
@@ -217,11 +218,13 @@ void XDesktop::poll() {
     Window root, child;
     int x, y, wx, wy;
     unsigned int mask;
-    XQueryPointer(dpy, DefaultRootWindow(dpy), &root, &child,
-                  &x, &y, &wx, &wy, &mask);
-    x -= geometry->offsetLeft();
-    y -= geometry->offsetTop();
-    server->setCursorPos(rfb::Point(x, y), false);
+
+    if (XQueryPointer(dpy, DefaultRootWindow(dpy), &root, &child,
+                      &x, &y, &wx, &wy, &mask)) {
+      x -= geometry->offsetLeft();
+      y -= geometry->offsetTop();
+      server->setCursorPos(rfb::Point(x, y), false);
+    }
   }
 }
 
@@ -253,7 +256,14 @@ void XDesktop::start(VNCServer* vs) {
 #endif
 
 #ifdef HAVE_XFIXES
-  setCursor();
+  Window root, child;
+  int x, y, wx, wy;
+  unsigned int mask;
+  // Check whether the cursor is initially on our screen
+  if (XQueryPointer(dpy, DefaultRootWindow(dpy), &root, &child,
+                    &x, &y, &wx, &wy, &mask))
+    setCursor();
+
 #endif
 
   server->setLEDState(ledState);
@@ -701,6 +711,15 @@ bool XDesktop::handleGlobalEvent(XEvent* ev) {
     if (cev->subtype != XFixesDisplayCursorNotify)
       return false;
 
+    Window root, child;
+    int x, y, wx, wy;
+    unsigned int mask;
+
+    // Check whether the cursor is initially on our screen
+    if (!XQueryPointer(dpy, DefaultRootWindow(dpy), &root, &child,
+                      &x, &y, &wx, &wy, &mask))
+      return false;
+
     return setCursor();
 #endif
 #ifdef HAVE_XRANDR
@@ -751,6 +770,33 @@ bool XDesktop::handleGlobalEvent(XEvent* ev) {
       server->setScreenLayout(computeScreenLayout());
     }
 
+    return true;
+#endif
+#ifdef HAVE_XFIXES
+  } else if (ev->type == EnterNotify) {
+    XCrossingEvent* cev;
+
+    if (!running)
+      return true;
+
+    cev = (XCrossingEvent*)ev;
+
+    if (cev->window != cev->root)
+      return false;
+
+    return setCursor();
+  } else if (ev->type == LeaveNotify) {
+    XCrossingEvent* cev;
+
+    if (!running)
+      return true;
+
+    cev = (XCrossingEvent*)ev;
+
+    if (cev->window == cev->root)
+      return false;
+
+    server->setCursor(0, 0, Point(), NULL);
     return true;
 #endif
   }


unix/x0vncserver/XDesktop.cxx
```diff
@@ -192,7 +192,7 @@ XDesktop::XDesktop(Display* dpy_, Geometry *geometry_)
                    RRScreenChangeNotifyMask | RRCrtcChangeNotifyMask);
     /* Override TXWindow::init input mask */
     XSelectInput(dpy, DefaultRootWindow(dpy),
-                 PropertyChangeMask | StructureNotifyMask | ExposureMask);
+                 PropertyChangeMask | StructureNotifyMask | ExposureMask | EnterWindowMask | LeaveWindowMask);
```
CendioOssman, line 195: This line looks long. Please try to keep things under 72 columns.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -701,6 +711,9 @@ bool XDesktop::handleGlobalEvent(XEvent* ev) {
     if (cev->subtype != XFixesDisplayCursorNotify)
       return false;
 
+    if (!isCursorVisibleOnScreen)
+      return false;
```
CendioOssman, line 715: You seem to check this actively in the other places. Why not here?

CendioOssman (changes requested):
Code looks mostly good. But I'm not sure I understand what the actual issue is. Won't the coordinates be out of screen and the cursor hidden anyway?

unix/x0vncserver/XDesktop.cxx
grulja: I guess you mean why I don't check it in `EnterNotfiy` and `LeaveNotify` events? It's because we want to either hide or show the cursor.

grulja:
> Code looks mostly good. But I'm not sure I understand what the actual issue is. Won't the coordinates be out of screen and the cursor hidden anyway?

XQueryPointer() will report correct coordinates, but for other screen. From the documentation:
`If XQueryPointer() returns False, the pointer is not on the same screen as the specified window,`

We also need to hook up `LeaveNotify` and `EnterNotify` to correctly show/hide the cursor as it moves between screens.

unix/x0vncserver/XDesktop.cxx
CendioOssman: No, I mean here. Why rely on the cached `isCursorVisibleOnScreen` here rather than calling `XQueryPointer()` and checking the current state directly?

It would mean not having to have that cached value at all.

grulja: I was told that XQueryPointer() is an expensive call so that's why it is cached and not called again.

CendioOssman:
> 
> XQueryPointer() will report correct coordinates, but for other screen. From the documentation: `If XQueryPointer() returns False, the pointer is not on the same screen as the specified window,`
> 

Ah. I see the issue now. We're grabbing the root relative coordinates. What if we use the window relative coordinates instead? Should be simpler as we don't need all the extra logic?

> We also need to hook up `LeaveNotify` and `EnterNotify` to correctly show/hide the cursor as it moves between screens.

It looks like it is only called when we enter the correct screen, though. Where's the hiding handled?

unix/x0vncserver/XDesktop.cxx
grulja: I do it only once in the `Start()` as we might also show the cursor on the wrong screen initially.

CendioOssman: That bridge is long passed. You are now calling it in the most expensive place (each loop), but caching for the least expensive place (when the cursor image changes). :)

CendioOssman: It is called once per loop, so we can track where the cursor is on screen anyway. So, we are constantly spamming that call.

grulja:
> > XQueryPointer() will report correct coordinates, but for other screen. From the documentation: `If XQueryPointer() returns False, the pointer is not on the same screen as the specified window,`
> 
> Ah. I see the issue now. We're grabbing the root relative coordinates. What if we use the window relative coordinates instead? Should be simpler as we don't need all the extra logic?
> 

We would still need to check the cursor location for the XFixes event if I'm not mistaken.

> > We also need to hook up `LeaveNotify` and `EnterNotify` to correctly show/hide the cursor as it moves between screens.
> 
> It looks like it is only called when we enter the correct screen, though. Where's the hiding handled?

I thought the hiding is done through `setCursor()`.

unix/x0vncserver/XDesktop.cxx
grulja: I changed it to not use cached value.

CendioOssman:
> We would still need to check the cursor location for the XFixes event if I'm not mistaken.

Maybe? I guess we would be sending the cursor image from the other screen otherwise. But is that so bad?

> > It looks like it is only called when we enter the correct screen, though. Where's the hiding handled?
> 
> I thought the hiding is done through `setCursor()`.

It will send whatever cursor image it gets back. Which likely is the current image, even if it happens to be on another screen. You also added `if (cev->window != cev->root)`, which I'm not entirely sure when it will trigger with regard to enter and leave events.

grulja:
> > We would still need to check the cursor location for the XFixes event if I'm not mistaken.
> 
> Maybe? I guess we would be sending the cursor image from the other screen otherwise. But is that so bad?
> 

I believe this would make the cursor to show up again at the position it was left on the first screen when moving to the other one.

> > > It looks like it is only called when we enter the correct screen, though. Where's the hiding handled?
> > 
> > 
> > I thought the hiding is done through `setCursor()`.
> 
> It will send whatever cursor image it gets back. Which likely is the current image, even if it happens to be on another screen. You also added `if (cev->window != cev->root)`, which I'm not entirely sure when it will trigger with regard to enter and leave events.

Ok, I'm not sure right now what makes the cursor hidden when moving from one screen to the other.  When you get `LeaveNotify` or `EnterNotify` for the screen you are interested in it should be `cev->window == cev->root`. This was all tested by @ofourdan and he was also the one who gave me all the information about this API. Unfortunately he will be on vacation for two more weeks.

grulja:
Alright, I managed to create the setup and reproduce the issue. You were right with the LeaveEvent so I fixed it and made it to actually hide the cursor. The check for the cursor position in the XFixes event is needed, otherwise it will again show the cursor at the position it was left on.

unix/x0vncserver/XDesktop.cxx
```diff
@@ -753,6 +772,33 @@ bool XDesktop::handleGlobalEvent(XEvent* ev) {
...
+  } else if (ev->type == LeaveNotify) {
+    XCrossingEvent* cev;
+
+    if (!running)
+      return true;
+
+    cev = (XCrossingEvent*)ev;
+
+    if (cev->window == cev->root)
+      return false;
```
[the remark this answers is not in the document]
CendioOssman: It also doesn't seem to trigger when intended as I don't get a blank cursor when leaving the screen.

CendioOssman (changes requested):
I tested this here using Xephyr, and it mostly works, but I'm seeing some bugs:

 * When leaving a screen, the last coordinate seem to get stuck (move the mouse quickly, so the last coordinate isn't at the edge)
 * If the other screen is active locally, then x0vncserver cannot get the screen it is working with active again, effectively going view only

unix/x0vncserver/XDesktop.cxx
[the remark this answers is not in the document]
grulja: Could this be because of Xephyr? I have never used it so I don't know how it works. I tested this one more time, adding debug messages into both `EnterNotify` and `LeaveNotify` as well as debug messages after the check so I can see whether it makes the cursor visible or whether it gets hidden and it does what I expect:
```
Thu Sep  1 08:09:17 2022
 XDesktop:    EnterNotify
 XDesktop:    Showing the cursor

Thu Sep  1 08:09:18 2022
 XDesktop:    LeaveNotify
 XDesktop:    Hiding the cursor
```
The `EnterNotify` is called correctly when moving back to the screen I'm connected to and same with the `LeaveNotify`.

grulja:
In case you have issues to create the setup, this is my xorg.conf file:
```
Section "Device"
    Identifier     "Device0"
    Driver         "modesetting"
    VendorName     "Intel"
    BusID          "PCI:0:2:0"
    Screen          0
    Option         "ZaphodHeads" "eDP-1"
EndSection

Section "Device"
    Identifier     "Device1"
    Driver         "modesetting"
    VendorName     "Intel"
    BusID          "PCI:0:2:0"
    Screen          1
    Option         "ZaphodHeads" "DP-5"
EndSection

Section "Screen"
    Identifier "Screen0"
    Device     "Device0"
    SubSection     "Display"
        Depth       24
    EndSubSection
EndSection

Section "Screen"
    Identifier "Screen1"
    Device     "Device1"
    SubSection     "Display"
        Depth       24
    EndSubSection
EndSection

Section "Monitor"
    # HorizSync source: edid, VertRefresh source: edid
    Identifier     "Monitor0"
    VendorName     "Unknown"
    Option         "DPMS"
EndSection

Section "Monitor"
    # HorizSync source: edid, VertRefresh source: edid
    Identifier     "Monitor1"
    VendorName     "Unknown"
    Option         "DPMS"
EndSection

Section "ServerLayout"
    Identifier     "Layout0"
    Screen      0  "Screen0" 0 0
    Screen      1  "Screen1" RightOf "Screen0"
    InputDevice    "Keyboard0" "CoreKeyboard"
    InputDevice    "Mouse0" "CorePointer"
    Option         "Xinerama" "0"
EndSection

Section "InputDevice"
    # generated from default
    Identifier     "Mouse0"
    Driver         "mouse"
    Option         "Protocol" "auto"
    Option         "Device" "/dev/input/mice"
    Option         "Emulate3Buttons" "no"
    Option         "ZAxisMapping" "4 5"
EndSection

Section "InputDevice"
    # generated from default
    Identifier     "Keyboard0"
    Driver         "kbd"
EndSection
```
And this is what I run:
```
#!/bin/sh
xsetroot -display :0.0 -solid grey30
xsetroot -display :0.1 -solid grey30
mwm -xrm "Mwm*freezeOnConfig: False" -multiscreen
```
I test this with mwm (Motif Window Manager?) as that's what the customer, who reported this issue, is using and I'm not sure whether this setup is well supported by any other major desktop environment.

unix/x0vncserver/XDesktop.cxx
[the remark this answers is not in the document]
CendioOssman: I doubt it is Xephyr, at least not all of the symptoms. Are you testing with a full desktop environment? Or with a minimal case? A key issue might be if the root window is actually visible, or if it is obscured by a "desktop" window.

In my case, I have just xterm and metacity running.

CendioOssman:
I'm afraid I don't have a physical machine to butcher right now, and vSphere doesn't seem to handle multiple monitors. Hence trying Xephyr.

unix/x0vncserver/XDesktop.cxx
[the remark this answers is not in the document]
grulja: I test this with Motif Window Manager (MWM), but now I also tried Metacity, using just `metacity` instead of the `mwm` command in the configuration I provided below and even though I couldn't do anything (e.g. type into xterm), it was still working for me as expected.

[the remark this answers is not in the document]
grulja: I will try with Xephyr.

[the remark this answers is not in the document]
grulja: Ok, I have tested Xephyr, but I don't think it can be used to verify this change. With Xephyr you have two separate X sessions, while with zaphod mode I think they are supposed to act like one. That's the reason why there was an issue as x0vncserver was getting events for the other screen.

CendioOssman (approved):
Sorry for the long delay here. I got a physical setup now (finally), and it seems to work well. I'm still seeing issues with Xephyr in zaphod mode, but let's not let that block things.

PR #1519, opened 2022-08-21
Add support for RSA-AES security types

pdlan:
This PR adds support for the RSA-AES security types including RA2, RA2ne, RA2_256, RA2ne_256, allowing the RealVNC/TigerVNC clients/servers to be used together with encryption enabled and with username/password authentication. I added to the CMakeLists.txt the code that searches for the Nettle cryptographic library (a dependency for GnuTLS) as well as libhogweed (a part of Nettle) and libgmp. Now only when they are found will the GnuTLS be enabled.


CendioOssman:
Ohh! Awesome!

I have a bit of a backlog of TigerVNC stuff, but I'll try to get to reviewing this soon.

common/rfb/CSecurityRSAAES.cxx
```diff
@@ -0,0 +1,444 @@
...
+      if (readHash())
+        state = ReadSubtype;
+    case ReadSubtype:
+      if (readSubtype()) {
+        writeCredentials();
+        return true;
+      }
+      return false;
+  }
+  throw AuthFailureException("unreachable");
```
CendioOssman, line 146: An `assert()` is generally more appropriate for protection against coding mistakes.

common/rfb/CSecurityRSAAES.cxx
```diff
@@ -0,0 +1,444 @@
...
+    case secTypeRA256:    return "RSA-AES-256 security type";
+    case secTypeRAne256:  return "RSA-AES-256 unencrypted security type";
+  }
+  return NULL;
+}
+
+static void random_func(void* ctx, size_t length, uint8_t* dst)
+{
+  gnutls_rnd(GNUTLS_RND_RANDOM, dst, length);
+}
```
CendioOssman, line 163: Why is this it's own method()? Can't things call the random function directly?

common/rfb/CSecurityRSAAES.cxx
```diff
@@ -0,0 +1,444 @@
...
+    NULL, random_func, NULL, NULL, clientKeyLength, 0))
+    throw AuthFailureException("failed to generate key");
+  clientKeyN = new rdr::U8[rsaKeySize];
+  clientKeyE = new rdr::U8[rsaKeySize];
+  nettle_mpz_get_str_256(rsaKeySize, clientKeyN, clientPublicKey.n);
+  nettle_mpz_get_str_256(rsaKeySize, clientKeyE, clientPublicKey.e);
+  rawos->writeU32(clientKeyLength);
+  rawos->writeBytes(clientKeyN, rsaKeySize);
+  rawos->writeBytes(clientKeyE, rsaKeySize);
+  rawos->flush();
```
CendioOssman, line 186: I would prefer you use normal stream names until the switch. This gives the impression that encryption is already active, and these writes are actively bypassing it.

common/rfb/CSecurityRSAAES.cxx
```diff
@@ -0,0 +1,444 @@
...
+  sha1_update(&ctx, serverKey.size, serverKeyE);
+  sha1_digest(&ctx, sizeof(f), f);
+  const char *title = "Server key fingerprint";
+  CharArray text;
+  text.format(
+    "The server has provided the following identifying information:\n"
+    "Fingerprint: %02x-%02x-%02x-%02x-%02x-%02x-%02x-%02x\n"
+    "Please verify that the information is correct and press \"Yes\". "
+    "Otherwise press \"No\"", f[0], f[1], f[2], f[3], f[4], f[5], f[6], f[7]);
+  if (!CSecurityTLS::msg->showMsgBox(UserMsgBox::M_YESNO, title, text.buf))
```
CendioOssman, line 237: This is an odd connection, so I think it is time to move this connection to `CSecurity` instead.

common/rfb/CSecurityRSAAES.cxx
```diff
@@ -0,0 +1,444 @@
...
+    sha256_digest(&ctx, 32, key);
+    rais = new rdr::AESInStream(rawis, key, 256);
+    sha256_init(&ctx);
+    sha256_update(&ctx, 32, serverRandom);
+    sha256_update(&ctx, 32, clientRandom);
+    sha256_digest(&ctx, 32, key);
+    raos = new rdr::AESOutStream(rawos, key, 256);
+  }
+  if (secType == secTypeRA2 || secType == secTypeRA256)
+    cc->setStreams(rais, raos);
```
CendioOssman, line 315: I think it might be clearer to always switch streams, as these will always be active for the rest of the handshake. So instead there will be an `if` if we should switch back or not after the handshake is done.

common/rfb/CSecurityRSAAES.cxx
```diff
@@ -0,0 +1,444 @@
...
+    throw AuthFailureException("hash doesn't match");
+  return true;
+}
+
+bool CSecurityRSAAES::readSubtype()
+{
+  if (!rais->hasData(1))
+    return false;
+  subtype = rais->readU8();
+  if (subtype != 1 && subtype != 2)
```
CendioOssman, line 416: Could we have some constants for these?

common/rfb/SSecurityPlain.cxx
```diff
@@ -36,7 +36,8 @@ using namespace rfb;
 
 StringParameter PasswordValidator::plainUsers
 ("PlainUsers",
- "Users permitted to access via Plain security type (including TLSPlain, X509Plain etc.)",
+ "Users permitted to access via Plain security type (including TLSPlain, X509Plain etc.) "
+ "or RSA-AES security types (RA2, RA2ne, RA2_256, RA2ne_256)",
```
CendioOssman, line 40: Not a huge fan of these names, and having so many of them. But it seems to be how RealVNC are naming things, so I guess we shouldn't come up with something different.

common/rfb/Security.cxx
```diff
@@ -163,6 +163,8 @@ rdr::U32 rfb::secTypeNum(const char* name)
   if (strcasecmp(name, "SSPI") == 0)       return secTypeSSPI;
   if (strcasecmp(name, "SSPIne") == 0)     return secTypeSSPIne;
   if (strcasecmp(name, "VeNCrypt") == 0)   return secTypeVeNCrypt;
+  if (strcasecmp(name, "RA2_256") == 0)    return secTypeRA256;
+  if (strcasecmp(name, "RA2ne_256") == 0)  return secTypeRAne256;
```
CendioOssman, line 167: Why not group them with the 128 bit ones?

CendioOssman (changes requested):
I haven't reviewed everything yet, but here's a start at least.

Have you tested with both RealVNC's client and server?

And with noVNC?

CMakeLists.txt
```diff
@@ -272,8 +272,9 @@ endif()
 option(ENABLE_GNUTLS "Enable protocol encryption and advanced authentication" ON)
 if(ENABLE_GNUTLS)
   find_package(GnuTLS)
-  if (GNUTLS_FOUND)
-    include_directories(${GNUTLS_INCLUDE_DIR})
+  find_package(Nettle)
+  if (GNUTLS_FOUND AND NETTLE_FOUND)
+    include_directories(${GNUTLS_INCLUDE_DIR} ${NETTLE_INCLUDE_DIR})
```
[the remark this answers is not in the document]
CendioOssman: I think it is best to keep them separate for clarity. And in case someone wants to avoid a GnuTLS dependency, for some odd reason.

common/rdr/AESInStream.cxx
```diff
@@ -0,0 +1,106 @@
...
+  end += maxSize - n;
+  return true;
+}
+
+bool AESInStream::readMessage()
+{
+  if (!in->hasData(2))
+    return false;
+  const U8* ptr = in->getptr(2);
+  size_t length = ((int)ptr[0] << 8) | (int)ptr[1];
```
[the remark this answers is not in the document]
CendioOssman: Yeah, that complicates things. Is this the case in every place where this kind of low-level stuff is done?

common/rfb/CSecurityRSAAES.cxx
```diff
@@ -0,0 +1,444 @@
...
+  : CSecurity(cc), state(ReadPublicKey), secType(_secType),
+    clientKey{0}, clientPublicKey{0}, serverKey{0},
+    serverKeyN(NULL), serverKeyE(NULL),
+    clientKeyLength(RSAKeyLength), clientKeyN(NULL), clientKeyE(NULL),
+    rais(NULL), raos(NULL), rawis(NULL), rawos(NULL)
+{
+  rawis = cc->getInStream();
+  rawos = cc->getOutStream();
+  if (gnutls_global_init() != GNUTLS_E_SUCCESS)
+    throw AuthFailureException("gnutls_global_init failed");
```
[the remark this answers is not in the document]
CendioOssman: Hmm... I think it is excessive to require GnuTLS just to have a random function. Have you looked at the `RandomStream` class?

common/rfb/CSecurityRSAAES.cxx
CendioOssman: Ah. I missed that. Sorry.

CendioOssman (changes requested):
Still no time to go through all of it yet, but here's another round.

Nice work with the testing. That should cover the important point of interoperability. :)

cmake/Modules/FindNettle.cmake
```diff
@@ -0,0 +1,41 @@
+
+find_package(PkgConfig)
+pkg_check_modules(NETTLE QUIET nettle)
+pkg_check_modules(HOGWEED QUIET hogweed)
+pkg_check_modules(GMP QUIET gmp)
```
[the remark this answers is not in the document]
CendioOssman: A bit odd, but you are right. It is also what nettle recommends apparently.

common/rdr/AESInStream.cxx
```diff
@@ -0,0 +1,106 @@
...
+
+#ifdef HAVE_GNUTLS
+using namespace rdr;
+
+const size_t MaxMessageSize = 65536;
+
+AESInStream::AESInStream(InStream* _in, const U8* key, int _keySize)
+  : keySize(_keySize), in(_in), msgLength(0), msgOffset(0), counter{0}
+{
+  msg = new U8[MaxMessageSize];
```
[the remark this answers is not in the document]
CendioOssman: Sure. It's merged now as 0ae1557ed9a7f8b46b6d7e2feaf2fad4cc2c3b3d .

Your PR here exposed some more issues in the existing code. Do you want me to merge any of these:

 * 6bd09e8c38f5dbb3f9b473ce1725a4d15ea8d52e
 * f1bfbd3dd74169ef6f4d83f0b3513475d0fcdac2
 * 66da1b2adf7ca9f2e177830ec87377ed248cd105
 * 83031fc1d466d78481c069ae4a075de182a87694

Otherwise, I'll just clean up things once everything here is merged.

common/rdr/AESInStream.cxx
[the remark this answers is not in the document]
CendioOssman: You shouldn't be too afraid of restore points. They were designed to help out precisely with things like the credentials. I'll make a note there separately, though.

common/rfb/SSecurityRSAAES.cxx
```diff
@@ -0,0 +1,587 @@
...
+
+bool SSecurityRSAAES::readCredentials()
+{
+  if (!rais->hasData(1))
+    return false;
+  const rdr::U8* ptr = rais->getptr(1);
+  rdr::U8 lenUsername = ptr[0];
+  if (!rais->hasData(1 + lenUsername + 1))
+    return false;
+  ptr = rais->getptr(1 + lenUsername + 1);
```
CendioOssman, line 528: This is precisely the kind of thing that `hasDataOrRestore()` was designed to handle. So please use that so this code is consistent with other protocol code.

common/rfb/CSecurityRSAAES.cxx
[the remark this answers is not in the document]
CendioOssman: I think that can work as a start. But I think PEM files are just DER files with Base64 encoding, so it would hopefully not be too difficult to write a small conversion function that doesn't need GnuTLS.

vncviewer/OptionsDialog.cxx
```diff
@@ -654,6 +684,15 @@ void OptionsDialog::createSecurityPage(int tx, int ty, int tw, int th)
...
     ty += CHECK_HEIGHT + TIGHT_MARGIN;
 
+#ifdef HAVE_NETTLE
+    encRSAAESCheckbox = new Fl_Check_Button(LBLRIGHT(tx, ty,
+                                                   CHECK_MIN_WIDTH,
+                                                   CHECK_HEIGHT,
+                                                   _("RSA-AES")));
+    encRSAAESCheckbox->callback(handleRSAAES, this);
+    ty += CHECK_HEIGHT + TIGHT_MARGIN;
+#endif
```
CendioOssman, line 694: The height of the containing `Fl_Group` needs to be adjusted as well. It should also be dynamic based on `#ifdef`s.

CendioOssman (changes requested):
I think this looks mostly good. Nice work!

There are a bunch of minor tweaks that are needed, after which I can go ahead and merge this.

The man pages also need to be updated for vncviewer, x0vncserver and Xvnc.

common/rfb/CSecurityRSAAES.h
```diff
@@ -0,0 +1,88 @@
...
+namespace rfb {
+  class UserMsgBox;
+  class CSecurityRSAAES : public CSecurity {
+  public:
+    CSecurityRSAAES(CConnection* cc, rdr::U32 secType,
+                    int keySize, bool isAllEncrypted);
+    virtual ~CSecurityRSAAES();
+    virtual bool processMsg();
+    virtual int getType() const { return secType; }
+    virtual bool isSecure() const { return isAllEncrypted; }
```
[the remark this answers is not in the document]
CendioOssman: It would be nice if someone had audited it for non-obvious stuff as well, but I suppose I can live with sticking to just the SHA-256 version right now.

common/rfb/SSecurityRSAAES.cxx
```diff
@@ -0,0 +1,581 @@
...
+next:
+    continue;
+  }
+  return -1;
+}
+
+static bool loadPEM(rdr::U8* data, size_t size, const char *begin,
+                    const char *end, rdr::U8** der, size_t *derSize)
+{
+  ssize_t pos1 = findSubstr(data, size, begin);
```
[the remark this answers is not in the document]
CendioOssman: Fair enough. There is also `memmem()`, but it doesn't seem to be available on Windows.

CendioOssman (changes requested):
Nice! I think this is pretty much ready to be merged now. What I'm missing is:

 * Updates to man pages
 * Fix the build on macOS
 * Squash/split everything to a final commit history. This might end up as a single big commit, but if you can find some way to split things, then it would be nice.

CendioOssman (approved):
Looks great!

A Java version would be much appreciated, if you feel you have the time.

bphinz:
That would be awesome if you can

PR #1521, opened 2022-08-24
Ignore leading and trailing whitespace

CatherineTower:
This makes the address box a little bit more tolerant of insignificant typos

The impetus for this change came from a troubleshooting session with my boss where he wasn't able to articulate that there was a problem because it was nearly invisible


common/rfb/Hostname.h
```diff
@@ -31,6 +31,10 @@ namespace rfb {
     const char* hostEnd;
     const char* portStart;
 
+    // Trim leading whitespace
+    while(isspace(*hi))
+      hi++;
+
     if (hi == NULL)
       throw rdr::Exception("NULL host specified");
```
CendioOssman, line 39: This check needs to be first or your new code will crash.

CendioOssman (changes requested):
This looks like a good change. But I'm missing some things in the actual code. I don't see any trimming of the hostname if there is no port specified?

We also have some unit tests for this code. I'd very much like to see some new test cases there that checks that this behaviour is correct.

common/rfb/Hostname.h
CatherineTower: You're dead on, I apologize for missing that

CatherineTower:
> I don't see any trimming of the hostname if there is no port specified?

You're right, and it causes undesired behavior in that case. Thanks for catching that, I'll work on getting that added

> We also have some unit tests for this code. I'd very much like to see some new test cases there that checks that this behaviour is correct.

Absolutely, I'll see what I can do! I may ask for help if I have too much trouble figuring out the unit test system

CatherineTower:
Okay, I made the requested changes. Commit 0099bdc shows some behavior that may be undesirable, could I get you to take a look at that? I can change my approach if you'd like it not to work the way it does

common/rfb/Hostname.h
```diff
@@ -44,7 +60,7 @@ namespace rfb {
         throw rdr::Exception("unmatched [ in host");
 
       portStart = hostEnd + 1;
-      if (*portStart == '\0')
+      if (*portStart == '\0' || isAllSpace(portStart))
```
CendioOssman, line 63: Can't the first part of this `if` get removed now?

CendioOssman (changes requested):
Looks good. Thanks!

Please squash together everything to a single commit, though. You can use `git rebase -i ` to do that. Then you'll need to do a force push to update your branch as you've rewritten history.

CendioOssman:
> Okay, I made the requested changes. Commit [0099bdc](https://github.com/TigerVNC/tigervnc/commit/0099bdc8d01a10aa659f9868e8d28a243bdd798b) shows some behavior that may be undesirable, could I get you to take a look at that? I can change my approach if you'd like it not to work the way it does

I think it might be desirable to support this. The modern source of extra whitespace like this is for autocorrecting touch keyboards. And they are generally very eager to insert whitespace around punctation like `:`.

There should be very little risk of errors here, since whitespace is illegal in both hosts and ports.

CatherineTower:
All right, it is done! Merges cleanly onto `master` as well

CendioOssman (approved):
Great. Thanks again for your contribution!

PR #1526, opened 2022-09-03
Fix up deb packages, and add Ubuntu 22.04 packages

CendioOssman:


CendioOssman:
The 18.04 build currently fails because the base container is broken (it contains packages from "pending"). It should resolve itself once those packages hit "updates".

CendioOssman:
Complained a bit to the guys maintaining the docker image as well:

https://bugs.launchpad.net/cloud-images/+bug/1988753

Issue #1518, opened 2022-08-20
What if I don't care about systemd?


perpen:
I just found out about the disastrous https://github.com/TigerVNC/tigervnc/pull/838

Background: I have been using tigervnc to display X clients on the Plan 9 UI.

Every time an X11 client is started, I create a dedicated VNC server with geometry identical to the Plan 9 window, and I run a custom window manager from ~/.vnc/xstartup.
See https://github.com/perpen/lx/blob/4bbdae224e85c3bfc143d93ce0beaebbac2e08ba/lxsrv.c#L533

But then it was decided to support **one** use case only, and to integrate with the pam/systemd rubbish. As @FredericHamel commented on the pr page, why not add a layer for this but continue to support the old, flexible mechanisms??
Anyone remembers the old unix philosophy? Or the time when ppl tried to make simple, portable software?

For now I will probably just revert to an earlier version of tigervnc, but we are making the world a less hackable place, sad.

PR #1541, opened 2022-10-31
Reintroduce xstartup support alongside Xsession

FredericHamel:
This resolve de issue #1518 even though it was already closed.

This PR reintroduce the possibility to use  `xstartup` script instead of Xsession.
The default behavior is to use Xsession.

You can provide the a custom xstartup script with `-xstartup` or use the default on by specifying `-use-xstartup`.

# Starting a vnc server with Xsession on display :1
```
vncserver :1 
```

# Starting a vnc server on display :1 using xstartup
```
vncserver -use-xstartup :1
```

# Starting a vnc server in background
```
# Using xstartup
vncserver -bg -use-xstartup :1

# Using Xsession
vncserver -bg :2
```

This PR also allows listing running server and killing a running server. A server can be started in the background using the `-bg` option. This options works with Xsession and xstartup script.

I propose this change to allow more flexibility.

### DIFF
--- a/unix/vncserver/vncserver.in
+++ b/unix/vncserver/vncserver.in
@@ -270,7 +270,7 @@ if ($cookie eq "" && open(URANDOM, '<', '/dev/urandom')) {
   close(URANDOM);
 }
 if ($cookie eq "") {
-  srand(time+$$+unpack("L",&readFile("$vncUserDir/passwd"));
+  srand(time+$$+unpack("L",&readFile("$vncUserDir/passwd")));
   for (1..16) {
     $cookie .= sprintf("%02x", int(rand(256)) % 256);
   }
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm afraid this is not something we'll merge. This is largely just a revert of #838. As such, it doesn't solve the main reason why `xstartup` was removed and `vncserver` was hidden¹, which is that it way too often just doesn't work.

We want to avoid including something in the official TigerVNC packaging that results in lots of user bug reports that we simply cannot fix.

We're not against the use case of users easily starting new VNC sessions. But the old model no longer solves that use case good enough, so something new would be needed. And likely cannot use  `xstartup` anyway because the internet is filled with examples for it that no longer work.

¹ Your system must have some modified version of TigerVNC, as there shouldn't be a `vncserver` command available.

FredericHamel:
Thanks for your reply. I will move the script to a separate project that will depend on TigerVNC.
I will call it `Xvnc-cli,` this will give an interface to start TigerVNC using xstartup script.
It will also manage multiple instance of the vncserver.

Is there a way to solve the main reason why `xstartup` was removed without forcing Xsession?

PR #1566, opened 2022-12-17
Add client-side support for the ARD and MSLogonII security types.

pdlan:
This PR adds the ARD and MSLogonII security types for the viewer. The Nettle library is required for building.

### DIFF
--- a/common/rfb/Security.cxx
+++ b/common/rfb/Security.cxx
@@ -165,6 +165,7 @@ rdr::U32 rfb::secTypeNum(const char* name)
   if (strcasecmp(name, "SSPI") == 0)       return secTypeSSPI;
   if (strcasecmp(name, "SSPIne") == 0)     return secTypeSSPIne;
   if (strcasecmp(name, "VeNCrypt") == 0)   return secTypeVeNCrypt;
+  if (strcasecmp(name, "DH") == 0)         return secTypeDH;
 
   /* VeNCrypt subtypes */
   if (strcasecmp(name, "Plain") == 0)      return secTypePlain;
@@ -191,6 +192,7 @@ const char* rfb::secTypeName(rdr::U32 num)
   case secTypeSSPI:       return "SSPI";
   case secTypeSSPIne:     return "SSPIne";
   case secTypeVeNCrypt:   return "VeNCrypt";
+  case secTypeDH:         return "DH";
 
   /* VeNCrypt subtypes */
   case secTypePlain:      return "Plain";
--- a/common/rfb/Security.cxx
+++ b/common/rfb/Security.cxx
@@ -166,6 +166,7 @@ rdr::U32 rfb::secTypeNum(const char* name)
   if (strcasecmp(name, "SSPIne") == 0)     return secTypeSSPIne;
   if (strcasecmp(name, "VeNCrypt") == 0)   return secTypeVeNCrypt;
   if (strcasecmp(name, "DH") == 0)         return secTypeDH;
+  if (strcasecmp(name, "MSLogonII") == 0)  return secTypeMSLogonII;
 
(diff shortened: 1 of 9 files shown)


CendioOssman (changes requested):

I think it would be clearer if we split ARD and MSLogonII to separate classes, even if they might share some code. Do you think you could have a look at doing that?

Please let them be separate commits as well.

common/rfb/CSecurityARD.cxx
```diff
@@ -0,0 +1,150 @@
...
+  rdr::InStream* is = cc->getInStream();
+  if (!is->hasData(4))
+    return false;
+  is->setRestorePoint();
+  rdr::U16 gen = is->readU16();
+  keyLength = is->readU16();
+  if (keyLength < MinKeyLength)
+    throw AuthFailureException("DH key is too short");
+  if (keyLength > MaxKeyLength)
+    throw AuthFailureException("DH key is too long");
```
CendioOssman, line 92: Why the length checks? No such checks were added to the noVNC implementation of this.

common/rfb/CSecurityARD.h
```diff
@@ -0,0 +1,53 @@
...
+    virtual int getType() const { return secTypeARD; }
+    virtual bool isSecure() const { return false; }
+
+  private:
+    bool readKey();
+    void writeCredentials();
+
+    int keyLength;
+    mpz_t g, p, A, b, B, k;
+    rdr::RandomStream rs;
```
CendioOssman, line 49: I don't suppose we can move this in to `writeCredentials()`? Lowers the complexity of the object and the header.

common/rfb/CSecurityARD.h
```diff
@@ -0,0 +1,53 @@
...
+#endif
+
+#include <nettle/bignum.h>
+#include <rfb/CSecurity.h>
+#include <rfb/Security.h>
+#include <rfb/UserMsgBox.h>
+#include <rdr/RandomStream.h>
+
+namespace rfb {
+  class UserMsgBox;
```
CendioOssman, line 34: This isn't used here in the header as far as I can tell?

common/rfb/Security.cxx
```diff
@@ -162,6 +162,7 @@ rdr::U32 rfb::secTypeNum(const char* name)
   if (strcasecmp(name, "RA2ne") == 0)      return secTypeRA2ne;
   if (strcasecmp(name, "RA2_256") == 0)    return secTypeRA256;
   if (strcasecmp(name, "RA2ne_256") == 0)  return secTypeRAne256;
+  if (strcasecmp(name, "ARD") == 0)        return secTypeARD;
```
CendioOssman, line 165: Please use the same sorting as in `Security.h` so it's easier to find things.

common/rfb/Security.h
```diff
@@ -43,6 +43,8 @@ namespace rfb {
   const rdr::U8 secTypeTLS     = 18;
   const rdr::U8 secTypeVeNCrypt= 19;
 
+  const rdr::U8 secTypeARD     = 30;
```
CendioOssman: Please try to use the same naming as in rfbproto. So perhaps `secTypeDH`?

CendioOssman (changes requested):
Thanks. I think things generally look very nice. There are some small tweaks I'd like you to have a look at, though.

common/rfb/CSecurityARD.cxx
CendioOssman: Sounds reasonable.

common/rfb/CSecurityARD.h
CendioOssman: I think this got overlooked. :)

CendioOssman (approved):

PR #1574, opened 2023-01-04
 Enable more warnings by default

CendioOssman:
These are also useful warnings that tend to find real bugs, so let's enable them.

The downside is they require us to mark certain things so the compiler knows if they were on purpose or not.


CendioOssman:
Are you sure? It looks correct to me. The documentation you sent to rfbproto says that the server should send the subtype right after the hash. A `return` doesn't seem correct.

CendioOssman:
Ah, I see what you mean. I can have a look.

It seems these warnings are a good thing and help us find issues. :)

CendioOssman:
Sure.

PR #1587, opened 2023-01-24
Massive type cleanup

CendioOssman:
This is a big clean-up to get rid of all the custom types we have in the codebase. Using more standard types should make things easier for other developers.

It also includes a clean-up of the memory management around strings and byte buffers, as I noticed that it was a mess as I was cleaning up the string typing. I'm fairly sure there were several bugs that got squashed as part of this clean-up.

The diff size can be a bit daunting, but the big changes are the first two commits, which are fairly simple conceptually.


CendioOssman:
Also note that the first commit mostly just moves code around, so it's best viewed with a git that has been configured to show moves (`diff.colorMoved`).

CendioOssman:
This PR also removes the "password" classes. They made some attempt at clearing the password from memory to reduce risk of leaking it. However, they didn't use the correct method for it, and we have many other places where we copy the password around in buffers. In other words, the classes provided a false sense of security and hence aren't a real loss IMO.

If we want to add that hardening to TigerVNC, then we need to do a wider audit and probably redesign of some things.

CendioOssman:
I've tested this in various ways, and I haven't seen any breakage. Performance also appears to be identical. So I'll go ahead and merge this. :)

PR #1592, opened 2023-02-01
Add <url> and <releases> to the metainfo

acperkins:
Flathub requires `<url>` and `<releases>` tags in the AppData file.

https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Releases.html

The `<releases>` tag can be much more detailed and include descriptions, issues, artifacts, and much more. For now I've only included the minimum required: version and date.

I've included dates of all of the previous versions that were listed on Github, as the information was available, but it would be sufficient to only include the latest release for now.

### DIFF
--- a/vncviewer/org.tigervnc.vncviewer.metainfo.xml.in
+++ b/vncviewer/org.tigervnc.vncviewer.metainfo.xml.in
@@ -13,6 +13,8 @@
   <name>TigerVNC Viewer</name>
   <summary>Connect to VNC server and display remote desktop</summary>
   <content_rating type="oars-1.1"/>
+  <url type="homepage">https://tigervnc.org/</url>
+  <url type="bugtracker">https://github.com/TigerVNC/tigervnc/issues</url>
   <description>
     <p>
       Virtual Network Computing (VNC) is a remote display system that allows you to view and interact with a virtual desktop environment running on another computer on the network.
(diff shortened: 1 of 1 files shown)


CendioOssman:
Thank you for your additions.

We explicitly left `<releases>` out of the document, as we didn't want another thing to keep updated. :/

I couldn't find any reference to this being required by Flathub in their wiki. Do you have a reference for this?

acperkins:
> I couldn't find any reference to this being required by Flathub in their wiki. Do you have a reference for this?

The documentation is [minimal](https://github.com/flathub/flathub/wiki/App-Requirements#appstream) but mentions that the appdata file must pass the validator. The requirements for `<url>` and `<releases>` are highlighted when you run that against the metainfo file:

    $ flatpak run org.freedesktop.appstream-glib validate org.tigervnc.vncviewer.metainfo.xml.in
    org.tigervnc.vncviewer.metainfo.xml.in: FAILED:
    • tag-missing           : <release> required
    • tag-missing           : <url> is not present
    Validation of files failed
    $

CendioOssman:
Hmm... So it's a flatpak thing, not a flathub one. I dug around a bit more, and I found the spec suggests adding `<releases>`:

https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#qsr-app-contents-suggestions

That section also states that `<provides>` should be included, which the tool doesn't seem to complain about...

CendioOssman:
This seems to be a mess where Flathub doesn't document their requirements, and appstream-util tries to arbitrary follow. And no one seems to care about the specification:

https://github.com/hughsie/appstream-glib/issues/457

The metainfo passes on `validate-relax` though, which is the check level recommended by flatpak. Could you poke Flathub back and see why they require stricter checks than that?

acperkins:
I don't know anyone at Flathub that I could poke, I was just following the submission guidelines.

I've patched the required tags as part of the Flathub build instead and that's worked.

PR #1602, opened 2023-02-20
Vncsession create homedir

zpytela:


grulja:
I can confirm this fixes `vncsession` failing when `~/.vnc` doesn't exist. After I apply this patch I see no AVC anymore and `~/.vnc` is created.

CendioOssman:
Looks reasonable. But is this a regression somehow? Or did we manage to miss this in all testing up until now?

grulja:
> Looks reasonable. But is this a regression somehow? Or did we manage to miss this in all testing up until now?

It's not a regression, it was just not spotted before. I think the usual workflow is to run `vncpasswd` first, which creates `~/.vnc` without any issue and therefore this issue doesn't occur most of the time.

grulja:
This would be also a good candidate for 1.13.1 in case you are going to release it with the fix for the keyboard handling issue.

CendioOssman:
Sounds good. I've included it on the 1.13 branch.

PR #1611, opened 2023-03-20
option -lowspeedfixes, fix random key press repeat

m0r1k:
Hi All!

Please accept my pull request.

For many years, I and others have used TigerVNC for work and hobbies, but when the network connection speed is slow, it becomes terrible, because the OS software starts repeating keystrokes at random. When you try to fix it, by pressing backspace, instead of one backspace, you can end up with two, three, or much more. And your whole line in the terminal that you typed for 10 minutes will be destroyed :) I lost a lot of keyboards because I was angry about it :(

Yesterday, I made the decision to find out and fix it, so the problem was very trivial and I don't know why I waited so many years. When we have very slow network speeds and high latency, we have a lot of delay between key press and key release events, so it looks like to the software as the long pressed key and it triggers auto-repeat.

The fix is very simple: if the -lowspeedfixes option is enabled and we have a keypress event for a regular key, we immediately create a fake keyrelease event (we shouldn't do this for modifiers, otherwise they won't be in the pressed state when the regular key event comes)

before:

image

after:

image

Thank you! Regards, Roman (m0r1k)

### DIFF
--- a/unix/xserver/hw/vnc/Xvnc.man
+++ b/unix/xserver/hw/vnc/Xvnc.man
@@ -1,9 +1,9 @@
 .TH Xvnc 1 "" "TigerVNC" "Virtual Network Computing"
 .SH NAME
-Xvnc \- the X VNC server 
+Xvnc \- the X VNC server
 .SH SYNOPSIS
 .B Xvnc
-.RI [ options ] 
+.RI [ options ]
 .RI : display#
 .SH DESCRIPTION
 .B Xvnc
@@ -50,9 +50,17 @@ depth 16 is RGB565 and for depth 24 and 32 is RGB888.
 Listen on interface. By default Xvnc listens on all available interfaces.
 .
 .TP
-.B \-inetd 
+.B \-inetd
 This significantly changes Xvnc's behaviour so that it can be launched from
 inetd.  See the section below on usage with inetd.
+.
+.TP
+.B \-lowspeedfixes
+FC37, the problem occurs on low-speed connections, when the time between
+keydown and keyup events is long, the OS starts repeating the keystroke,
+so we need generate the fake keyup event
+for normal keys (not for alt, ctrl, shift) right after receiving
+a keydown event to avoid it
 .
 .TP
 .B \-help
--- a/unix/xserver/hw/vnc/vncExtInit.h
+++ b/unix/xserver/hw/vnc/vncExtInit.h
@@ -34,6 +34,9 @@ void vncAddExtension(void);
 
 int vncNotifyQueryConnect(void);
 
+// xvnc.c
+extern int vncLowSpeedFixes;
+
 // vncExtInit.cc
 extern void* vncFbptr[];
 extern int vncFbstride[];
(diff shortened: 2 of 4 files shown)


CendioOssman:
Key repeat should already be suppressed precisely to avoid this misbehaviour. So, there is a bug somewhere...

Could you share more details about your setup? How are things started? And what applications are you seeing the issue in?

m0r1k:
Hello Pierre!

please let me know if you need more information (the vnc server was started without the -lowspeedfixes option)

image

my debug

image

patch with debug (for the entered text: "morik"):

[tigervnc-rchechnev.patch.txt](https://github.com/TigerVNC/tigervnc/files/11030068/tigervnc-rchechnev.patch.txt)

```

/---------------------------------------------------\                                 /---------------------------------------------------------------\
| vnc client (vinagre-3.22)                         |   <---- openvpn over udp ---->  | vnc server :1 (tiger vnc 1.12.0)                              |
| (fc37, kde plasma 5.27.2,                         |         (1 Mbit and less)       | (fc36, kde plasma 5.27.1,                                     |
| xorg-x11-server-Xorg-1.20.14-18)                  |         ping  400-1000 ms       | xorg-x11-server-Xorg-1.20.14-18)                              |
|                                                   |                                 |  /-----------------------------------------------------\      |
|                                                   |                                 |  | vmware  Workstation 16 Pro                          |      |
|                                                   |                                 |  |                                                     |      |
|                                                   |                                 |  |   / --------------------------------------------\   |      |
|                                                   |                                 |  |   |   vmware guest  (fc37)                      |   |      |
|                                                   |                                 |  |   |                                             |   |      |
|                                                   |                                 |  |   |   fedora login:  mmmorrrrrrik               |   |      |
|                                                   |                                 |  |   |                                             |   |      |
|                                                   |                                 |  |   \---------------------------------------------/   |      |
|                                                   |                                 |  \-----------------------------------------------------/      |
\---------------------------------------------------/                                 \---------------------------------------------------------------/

```

----- vnc client ----- 
vinagre-3.22.0-26.fc37.x86_64)
Fedora release 37 (Thirty Seven)
Linux localhost.localdomain 6.1.7-200.fc37.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Jan 18 17:11:49 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

image

----- vnc server ----- 
tigervnc-1.12.0.tar.gz
Fedora release 36 (Thirty Six)
Linux work 6.1.13-222.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Feb 28 16:57:03 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux

image

image

imho, my guess - not all programs checks "event->device_event.key_repeat = TRUE;" and begin to auto-repeat of the pressed key by themselves

https://www.youtube.com/watch?v=K1VLaXoRRdk&ab_channel=UKFDubstep

m0r1k:
Hello Pierre!

> Do you see the problem in more standard applications? Like a terminal window, or a word processor?

I do/did not see. Moreover, I have checked the QEMU, the problem does not appear there, it seems me this is only affect vmware.

image

CendioOssman:
I'm afraid it sounds like some bad behaviour in VMware, then. That's unlikely we can provide a clean workaround for. The suggestion in this PR would be something that users would need to know about, and enable. And it causes other problems, specifically that it makes it impossible to hold down keys. So it has very limited usefulness.

I would suggest reporting the issue to VMware. The fact that it works in QEMU should provide a good case that VMware can improve things.

Issue #1441, opened 2022-03-07
Feature request: Allow passing a hard-coded hostname or IP to match TLS/X509 certificates against


exikyut:
**Is your feature request related to a problem? Please describe.**
TigerVNC shows a non-optional security warning when the certificate returned by a server doesn't match the host or IP being connected to.

https://github.com/TigerVNC/tigervnc/blob/7fe85cd4a7ce89b55662fd445b78d50001bad68a/common/rfb/CSecurityTLS.cxx#L361-L366

This can happen in scenarios where
- the destination might not have a DNS address and has a dynamic (or DHCP-set) IP
- a reverse connection is being made from a machine that obviously is not going to have address in the server's cert

**Describe the solution you'd like**
I'd like to be able to pass a commandline or config-file option to specify a comma-delimited set of free-form strings to be passed to `gnutls_x509_crt_check_hostname()` in sequence (succeeding on the first match), to be considered with equal precedence to the string representation of the host/IP being connected to.

Perhaps the host/IP can be tested/matched first; perhaps also the matching string can be reported in the log (I'm not sure at what debug level).

**Describe alternatives you've considered**
- In one dimension one might consider disabling the popup and relegating the warning to stdout or a log file. Something something "`curl -k` considered harmful" etc. No. :)
- It's possible only passing a single string will prove to be perfectly adequate. I'm trying to envisage obscure scenarios involving load-balancers or multiple worker/node machines in HPC that might return different certificates. Given that I'm the first person reporting this it's perhaps possible I'm completely overthinking the engineering here :)

**Additional context**
It's possible this is the only viable solution to handle the reverse connection edge-case when using X509Vnc (https://github.com/TigerVNC/tigervnc/issues/1440).

CendioOssman:
This is very non-standard so I'd rather not go this route. I prefer if we look at how browsers handle this as they have a lot more experience dealing with certificates. And I've never seen something like this for them.

Browsers don't have a reverse connection mode though, so if we narrow the coverage this might be more interesting. Basically a list of acceptable reverse connection machines.

CendioOssman:
Hmm... SNI is also an issue. We are supposed to provide a single name as part of the TLS handshake. So if we have a list then which one to use?

Perhaps not an issue in practice and we can just not provide any name at that point.

aussielunix:
Hitting this issue too.
Running Ubuntu 22.04 in AWS on EC2 with a private IP.
[...]

PR #1643, opened 2023-07-05
Improve TLS host exception handling

CendioOssman:
Various fixes to get our exception handling for bad certificates more in line with how browsers behave. Most prominently, it now remembers an exception for when the hostname doesn't match the certificate.


CendioOssman:
Fixes #1441.

PR #1645, opened 2023-07-10
Use compressLevel in ZRLEEncoder

adamhalim:
Currently, setting a custom compression level in vncviewer only affects the TightEncoder. This change makes the ZRLEEncoder respect a client's desired compressLevel as well.

Tested using x0vncserver, seems to work.

### DIFF
--- a/common/rfb/ZRLEEncoder.cxx
+++ b/common/rfb/ZRLEEncoder.cxx
@@ -28,15 +28,23 @@
 #include <rfb/SConnection.h>
 #include <rfb/ZRLEEncoder.h>
 #include <rfb/Configuration.h>
+#include <rfb/LogWriter.h>
 
 using namespace rfb;
 
-IntParameter zlibLevel("ZlibLevel","Zlib compression level",-1);
+static LogWriter vlog("ZRLEEncoder");
+
+IntParameter zlibLevel("ZlibLevel","[DEPRECATED] Zlib compression level",-1);
 
 ZRLEEncoder::ZRLEEncoder(SConnection* conn)
   : Encoder(conn, encodingZRLE, EncoderPlain, 127),
-  zos(0,zlibLevel), mos(129*1024)
+  zos(0, 2), mos(129*1024)
 {
+  if (zlibLevel != -1) {
+    vlog.info("Warning: The ZlibLevel option is deprecated and is "
+              "ignored by the server. The compression level can be set "
+              "by the client instead.");
+  }
   zos.setUnderlying(&mos);
 }
 
@@ -48,6 +56,11 @@ ZRLEEncoder::~ZRLEEncoder()
 bool ZRLEEncoder::isSupported()
 {
   return conn->client.supportsEncoding(encodingZRLE);
+}
+
+void ZRLEEncoder::setCompressLevel(int level)
+{
+  zos.setCompressionLevel(level);
 }
 
 void ZRLEEncoder::writeRect(const PixelBuffer* pb, const Palette& palette)
--- a/common/rfb/ZRLEEncoder.h
+++ b/common/rfb/ZRLEEncoder.h
@@ -32,6 +32,8 @@ namespace rfb {
 
     virtual bool isSupported();
 
+    virtual void setCompressLevel(int level);
+
     virtual void writeRect(const PixelBuffer* pb, const Palette& palette);
     virtual void writeSolidRect(int width, int height,
                                 const PixelFormat& pf,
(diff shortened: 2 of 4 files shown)


CendioOssman:
Thanks. There is the "ZlibLevel" setting that also needs to be marked deprecated in some way. The documentation needs to be updated at least, but ideally, there is some warning to users that use this setting.

adamhalim:
I've updated it to work in the same way the `fullScreenAllMonitors` parameter is marked deprecated in `vncviewer`.

common/rfb/ZRLEEncoder.cxx
```diff
@@ -31,7 +31,7 @@
 
 using namespace rfb;
 
-IntParameter zlibLevel("ZlibLevel","Zlib compression level",-1);
+IntParameter zlibLevel("ZlibLevel","[DEPRECATED] Zlib compression level",-1);
```
CendioOssman, line 34: We should also stop using it so we get a clear switch to using the normal compression setting.

CendioOssman (changes requested):

unix/x0vncserver/x0vncserver.cxx
```diff
@@ -113,6 +113,18 @@ static void CleanupSignalHandler(int /*sig*/)
...
+{
+  rfb::VoidParameter* param = Configuration::getParam("ZlibLevel");
+  if (param == NULL)
+    return;
+  
+  int compressLevel = *((rfb::IntParameter*) param);
+  if (compressLevel != -1) {
+    vlog.info("ZlibLevel is deprecated, let clients set the compression level instead.");
+  }
+}
```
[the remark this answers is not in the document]
CendioHalim: My mistake, completely forgot about Xvnc.

The encoders are only constructed once a client actually connects to the servers. I figured it would be nice to display a warning message as soon as the server starts instead of when a client connects, so I moved the logging to Configuration.cxx which parses arguments during startup in both x0vncserver and Xvnc.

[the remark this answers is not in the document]
adamhalim: My mistake, completely forgot about Xvnc.

The encoders are only constructed once a client actually connects to the servers. I figured it would be nice to display a warning message as soon as the server starts instead of when a client connects, so I moved the logging to Configuration.cxx which parses arguments during startup in both x0vncserver and Xvnc.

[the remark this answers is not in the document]
CendioOssman: Reasonable concern, but I don't think it is worth breaking modularity in `Configuration.cxx` for. I think this is a very rarely used option. So we can live with the fact that the warning isn't perfect to keep the code simple and clean.

If we get many confused users, then we can always revisit the issue again.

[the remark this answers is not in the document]
adamhalim: Fair enough, I've moved the warning to the ZRLEEncoder's constructor.

common/rfb/ZRLEEncoder.cxx
```diff
@@ -17,6 +17,7 @@
  * USA.
  */
 
+#include <cstdio>
```
CendioOssman, line 20: Sorry, I missed this before. This doesn't look like it is used?

common/rfb/ZRLEEncoder.cxx
```diff
@@ -28,15 +29,23 @@
...
 
 ZRLEEncoder::ZRLEEncoder(SConnection* conn)
   : Encoder(conn, encodingZRLE, EncoderPlain, 127),
-  zos(0,zlibLevel), mos(129*1024)
+  zos(0, 2), mos(129*1024)
 {
+  if (zlibLevel != -1) {
+    vlog.info("Warning: The ZlibLevel option is deprecated and is "
+               "ignored by the server. The compression level can be set "
+               "by the client instead.");
```
CendioOssman, line 47: The indentation looks like it is off here?

CendioOssman (changes requested):
Just one more thing. There was a header there that doesn't seem to be used. From an earlier iteration?

common/rfb/ZRLEEncoder.cxx
CendioHalim: You're correct, removed it now.

adamhalim: You're correct, removed it.

common/rfb/ZRLEEncoder.cxx
adamhalim: Fixed

CendioOssman (approved):

PR #1646, opened 2023-07-14
Update colors and widget appearance for UI

CendioOssman:
Another part broken out of #1410. This time it is updating the colors to match modern systems better, and a new look for the widgets (as much as FLTK easily allows).

This is mainly trying to mimic Windows Fluent, but that is also close to macOS styling. It doesn't follow GNOME's adwaita (old or new), but UI look is less standardised on Linux so it should be acceptable there as well.

Before:

Screenshot from 2023-07-14 14-41-27

After:

Screenshot from 2023-07-14 14-41-19

### DIFF
--- a/vncviewer/ServerDialog.cxx
+++ b/vncviewer/ServerDialog.cxx
@@ -66,6 +66,8 @@ ServerDialog::ServerDialog()
 
   serverName = new Fl_Input_Choice(LBLLEFT(x, y, w() - OUTER_MARGIN*2,
                                    INPUT_HEIGHT, _("VNC server:")));
+  // Bug fix for wrong background
+  serverName->color(FL_BACKGROUND2_COLOR);
   y += INPUT_HEIGHT + INNER_MARGIN;
 
   x2 = x;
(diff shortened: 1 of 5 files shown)


samhed:
Looks nice when I build things myself on Fedora 38, but the RHEL9 nightly (from http://tigervnc.bphinz.com/nightly/) looks wrong:

image

CendioOssman:
Huh. Very odd. I can reproduce it by rebuilding the RPM packages on a RHEL 9 machine here. It doesn't happen if I build TigerVNC manually on the same machine though.

CendioOssman:
After some more digging, the triggering condition is enabling PIC/PIE, which Red Hat does by default for RPM builds. Unclear why this breaks things, though.

CendioOssman:
Found it. Box types cannot be larger than 255, and unfortunately FLTK doesn't protect against this. So we were doing out-of-bounds access. PIC/PIE changes the memory layout, which fortunately was enough to make the issue visible.

PR #1655, opened 2023-08-12
Make NSHighResolutionCapable a Boolean in Info.plist.in

ryandesign:
`NSHighResolutionCapable` should be a Boolean, not a string.


CendioOssman:
Thanks. Seems like people are very divided on having a string or boolean here. But the official docs say boolean, so let's follow that. :)

Did you encounter any practical issue with this incorrectly being a string?

PR #1659, opened 2023-08-22
update keysymdef.h to latest version from xorgproto.

### LINKED ISSUE #93: Adding unknown keysym until keysym table is full and no further input possible

hjicks:
also partially fixes #93, as most(?) keyboard symbols aren't "unknown" anymore.

tested with OpenBSD 7.3.

### DIFF
--- a/common/rfb/keysymdef.h
+++ b/common/rfb/keysymdef.h
@@ -1,9 +1,11 @@
-/* $TOG: keysymdef.h /main/28 1998/05/22 16:18:01 kaleb $ */
-
 /***********************************************************
 Copyright 1987, 1994, 1998  The Open Group
 
-All Rights Reserved.
+Permission to use, copy, modify, distribute, and sell this software and its
+documentation for any purpose is hereby granted without fee, provided that
+the above copyright notice appear in all copies and that both that
+copyright notice and this permission notice appear in supporting
+documentation.
 
 The above copyright notice and this permission notice shall be included
 in all copies or substantial portions of the Software.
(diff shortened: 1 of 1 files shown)


CendioOssman:
I'm afraid the keysym constants has nothing to do with issue #93.

In fact, the constants aren't really relevant unless there is some code referring to them. Which means that there doesn't seem to be a need for updating this list right now.

Was there a specific issue you were trying to address by updating this file?

hjicks:
Yes. old `keysymdef.h` doesn't map to unicode correctly, while some characters do, many don't, which from what i can understand causes some keys to go unknown as they don't match any keysym in X11, while new one does match unicode and i can see in patched version there is no `Input:       Added unknown keysym X to keycode Y` in logs anymore, nor weird behiavor reported in #93, at least for `ir` (persian) kbmap.

tested with  turbovnc's and `vncv` (plan 9) clients.

CendioOssman:
I'm afraid that file is not involved in that process. The message you are referring to means that the keysym cannot be found in the active keyboard layout, which is a different thing.

That means there must be some other change in your setup that improved things, not this. We can't really help out until we know what that change is. I suggest comparing the output of `xmodmap -pke` between tests to see the differences in the loaded keyboard layout.

hjicks:
I see.
I've checked and output of `xmodmap -pke` did differ in these two.

by default lots of space for keycodes is used by keyscodes which i don't use, like `XF86WWAN`, `XF86Next_VMode`, `XF86MailForward` etc.

i've managed to get rid of them by appending `-keycodes xfree86` to my `setxbmap` arguments.

tested for a while. seems okay. :)
thanks for help.

PR #1671, opened 2023-09-10
Support the "%u" username alias in PlainUsers

casantos:
This permits to enable PAM for the effective user of the Xvnc process by adding this to ~/.vnc/config or /etc/tigervnc/vncserver-config-defaults:

   SecurityTypes=TLSPlain
   PlainUsers=%u

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2233204

### DIFF
--- a/common/rfb/SSecurityPlain.cxx
+++ b/common/rfb/SSecurityPlain.cxx
@@ -28,6 +28,8 @@
 #include <rdr/InStream.h>
 #if !defined(WIN32) && !defined(__APPLE__)
 #include <rfb/UnixPasswordValidator.h>
+#include <unistd.h>
+#include <pwd.h>
 #endif
 #ifdef WIN32
 #include <rfb/WinPasswdValidator.h>
@@ -53,6 +55,13 @@ bool PasswordValidator::validUser(const char* username)
   for (size_t i = 0; i < users.size(); i++) {
     if (users[i] == "*")
       return true;
+#if !defined(WIN32) && !defined(__APPLE__)
+    if (users[i] == "%u") {
+      struct passwd *pw = getpwnam(username);
+      if (pw && pw->pw_uid == getuid())
+        return true;
+    }
+#endif
     if (users[i] == username)
       return true;
   }
--- a/unix/x0vncserver/x0vncserver.man
+++ b/unix/x0vncserver/x0vncserver.man
@@ -125,8 +125,9 @@ parameter instead.
 .B \-PlainUsers \fIuser-list\fP
 A comma separated list of user names that are allowed to authenticate via
 any of the "Plain" security types (Plain, TLSPlain, etc.). Specify \fB*\fP
-to allow any user to authenticate using this security type. Default is to
-deny all users.
+to allow any user to authenticate using this security type. Specify \fB%u\fP
+to allow the effective user of the server process. Default is to deny all
+users.
 .
 .TP
 .B \-pam_service \fIname\fP, \-PAMService \fIname\fP
--- a/unix/xserver/hw/vnc/Xvnc.man
+++ b/unix/xserver/hw/vnc/Xvnc.man
@@ -200,8 +200,9 @@ parameter instead.
 .B \-PlainUsers \fIuser-list\fP
 A comma separated list of user names that are allowed to authenticate via
 any of the "Plain" security types (Plain, TLSPlain, etc.). Specify \fB*\fP
-to allow any user to authenticate using this security type. Default is to
-deny all users.
+to allow any user to authenticate using this security type. Specify \fB%u\fP
+to allow the effective user of the server process. Default is to deny all
+users.
 .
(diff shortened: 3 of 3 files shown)


CendioOssman (approved):
This looks nice and clean! Thanks!

I wonder if there is a more descriptive marker we can use, though. I understand why systemd uses a single character, as it behaves more like a format string there. But in our case, it's a complete entry.

Did you stumble upon any more examples of configuration that has a marker like this? And wasn't using `$USER`. :)

`%u` probably works if we can't find anything better. But this is something we can't easily change later, so I think it is worth looking a bit extra at what options we have here.

CendioOssman:
Dovecot also uses `%u` for the current user:

https://doc.dovecot.org/configuration_manual/config_file/config_variables/#variables-mail-user

It also has longer names in the notation `%{user}`. That syntax is also used by Apache's mod_rewrite:

https://httpd.apache.org/docs/2.4/mod/mod_rewrite.html

`.desktop` files unfortunately use `%u` for URLs:

https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

Not that I think that is a blocker.

casantos:
My first draft recognized`$USER` and `$LOGNAME` but I switched to `%u` to avoid confusing the parameter value with an environment variable.

I think `%(user}` is unnecessarily verbose. BASH uses `\u` for the prompt definitions but I don't consider `\` better than `%`.

CendioOssman:
We don't seem to find any better alternatives, so let's proceed with `%u`.

common/rfb/SSecurityPlain.cxx
```diff
@@ -53,6 +55,13 @@ bool PasswordValidator::validUser(const char* username)
   for (size_t i = 0; i < users.size(); i++) {
     if (users[i] == "*")
       return true;
+#if !defined(WIN32) && !defined(__APPLE__)
```
CendioOssman, line 58: I don't suppose you checked if we can get Windows handling as well?

common/rfb/SSecurityPlain.cxx
```diff
@@ -53,6 +55,13 @@ bool PasswordValidator::validUser(const char* username)
...
+    if (users[i] == "%u") {
+      struct passwd *pw = getpwuid(geteuid());
```
CendioOssman, line 60: Why effective uid rather than real uid? I would assume most other things use the real uid, so this difference might be surprising.

common/rfb/SSecurityPlain.cxx
```diff
@@ -53,6 +55,13 @@ bool PasswordValidator::validUser(const char* username)
...
+      if (pw && !strcmp(pw->pw_name, username))
```
CendioOssman, line 61: This is technically not the correct way to do things, but I guess it's an existing problem. The system should be comparing uid:s, not strings, as there might be multiple valid usernames for the same user.

unix/x0vncserver/x0vncserver.man
```diff
@@ -125,8 +125,9 @@ parameter instead.
 .B \-PlainUsers \fIuser-list\fP
 A comma separated list of user names that are allowed to authenticate via
 any of the "Plain" security types (Plain, TLSPlain, etc.). Specify \fB*\fP
-to allow any user to authenticate using this security type. Default is to
-deny all users.
+to allow any user to authenticate using this security type. Specify \fB%u\fP
+to allow the effective user of the server process. Default is to deny all
+users.
```
CendioOssman, line 130: Great that you also remembered the documentation!

CendioOssman (changes requested):

common/rfb/SSecurityPlain.cxx
casantos: I didn't build tigervnc for windows. Also as far as I can tell the only authentication methods supported on Windows are "None" and "Standard VNC". I didn't find any way to enable "Plain" authentication.

common/rfb/SSecurityPlain.cxx
casantos: Oh yes, that was a mistake. It must be getuid or else it will be necessary to log in as root when Xvnc is installed with the set-user-id bit on.

common/rfb/SSecurityPlain.cxx
casantos: I have a better solution for this:
```
    if (users[i] == "%u") {
      struct passwd *pw = getpwnam(username);
      if (pw && pw->pw_uid == getuid())
        return true;
    }
```

CendioOssman (approved):

unix/x0vncserver/x0vncserver.man
casantos: I think we need to remove "effective" from the description, since we are using `getuid()`, not `geteuid()`. I will submit a follow-up PR.

PR #1692, opened 2023-11-03
fix spelling errors

wangqi123a:
fix   preceding


samhed:
Thanks.

I removed the whitespace changes in the commit. And instead of pushing that modification back to your branch, I accidentally pushed TigerVNC/master to wangqi123a/master.

Anyway, committed your fix in 2d0d516dac7a00bdcaf1f020009b38b367fe214b

PR #1698, opened 2023-11-22
C compatibility fixes

fweimer-rh:
Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/unix/xserver/hw/vnc/vncModule.c
+++ b/unix/xserver/hw/vnc/vncModule.c
@@ -36,7 +36,7 @@
 #include "XorgGlue.h"
 #include "RandrGlue.h"
 
-static void vncModuleInit(INITARGS);
+static void vncModuleInit(void);
 
 static MODULESETUPPROTO(vncSetup);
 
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindIconv.cmake
```diff
@@ -36,7 +36,7 @@ IF(ICONV_FOUND)
   #include <iconv.h>
   int main(){
     iconv_t conv = 0;
-    const char* in = 0;
+    char* in = 0;
```
CendioOssman, line 39: This looks wrong. The entire point of this code is to check if having that variable as `const` is valid or not.

unix/xserver/hw/vnc/vncModule.c
```diff
@@ -36,7 +36,7 @@
 #include "XorgGlue.h"
 #include "RandrGlue.h"
 
-static void vncModuleInit(INITARGS);
+static void vncModuleInit(void);
```
CendioOssman, line 39: Good catch. This looks like it should have been included in e4ec992984983ac9107efae1f1cc552007e4229e.

CendioOssman (changes requested):
Thanks for the fixes. One of them doesn't look quite right, though. Please have another look.

CendioOssman (approved):
Great. Thanks!

PR #1701, opened 2023-11-29
vncconfig: add option to force view-only remote client connections, round 2

casantos:
This is an alternative implementation of https://github.com/TigerVNC/tigervnc/pull/1670.Quite frankly I don't like it, since it's too intrusive and affects all SSecurity subclasses.


CendioOssman:
I agree, this is indeed a bit too invasive and not what I had in mind. The Security classes should not need to deal with this as this has nothing to do with them. The "view only" restriction should be merged with the security restrictions from the security types. So likely in `SConnection::processSecurityMsg()`, as that is where `setAccessRights()` is called.

But let's continue on the first PR to avoid confusion.

Issue #1691, opened 2023-10-30
Undefined references when building TigerVNC server


LMattsson:
**Describe the bug**
When doing an out-of-tree build of the TigerVNC server on RHEL 8 I get the following errors (`os::Mutex`, `rdr::Zlib...`, `rdr::RandomStream`):
<details>
  <summary>Error output</summary>

```
$ make TIGERVNC_SRCDIR=/home/cendio/tigervnc TIGERVNC_BUILDDIR=/home/cendio/tigervnc/build
...
/home/cendio/tigervnc/build/common/rfb/.libs/librfb.a(Configuration.cxx.o): In function `rfb::BinaryParameter::getValueStr[abi:cxx11]() const':
Configuration.cxx:(.text+0x142): undefined reference to `os::Mutex::lock()'
Configuration.cxx:(.text+0x15a): undefined reference to `os::Mutex::unlock()'
/home/cendio/tigervnc/build/common/rfb/.libs/librfb.a(Configuration.cxx.o): In function `rfb::StringParameter::getValueStr[abi:cxx11]() const':
Configuration.cxx:(.text+0x188): undefined reference to `os::Mutex::lock()'
Configuration.cxx:(.text+0x1ce): undefined reference to `os::Mutex::unlock()'
/home/cendio/tigervnc/build/common/rfb/.libs/librfb.a(Configuration.cxx.o): In function `rfb::VoidParameter::VoidParameter(char const*, char const*, rfb::ConfigurationObject)':
Configuration.cxx:(.text+0x77d): undefined reference to `os::Mutex::Mutex()'
/home/cendio/tigervnc/build/common/rfb/.libs/librfb.a(Configuration.cxx.o): In function `rfb::VoidParameter::~VoidParameter()':
Configuration.cxx:(.text+0x7cc): undefined reference to `os::Mutex::~Mutex()'
/home/cendio/tigervnc/build/common/rfb/.libs/librfb.a(Configuration.cxx.o): In function `rfb::StringParameter::setParam(char const*)':
Configuration.cxx:(.text+0xce2): undefined reference to `os::Mutex::lock()'
Configuration.cxx:(.text+0xd35): undefined reference to `os::Mutex::unlock()'
/home/cendio/tigervnc/build/common/rfb/.libs/librfb.a(Configuration.cxx.o): In function `rfb::BinaryParameter::setParam(unsigned char const*, unsigned long)':
[...]

PR #1704, opened 2023-12-11
Update building instructions

LMattsson:
When building the viewer and server on RHEL 9 we noted some requirements were missing. Also clarified some package details where needed.

### DIFF
--- a/BUILDING.txt
+++ b/BUILDING.txt
@@ -137,20 +137,17 @@ that the viewer has already been built, per above.)
   > ./configure --with-pic --without-dtrace --disable-static --disable-dri \
       --disable-xinerama --disable-xvfb --disable-xnest --disable-xorg \
       --disable-dmx --disable-xwin --disable-xephyr --disable-kdrive \
-      --disable-config-dbus --disable-config-hal --disable-config-udev \
-      --disable-dri2 --enable-install-libxf86config --enable-glx \
+      --disable-config-hal --disable-config-udev --disable-dri2 --enable-glx \
       --with-default-font-path="catalogue:/etc/X11/fontpath.d,built-ins" \
-      --with-fontdir=/usr/share/X11/fonts \
       --with-xkb-path=/usr/share/X11/xkb \
       --with-xkb-output=/var/lib/xkb \
       --with-xkb-bin-directory=/usr/bin \
-      --with-serverconfig-path=/usr/lib[64]/xorg \
-      --with-dri-driver-path=/usr/lib[64]/dri \
+      --with-serverconfig-path=/usr/lib64/xorg \
       {additional configure options}
     (NOTE: This is merely an example that works with Red Hat Enterprise/CentOS
-    6 and recent Fedora releases.  You should customize it for your particular
-    system.  In particular, it will be necessary to customize the font, XKB,
-    and DRI directories.)
+    9 and recent Fedora releases.  You should customize it for your particular
+    system.  In particular, it will be necessary to customize the XKB
+    directory.)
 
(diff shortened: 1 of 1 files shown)


BUILDING.txt
```diff
@@ -21,6 +23,7 @@ Build Requirements (All Systems)
 
 -- If building RSA-AES support:
    * Nettle 3.0 or later
+   * gmp
```
CendioOssman, line 28: Isn't this just a dependency of nettle? I'm not too fond of duplicating the dependency list recursively for everything we need, as that will not be maintained properly. :/

BUILDING.txt
```diff
@@ -36,12 +39,18 @@ Build Requirements (Unix)
 =========================
 
 -- Non-Mac platforms:
-   * X11 development kit
+  Development kit for the following:
+   * X11
+   * Xtst
+   * Xfixes
+   * Xrandr
+   * Xdamage
```
CendioOssman, line 49: Being this explicit worries me. This list tends to change, so I doubt we'll pay attention to updating this document. The existing message was meant to mean "all standard X11 libraries".

And some are optional as well, right?

BUILDING.txt
```diff
@@ -143,20 +143,17 @@ that the viewer has already been built, per above.)
...
-      --disable-config-dbus --disable-config-hal --disable-config-udev \
-      --disable-dri2 --enable-install-libxf86config --enable-glx \
+      --disable-config-hal --disable-config-udev --disable-dri2 --enable-glx \
       --with-default-font-path="catalogue:/etc/X11/fontpath.d,built-ins" \
-      --with-fontdir=/usr/share/X11/fonts \
       --with-xkb-path=/usr/share/X11/xkb \
       --with-xkb-output=/var/lib/xkb \
       --with-xkb-bin-directory=/usr/bin \
-      --with-serverconfig-path=/usr/lib[64]/xorg \
-      --with-dri-driver-path=/usr/lib[64]/dri \
```
CendioOssman, line 142: Removing this seems wrong. It needs to match your build of mesa.

CendioOssman (changes requested):
It's nice to get this updated for modern systems, but there are some details that concern me.

CendioOssman:
#1691 stated that you were building on RHEL 8, not 9. Did you change systems before doing this PR?

BUILDING.txt
LMattsson: Seems like you are correct, will remove.

BUILDING.txt
LMattsson: We removed them since these flags no longer exists for configure on RHEL 9.

BUILDING.txt
LMattsson: I agree that using "all standard X11 libraries" would be a neat way to express this, but it is not very helpful when unfamiliar with X11. If not mentioning explicit packages, could we mention higher level themes or similar to give a better starting point?

I have not looked into if the packages are optional or not.

LMattsson:
Yes, we swapped systems to RHEL 9.

BUILDING.txt
LMattsson: Starting from xorg v. 1.10, the dri is found automatically. Since TigerVNC requires xorg v. 1.16, these paths can safely be removed.

BUILDING.txt
LMattsson: I decided to go along the line of "all standard X11 libraries". Since the additional packages likely were a quite incomplete list I didn't include those.

CendioOssman (approved):

PR #1711, opened 2023-12-29
 Added the abillity for vnc to use 9 mouse buttons instead of 7.

manny33:
This adds the Xinput device buttons back and forward for the mouse.

Both Client and server message reader and writters as well as connections signal their support for the extra mouse buttons. The client signals the server through support of a psuedo encoding "pseudoEncodingExtendedMouseButtons". The server responds with a server message "msgTypeExtendedMouseSupport". Not sure if this is the best way to do this. After confirming both both client and server support the extended mouse button msg, the client starts sending new message with a u16 write instead of a u8. The button layout doesn't change to maintain support for clients that don't support more mouse buttons.

### DIFF
--- a/common/rfb/CConnection.cxx
+++ b/common/rfb/CConnection.cxx
@@ -860,5 +860,7 @@ void CConnection::updateEncodings()
   if (qualityLevel >= 0 && qualityLevel <= 9)
       encodings.push_back(pseudoEncodingQualityLevel0 + qualityLevel);
 
+  encodings.push_back(pseudoEncodingExtendedMouseButtons);
+
   writer()->writeSetEncodings(encodings);
 }
--- a/common/rfb/CMsgHandler.h
+++ b/common/rfb/CMsgHandler.h
@@ -58,6 +58,7 @@ namespace rfb {
     virtual void fence(uint32_t flags, unsigned len, const uint8_t data[]);
     virtual void endOfContinuousUpdates();
     virtual void supportsQEMUKeyEvent();
+    virtual void supportExtendedMouseButtons();
     virtual void serverInit(int width, int height,
                             const PixelFormat& pf,
                             const char* name) = 0;
--- a/common/rfb/CMsgWriter.cxx
+++ b/common/rfb/CMsgWriter.cxx
@@ -182,28 +182,21 @@ void CMsgWriter::writePointerEvent(const Point& pos, int buttonMask)
   if (p.x >= server->width()) p.x = server->width() - 1;
   if (p.y >= server->height()) p.y = server->height() - 1;
 
-  if (server->supportsExtendedMouseButtons)
-  {
-    writePointerEventExt(pos,buttonMask);
-    return;
+  if (server->supportsExtendedMouseButtons) {
+    startMsg(msgTypePointerEventExt);
+    os->writeU16(buttonMask);
+    os->writeU16(p.x);
+    os->writeU16(p.y);
+    endMsg();
   }
-
-  startMsg(msgTypePointerEvent);
-  os->writeU8(buttonMask);
-  os->writeU16(p.x);
-  os->writeU16(p.y);
-  endMsg();
+  else {
+    startMsg(msgTypePointerEvent);
+    os->writeU8(buttonMask);
+    os->writeU16(p.x);
+    os->writeU16(p.y);
+    endMsg();
+  }
 }
-
-void CMsgWriter::writePointerEventExt(const Point& p, int buttonMask)
-{
-  startMsg(msgTypePointerEventExt);
-  os->writeU16(buttonMask);
-  os->writeU16(p.x);
-  os->writeU16(p.y);
-  endMsg();
-}
-
 
 void CMsgWriter::writeClientCutText(const char* str)
 {
--- a/common/rfb/CMsgWriter.cxx
+++ b/common/rfb/CMsgWriter.cxx
@@ -182,8 +182,23 @@ void CMsgWriter::writePointerEvent(const Point& pos, int buttonMask)
   if (p.x >= server->width()) p.x = server->width() - 1;
   if (p.y >= server->height()) p.y = server->height() - 1;
 
+  if (server->supportsExtendedMouseButtons)
+  {
+    writePointerEventExt(pos,buttonMask);
+    return;
+  }
+
   startMsg(msgTypePointerEvent);
   os->writeU8(buttonMask);
+  os->writeU16(p.x);
+  os->writeU16(p.y);
+  endMsg();
+}
+
+void CMsgWriter::writePointerEventExt(const Point& p, int buttonMask)
+{
+  startMsg(msgTypePointerEventExt);
+  os->writeU16(buttonMask);
   os->writeU16(p.x);
   os->writeU16(p.y);
   endMsg();
--- a/common/rfb/SMsgReader.cxx
+++ b/common/rfb/SMsgReader.cxx
@@ -100,7 +100,7 @@ bool SMsgReader::readMsg()
     ret = readPointerEvent();
     break;
   case msgTypePointerEventExt:
-    ret = readPointerEvenExt();
+    ret = readPointerEventExt();
     break;
   case msgTypeClientCutText:
     ret = readClientCutText();
@@ -284,7 +284,7 @@ bool SMsgReader::readPointerEvent()
   return true;
 }
 
-bool SMsgReader::readPointerEvenExt()
+bool SMsgReader::readPointerEventExt()
 {
   if (!is->hasData(2 + 2 + 2))
     return false;
--- a/common/rfb/SMsgReader.cxx
+++ b/common/rfb/SMsgReader.cxx
@@ -99,6 +99,9 @@ bool SMsgReader::readMsg()
   case msgTypePointerEvent:
     ret = readPointerEvent();
     break;
+  case msgTypePointerEventExt:
+    ret = readPointerEvenExt();
+    break;
   case msgTypeClientCutText:
     ret = readClientCutText();
     break;
@@ -281,6 +284,16 @@ bool SMsgReader::readPointerEvent()
   return true;
 }
 
+bool SMsgReader::readPointerEvenExt()
+{
+  if (!is->hasData(2 + 2 + 2))
+    return false;
+  int mask = is->readU16();
+  int x = is->readU16();
+  int y = is->readU16();
+  handler->pointerEvent(Point(x, y), mask);
+  return true;
+}
 
 bool SMsgReader::readClientCutText()
 {
--- a/common/rfb/encodings.h
+++ b/common/rfb/encodings.h
@@ -36,6 +36,7 @@ namespace rfb {
 
   const int pseudoEncodingXCursor = -240;
   const int pseudoEncodingCursor = -239;
+  const int pseudoEncodingExtendedMouseButtons = -241;
   const int pseudoEncodingDesktopSize = -223;
   const int pseudoEncodingLEDState = -261;
   const int pseudoEncodingExtendedDesktopSize = -308;
--- a/unix/xserver/hw/vnc/vncInput.c
+++ b/unix/xserver/hw/vnc/vncInput.c
@@ -50,7 +50,7 @@ extern const unsigned int code_map_qnum_to_xorgevdev_len;
 extern const unsigned short code_map_qnum_to_xorgkbd[];
 extern const unsigned int code_map_qnum_to_xorgkbd_len;
 
-#define BUTTONS 7
+#define BUTTONS 9
 
 DeviceIntPtr vncKeyboardDev;
 DeviceIntPtr vncPointerDev;
@@ -206,6 +206,9 @@ static int vncPointerProc(DeviceIntPtr pDevice, int onoff)
 		btn_labels[4] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_WHEEL_DOWN);
 		btn_labels[5] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_LEFT);
 		btn_labels[6] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_RIGHT);
+		btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_BACK);
+		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_FORWARD);
+
 
 		axes_labels[0] = XIGetKnownProperty(AXIS_LABEL_PROP_REL_X);
 		axes_labels[1] = XIGetKnownProperty(AXIS_LABEL_PROP_REL_Y);
--- a/unix/xserver/hw/vnc/vncInput.c
+++ b/unix/xserver/hw/vnc/vncInput.c
@@ -207,11 +207,9 @@ static int vncPointerProc(DeviceIntPtr pDevice, int onoff)
 		btn_labels[5] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_LEFT);
 		btn_labels[6] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_RIGHT);
 		btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_SIDE);
-   		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_EXTRA);
-    	btn_labels[9] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_FORWARD);
-    	btn_labels[10] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_BACK);
-
-
+		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_EXTRA);
+		btn_labels[9] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_FORWARD);
+		btn_labels[10] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_BACK);
 
 		axes_labels[0] = XIGetKnownProperty(AXIS_LABEL_PROP_REL_X);
 		axes_labels[1] = XIGetKnownProperty(AXIS_LABEL_PROP_REL_Y);
--- a/unix/xserver/hw/vnc/vncInput.c
+++ b/unix/xserver/hw/vnc/vncInput.c
@@ -50,7 +50,7 @@ extern const unsigned int code_map_qnum_to_xorgevdev_len;
 extern const unsigned short code_map_qnum_to_xorgkbd[];
 extern const unsigned int code_map_qnum_to_xorgkbd_len;
 
-#define BUTTONS 9
+#define BUTTONS 11
 
 DeviceIntPtr vncKeyboardDev;
 DeviceIntPtr vncPointerDev;
@@ -206,8 +206,11 @@ static int vncPointerProc(DeviceIntPtr pDevice, int onoff)
 		btn_labels[4] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_WHEEL_DOWN);
 		btn_labels[5] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_LEFT);
 		btn_labels[6] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_RIGHT);
-		btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_BACK);
-		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_FORWARD);
+		btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_SIDE);
+   		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_EXTRA);
+    	btn_labels[9] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_FORWARD);
+    	btn_labels[10] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_BACK);
+
 
 
 		axes_labels[0] = XIGetKnownProperty(AXIS_LABEL_PROP_REL_X);
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -606,10 +606,13 @@ int Viewport::handle(int event)
     if (Fl::event_button3())
       buttonMask |= 4;
 
-    if(Fl::event_buttons()&FL_BUTTON(4))
-        buttonMask |= 128;
-    if(Fl::event_buttons()&FL_BUTTON(5))
-        buttonMask |= 256;
+    //Fl::event_button is only good for FL_PUSH and FL_RELEASE
+    if(event == FL_PUSH) {
+      if (Fl::event_button() == 8) //Mouse Back
+        buttonMask |= 1024;
+      if (Fl::event_button() == 9) //Mouse Forward
+        buttonMask |= 512;
+    }
 
     if (event == FL_MOUSEWHEEL) {
       wheelMask = 0;
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -606,6 +606,11 @@ int Viewport::handle(int event)
     if (Fl::event_button3())
       buttonMask |= 4;
 
+    if(Fl::event_buttons()&FL_BUTTON(4))
+        buttonMask |= 128;
+    if(Fl::event_buttons()&FL_BUTTON(5))
+        buttonMask |= 256;
+
     if (event == FL_MOUSEWHEEL) {
       wheelMask = 0;
       if (Fl::event_dy() < 0)
--- a/common/rfb/CMsgHandler.cxx
+++ b/common/rfb/CMsgHandler.cxx
@@ -80,6 +80,11 @@ void CMsgHandler::endOfContinuousUpdates()
   server.supportsContinuousUpdates = true;
 }
 
+void CMsgHandler::supportExtendedMouseButtons()
+{
+  server.supportsExtendedMouseButtons = true;
+}
+
 void CMsgHandler::supportsQEMUKeyEvent()
 {
   server.supportsQEMUKeyEvent = true;
--- a/common/rfb/CMsgReader.cxx
+++ b/common/rfb/CMsgReader.cxx
@@ -118,9 +118,6 @@ bool CMsgReader::readMsg()
     case msgTypeEndOfContinuousUpdates:
       ret = readEndOfContinuousUpdates();
       break;
-    case msgTypeExtendedMouseSupport:
-      ret = readSupportExtendedMouseButton();
-      break;
     default:
       throw Exception("Unknown message type %d", currentMsgType);
     }
@@ -205,6 +202,10 @@ bool CMsgReader::readMsg()
       handler->supportsQEMUKeyEvent();
       ret = true;
       break;
+    case pseudoEncodingExtendedMouseButtons:
+      handler->supportExtendedMouseButtons();
+      ret = true;
+      break;  
     default:
       ret = readRect(dataRect, rectEncoding);
       break;
@@ -454,12 +455,6 @@ bool CMsgReader::readFence()
 bool CMsgReader::readEndOfContinuousUpdates()
 {
   handler->endOfContinuousUpdates();
-  return true;
-}
-
-bool CMsgReader::readSupportExtendedMouseButton()
-{
-  handler->supportExtendedMouseButtons();
   return true;
 }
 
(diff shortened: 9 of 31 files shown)


common/rfb/SMsgReader.cxx
```diff
@@ -281,6 +284,16 @@ bool SMsgReader::readPointerEvent()
   return true;
 }
 
+bool SMsgReader::readPointerEvenExt()
```
CendioOssman, line 287: A missing `t` here.

common/rfb/encodings.h
```diff
@@ -36,6 +36,7 @@ namespace rfb {
 
   const int pseudoEncodingXCursor = -240;
   const int pseudoEncodingCursor = -239;
+  const int pseudoEncodingExtendedMouseButtons = -241;
```
CendioOssman, line 39: These need to be allocated with IANA before we can merge things:

https://www.iana.org/form/protocol-assignment
https://www.iana.org/assignments/rfb/rfb.xml

I'd also like to see a PR for documenting this new extension:

https://github.com/rfbproto/rfbproto

unix/xserver/hw/vnc/vncInput.c
```diff
@@ -50,7 +50,7 @@ extern const unsigned int code_map_qnum_to_xorgevdev_len;
 extern const unsigned short code_map_qnum_to_xorgkbd[];
 extern const unsigned int code_map_qnum_to_xorgkbd_len;
 
-#define BUTTONS 7
+#define BUTTONS 9
```
CendioOssman, line 53: If we're extending things, we might as well try to support everything.

evdev (which is the reference) has 8 buttons defined currently. Adding the 4 scroll buttons puts it at 12 buttons.

If we look at the Xorg evdev driver, it allocates room for 16 buttons, plus the 4 scroll buttons. Which means that we need more than 16 bits. I'm not sure if there is support for any such mice, though?

Perhaps we should at least use the full 16 (12+4) buttons we've made room for?

unix/xserver/hw/vnc/vncInput.c
```diff
@@ -206,6 +206,9 @@ static int vncPointerProc(DeviceIntPtr pDevice, int onoff)
 		btn_labels[4] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_WHEEL_DOWN);
 		btn_labels[5] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_LEFT);
 		btn_labels[6] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_RIGHT);
+		btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_BACK);
+		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_FORWARD);
```
CendioOssman, line 210: Annoyingly enough, this is not what the evdev driver does. It marks these as "side" and "extra".

It's code is a bit annoying to decipher, but it matches what this code does:

https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/hw/xfree86/drivers/inputtest/xf86-input-inputtest.c?ref_type=heads#L272

We should do the same for compatibility.

vncviewer/Viewport.cxx
```diff
@@ -606,6 +606,11 @@ int Viewport::handle(int event)
     if (Fl::event_button3())
       buttonMask |= 4;
 
+    if(Fl::event_buttons()&FL_BUTTON(4))
+        buttonMask |= 128;
+    if(Fl::event_buttons()&FL_BUTTON(5))
+        buttonMask |= 256;
```
CendioOssman, line 612: Are you sure about these? FLTK follows X11's button numbering. So I would expect 7 and 8 here.

(this also means we should probably pass through every high button from FLTK, to support all extra buttons)

CendioOssman, line 612: Have you looked at Windows and macOS support? It seems FLTK doesn't natively support those, but we might be able to fetch what we need indirectly.

CendioOssman (changes requested):
This looks promising. Nice work!

There are some things we should have another look at, though. See my comments.

common/rfb/CMsgWriter.cxx
```diff
@@ -182,13 +182,28 @@ void CMsgWriter::writePointerEvent(const Point& pos, int buttonMask)
...
 }
 
+void CMsgWriter::writePointerEventExt(const Point& p, int buttonMask)
+{
+  startMsg(msgTypePointerEventExt);
+  os->writeU16(buttonMask);
+  os->writeU16(p.x);
+  os->writeU16(p.y);
+  endMsg();
+}
```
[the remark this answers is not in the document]
manny33: for sure. I had it in different methods to lessen merge conflicts since i was just running this locally. it makes sense to make the main method handle both cases if it's going to be in master.

common/rfb/encodings.h
manny33: Thanks, i had no idea what these numbers were haha.

unix/xserver/hw/vnc/vncInput.c
manny33: yeah, if we're extending functionality it would be nice to fully extend it.

vncviewer/Viewport.cxx
manny33: Yeah, i tested and use these values. I did find it odd that java and c had different button combos (the clients). I'll look into this again after extending all buttons.

manny33: I've been testing from WSL and an Arch laptop i have. I need to set up my build chain for windows. I don't have a mac though.

manny33:
Update on this.

I've done all the changes requested in the PR. The handshake happens with an empty rect to reduce the need to for a specific Msg identifier.  Xinput buttons are mapped to reflect xserver's implementation. Typos have been corrected.

The main issue now is that FLTK doesn't properly support more than 7 mice buttons. FL::event_buttons masks the event state with 0x7f000000. This drops bit 31, which coincidentally is the back mouse button in the internal state. You can get that bit with the unmasked state "Fl::event_state()&FL_BUTTON(8)". But i haven't found a way to find if the forward button is pressed. Either way, bit 31 being high for mouse back seems more of a happy accident than something we can depend on. I'd have to track down the internal state in fltk source. 

I possibly could use the native message pump to handle only the back and forward button. The downside is that it would have to be implemented for every platform.

CendioOssman:
It doesn't look like `Fl::e_state` is particularly useful here, unfortunately. It basically maps to the X11 state, which only tracks the first five buttons.

So I think we'll need to convert things to track mouse button state ourselves and look at `Fl::e_keysym` instead.

manny33:
That worked. Thanks! 

Fl::event_button() maps back to 8, forward to 9.  It seems to have the button order flipped versus xservers "init_button_labels"

``` 
//Fl::event_button is only good for FL_PUSH and FL_RELEASE
    if(event == FL_PUSH) {
      if (Fl::event_button() == 8) //Mouse Back
        buttonMask |= 1024;
      if (Fl::event_button() == 9) //Mouse Forward
        buttonMask |= 512;
    }
``` 

vs the extended vncInput.c;

``` 
btn_labels[0] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_LEFT);
btn_labels[1] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_MIDDLE);
btn_labels[2] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_RIGHT);
btn_labels[3] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_WHEEL_UP);
btn_labels[4] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_WHEEL_DOWN);
btn_labels[5] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_LEFT);
btn_labels[6] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_RIGHT);
btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_SIDE);
btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_EXTRA);
btn_labels[9] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_FORWARD);
btn_labels[10] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_BACK);
```

I'll start getting that rfb assignment and working on the PR documentations.

CendioOssman:
If you push your updated code, we can have a look and test here as well.

unix/xserver/hw/vnc/vncInput.c
```diff
@@ -206,8 +206,11 @@ static int vncPointerProc(DeviceIntPtr pDevice, int onoff)
 		btn_labels[4] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_WHEEL_DOWN);
 		btn_labels[5] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_LEFT);
 		btn_labels[6] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_RIGHT);
-		btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_BACK);
-		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_FORWARD);
+		btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_SIDE);
+   		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_EXTRA);
+    	btn_labels[9] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_FORWARD);
+    	btn_labels[10] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_BACK);
+
```
manny33, line 213: I'll have to fix the indenting here. Not sure what happened.

manny33:
New code pushed.

vncviewer/Viewport.cxx
```diff
@@ -606,6 +606,14 @@ int Viewport::handle(int event)
     if (Fl::event_button3())
       buttonMask |= 4;
 
+    //Fl::event_button is only good for FL_PUSH and FL_RELEASE
+    if(event == FL_PUSH) {
+      if (Fl::event_button() == 8) //Mouse Back
+        buttonMask |= 1024;
+      if (Fl::event_button() == 9) //Mouse Forward
+        buttonMask |= 512;
+    }
```
CendioOssman, line 615: This does not work for me here with Fedora 38. I get button 10 and 11 in the session, instead of the expected 8 and 9.

CendioOssman, line 615: I cannot hold down any of these buttons and move the mouse. It then immediately thinks I released the button.

You'll need to keep track of the button mask, unfortunately. FLTK will not provide a correct one for you.

unix/xserver/hw/vnc/vncInput.c
```diff
@@ -50,7 +50,7 @@ extern const unsigned int code_map_qnum_to_xorgevdev_len;
 extern const unsigned short code_map_qnum_to_xorgkbd[];
 extern const unsigned int code_map_qnum_to_xorgkbd_len;
 
-#define BUTTONS 7
+#define BUTTONS 11
```
CendioOssman, line 53: Let's use all the 16 buttons we have, so we can support devices with many buttons.

vncviewer/Viewport.cxx
manny33: I feared this would be the case. What would be the solution here, read a config file for the button mapping?

CendioOssman:
I got my hands on a Logitech G604, which has 6 buttons on the side. Unfortunately, this thing is a bit magical and generates fake keyboard keys instead of mouse buttons for the extra ones. So it ended up not being particularly useful here. :/

vncviewer/Viewport.cxx
CendioOssman: That should not be necessary. Applications such as Chrome and Firefox do not have a config file for this, so it shouldn't be necessary for us.

Where did you get the values 1024 and 512?

unix/xserver/hw/vnc/vncInput.c
CendioOssman: The standard driver just uses `BTN_LABEL_PROP_BTN_UNKNOWN` for the extra ones, so that should be good enough for us as well.

vncviewer/Viewport.cxx
manny33: the way the buttons are mapped from the input is a bit shift in vncPointerButtonAction. Back is the 10th button in the map so 1<<10.

CendioOssman: Do you mean it's the one labeled `BTN_LABEL_PROP_BTN_BACK`? Unfortunately, that label is wrong. You'll need to compare with the events you get locally, since that is what the applications expect. Things such as `xev` should look identical locally and in the VNC session.

And no, we don't want to fix the labels. We need to be bug-for-bug compatible with a local system to avoid breaking applications.

CendioOssman:
How are things progressing here, @manny33? It would be great to get this merged.

manny33:
@CendioOssman , pardon i got a bit busy with some holiday travel and work deliverables. I was planning on looking into this again come next week.

CendioOssman:
Have you had a chance to look further at this, @manny33?

CendioHalim:
@manny33 I was wondering if we really need `msgTypePointerEventExt`?

I figured since there is only support for 7 buttons right now, and we send the button mask using 8 bits, could we not use that last bit to indicate that the message contains 8 more bits? I haven't got around to check if this would work properly, do you think it could work?

If you'd like, I could send a proposal to IANA for the pseudo-encoding, since we need it in any case. The msgType can always be added at a later time.

manny33:
@CendioHalim ahh i see, UTF-8 like tricks there. It should work for tiger VNC, but i don't know what every server would do if they found the high bit set, they may have some assert on it and cause a crash.

manny33:
@CendioHalim I see. That would be a nice trick. Once we handshake on the pseudo encodings, the client and server can both agree what the high bit means. We can then decode the message properly on the server. I'm all for it. It would simplify the code.

CendioHalim:
@manny33

My idea is the following:

With the buttonMask we get on the client, we keep the lower 7 bits as they are, and take the upper 9 bits and shift them 1 step to the left. This frees the 8th bit to be used only to indicate the message type, and the upper 7 bits can be used for different buttons.

On the server side, we can decode the message in two ways:

1. We start by reading 8 bits and look at the MSB to detect message type. If the MSB is 1, we read 8 more bits, shift them 8 steps to the left, and combine it with the first 8 bits.  
2. We call `is->setRestorePoint()` before reading 16 bits directly and look at the 8th bit. If it is 1, we can go ahead and decode the message. If the 8th bit is 0, we call `is->gotoRestorePoint()` and read 8 bits as usual.

I haven't had the time to get this working fully yet, but will probably continue after the weekend. Do you think this is a sane approach? I think I'm more in favour of the 2nd alternative.

Here's a small snippet I'm playing with (client side only):
```cpp
void CMsgWriter::writePointerEvent(const Point& pos, int buttonMask)
{
  Point p(pos);
  if (p.x < 0) p.x = 0;
  if (p.y < 0) p.y = 0;
  if (p.x >= server->width()) p.x = server->width() - 1;
  if (p.y >= server->height()) p.y = server->height() - 1;

  startMsg(msgTypePointerEvent);
  if (server->supportsExtendedMouseButtons && buttonMask >= 1 << 7) {
    int higherBits;
    int lowerBits;
    int extendedButtonMask;

    // Clear the lower 7 bits, and set the 8th bit to indicate extended
    // button event. The upper 8 bits are used to store the extended
    // button mask and are shifted 1 bit to the left.
    higherBits = buttonMask & ~127;
    higherBits <<= 1;
    higherBits |= 1 << 7;

    // Keep the lower 7 bits
    lowerBits = buttonMask & 127;

    extendedButtonMask = higherBits | lowerBits;

    os->writeU16(extendedButtonMask);
  } else {
    os->writeU8(buttonMask);
  }
    os->writeU16(p.x);
    os->writeU16(p.y);
    endMsg();
}
```

It relies heavily on other VNC implementations not using the 8th bit at all, so we would have to have a look and test with different clients/servers. I don't think this is a good idea if it doesn't play well with other implementations.

manny33:
@CendioHalim 
I didn't know about the Set / goto restore pointer. Those are neat for the reader. I think the approach would work. We also don't need to shift the high bits back, as we can just declare bits 9 and 10 as the back / forward button on the server. 

I am worried about the 8th bit compatibility like you said. That bit lives in limbo, and relying on it may raise compatibility issues. As long as we don't set it when the client isn't expecting it, it should be fine. Compatibility issues may arise, but at that point I suppose we'd need to talk about standardizing that bit or the extended mouse button support.  

The new message id approach does make it a clean break though. I don't know enough about the VNC community to make a clear judgement on it. There could be considerations with other client and servers I'm not aware of.

CendioHalim:
@manny33 
My concern with not shifting the bits back is what to do with the 8th index in `btn_labels`. I think it's safer to add some complexity on the protocol side rather than doing something in `vncinput.c` if the 8th bit happened to be set.

I'm also not sure about the new message type and how this approach plats with other implementations, but I feel like if we can design the protocol to avoid the need for a new message type it's worth it.

My hope is to have a version of the protocol to test with this week.

manny33:
@CendioHalim that would be great. Thanks again for taking point with this. I'm not very experienced with writing these proposals as I'm just a hobbyist.

Neustradamus:
To follow this PR :)

CendioHalim:
The pseudo-encoding has been registered with [IANA](https://www.iana.org/assignments/rfb/rfb.xhtml) and the assignment is -316 (the previous value was already allocated).

While writing the draft for rfbproto, I realized that the implementation might be a bit overcomplicated? If we decide that the server and client completely switch to 16 bit buttonmasks, after both are sure that the other supports the encoding, we can completely skip the 8th bit marker?

I tested this (https://github.com/CendioHalim/tigervnc/commit/8d0afe9709604aa7a1fb241f0a8ba95a868d9b76) and it worked, combining new/old client/servers. The implementation is much simpler, but I wonder if there are some unforeseen consequences I've missed?

Another thing I'm thinking about is what to do with the remaining 7 bits? I had a look at how [QT](https://doc.qt.io/qt-6/qt.html#MouseButton-enum) and [GTK](https://docs.gtk.org/gdk3/struct.EventButton.html) handle mouse buttons, and there doesn't seem to be a clear consensus.

Here is my draft https://github.com/CendioHalim/rfbproto/commit/b00f2d8ccc70665aa23e94c1aecdf3c0c85d492b, I'll make a PR to rfbproto soon.

CendioHalim:
I also had a look at some game engines: [Godot](https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#enum-globalscope-mousebutton), [Unity](https://docs.unity3d.com/ScriptReference/KeyCode.html) and [Unreal](https://docs.unrealengine.com/4.26/en-US/BlueprintAPI/Input/MouseEvents/), and they all looked like they only supported the normal 7 buttons, plus the forward and back buttons.

CendioHalim:
> While writing the draft for rfbproto, I realized that the implementation might be a bit overcomplicated? If we decide that the server and client completely switch to 16 bit buttonmasks, after both are sure that the other supports the encoding, we can completely skip the 8th bit marker?

I realized that this creates a race condition, where the server will start accepting PointerEvents as having 2b buttonMasks as soon as `SMsgHandler::setEncdoings()` is called, before responding with `SMsgWriter::writeExtendedMouseButtonRect()` The client will still send 1b buttonMasks until the response rect is sent back, causing a crash.

I've reverted https://github.com/CendioHalim/tigervnc/commit/8d0afe9709604aa7a1fb241f0a8ba95a868d9b76, going back to the previous implementation. The rfbproto draft will be updated soon.

CendioHalim:
* rfbproto PR: https://github.com/rfbproto/rfbproto/pull/57

vncviewer/Viewport.cxx
dcommander: I had a chance to look into Java's behavior in the context of implementing horizontal scroll wheel events.  Java Buttons 4 through 7 are assigned differently depending on the O/S, because horizontal scroll wheel events are implemented differently in different O/S's.  On macOS and Windows, horizontal scroll wheel events are implemented as vertical scroll wheel events with the Shift key held down.  Since Java has a separate event subclass for scroll wheel events, it doesn't need to use Java Buttons 4 and 5 for scroll wheel events.  Thus, on macOS and Windows, Java Buttons 4 and 5 are used for the front and back mouse buttons.  On X11, however, horizontal scroll wheel events are implemented using X11 Buttons 6 and 7.  Java/X11 still uses a separate event subclass for vertical scroll wheel events, but it uses Java Buttons 4 and 5 for horizontal scroll wheel events, which means that it uses Java Buttons 6 and 7 for the front and back mouse buttons.

dcommander:
@CendioHalim IMHO a better approach would be to use a different message type, much like the QEMU Extended Key Event message does.  Other VNC server implementations that still use C rather than C++ (such as LibVNCServer or TurboVNC) use fixed structures for different message types, e.g.:

```
typedef struct _rfbPointerEventMsg {
    CARD8 type;                 /* always rfbPointerEvent */
    CARD8 buttonMask;           /* bits 0-7 are buttons 1-8, 0=up, 1=down */
    CARD16 x;
    CARD16 y;
} rfbPointerEventMsg;

#define sz_rfbPointerEventMsg 6
```

(That is what TightVNC and RealVNC v3 and prior did as well.)  Thus, the message can be read in one throw, rather than with the separate reads that TigerVNC and RealVNC v4 and later do.  It's not a deal-breaker.  It's more about being consistent with how RFB extensions were traditionally implemented, which would make the extension friendlier to other implementations that still conform that that tradition.

dcommander:
@CendioHalim Another possibility, if you didn't want to get a new message type assigned to the extension, would be to send the extended button info after the base RFB pointer message. That would allow more traditional VNC implementations to continue using the `rfbPointerEventMsg` struct in **rfbproto.h**. They could simply analyze the `buttonMask` field after the struct is read and determine whether they should read an additional byte.

dcommander:
@CendioHalim Also, if you send buttons 8-15 as a separate byte after the base message, then the meaning of Bits 0-6 in the base message does not change based on whether the extension is in use. Implementations would always take the pressed state of Buttons 1-7 from those bits. If the pseudo-encoding is enabled, then implementations would use Bit 7 to determine whether to read the additional byte, then they would take the pressed state of Buttons 8-15 from that additional byte. That seems like it would be easier to document as well.

CendioHalim:
> @CendioHalim IMHO a better approach would be to use a different message type, much like the QEMU Extended Key Event message does. 

I think it's better to avoid a new message type if it is possible, which it seems to be right now. They are a limited resource.

> @CendioHalim Another possibility, if you didn't want to get a new message type assigned to the extension, would be to send the extended button info after the base RFB pointer message. That would allow more traditional VNC implementations to continue using the `rfbPointerEventMsg` struct in **rfbproto.h**. They could simply analyze the `buttonMask` field after the struct is read and determine whether they should read an additional byte.

Sending the extra byte after the base message seems like a good thing to do, especially if, as you say, it would play better with other implementations.

> @CendioHalim Also, if you send buttons 8-15 as a separate byte after the base message, then the meaning of Bits 0-6 in the base message does not change based on whether the extension is in use. Implementations would always take the pressed state of Buttons 1-7 from those bits. If the pseudo-encoding is enabled, then implementations would use Bit 7 to determine whether to read the additional byte, then they would take the pressed state of Buttons 8-15 from that additional byte. That seems like it would be easier to document as well.

This is pretty much how the protocol already is designed as of now.

> I suppose that, if you wanted to future-proof it, you could have the extension use an additional unsigned short rather than an additional byte, which would allow it to support up to 23 total buttons.

I was thinking this as well, but I'm not sure what to do with the additional bytes. Future-proofing is not a bad idea, you could even let the server respond with how many additional bytes are supported in the response rect if you want to be super flexible. I'm just not sure what this means for the protocol, as we'd have bit 0-9 defined, and 10-31 reserved for future use?

dcommander:
> I think it's better to avoid a new message type if it is possible, which it seems to be right now. They are a limited resource.

Fair enough.

> > I suppose that, if you wanted to future-proof it, you could have the extension use an additional unsigned short rather than an additional byte, which would allow it to support up to 23 total buttons.
> 
> I was thinking this as well, but I'm not sure what to do with the additional bytes. Future-proofing is not a bad idea, you could even let the server respond with how many additional bytes are supported in the response rect if you want to be super flexible. I'm just not sure what this means for the protocol, as we'd have bit 0-9 defined, and 10-31 reserved for future use?

Referring to https://github.com/TigerVNC/tigervnc/pull/1711#discussion_r1438309621, there are potentially 16 mouse buttons that need to be handled, but the current implementation only handles 15.  It would be straightforward to use an unsigned short rather than a byte for Buttons 8-N, which would give us room for 16 additional buttons on top of the 7-8 that can already be handled by the base pointer event message.  However, the counterarguments are:

1. Using an unsigned short would increase the overall pointer event message size by 33% rather than 17%.
2. The message size increase would apply to any implementation that needs more than 7 buttons.
3. Few if any implementations would ever need more than 15 buttons.
4. Implementations that need more than 15 buttons could always use the GII extension to support an arbitrary number of buttons.  (The counter-counter argument is that GII is such a complex and fine-grained extension that, effectively, every implementation of it defines a new protocol that won't be compatible with other GII implementations unless you intentionally make it so, whereas this mouse button extension is dead simple and has no compatibility wiggle room.)
 
Specifying the number of additional bytes would require either sacrificing more of the mask bits in the base pointer event message or sending yet another byte after the base message to indicate the number of bytes that follow.  I may be paranoid to care so much about the message size, but since pointer events are sent so frequently and they are normally only 6 bytes in length, it seems like a good idea to avoid sending even one additional byte unless absolutely necessary.  I like https://github.com/CendioHalim/tigervnc/commit/275670c905525e01eb0a444896bfc4a1aa572155 as it is currently implemented (minus the one comment I made on it), because it allows up to 7 buttons (enough to accommodate both vertical and horizontal scroll wheel events) without increasing the message size.  The argument in favor of extending it to handle 16-23 buttons is weak, and the argument in favor of extending it to handle 24+ buttons is even weaker.

CendioHalim:
Regarding https://github.com/TigerVNC/tigervnc/pull/1711#discussion_r1438309621:

From what I've tested, some apps that support forward and back mouse buttons listen on both
`[BTN_LABEL_PROP_BTN_SIDE, BTN_LABEL_PROP_BTN_FORWARD]` for forward, and
`[BTN_LABEL_PROP_BTN_EXTRA, BTN_LABEL_PROP_BTN_BACK]` for back.

The label usage is inconsistent across X applications. For example, I tested a Logitech G400s, which has 7 buttons (left, middle, right, forward, back, and three extra).  Two of the extra buttons are to change the DPI and don't generate any button events. The third button, which is `"button 10" / BTN_LABEL_PROP_BTN_FORWARD` with `xinput --test <id>`, navigates forward in Firefox and Chrome, but does nothing in Nautilus file explorer.

I've tested three different mice (2 logitech, 1 microsoft) with forward/back buttons, and they all map SIDE/EXTRA to forward/back. I'm assuming most X11 applications use these labels for forward/back.

Since the protocol will only specify which bits correspond to a forward/back button, does it not mean that we have to pick which label we use for forward/back in Xvnc? Otherwise, we'd need additional bits for full support, and that would only work with X11 VNC clients that also support these labels? I would be OK with only using SIDE/EXTRA in Xvnc and only allocating 2 bits for these.

Regarding 1 vs 2 extra bytes, I don't really know what's best. In my mind, forward/back buttons will satisfy 99 % of users. At the same time, clients can choose to only send an extended mask when necessary, and the total data increase for a session would be negligible.

common/rfb/CConnection.cxx
```diff
@@ -860,5 +860,7 @@ void CConnection::updateEncodings()
   if (qualityLevel >= 0 && qualityLevel <= 9)
       encodings.push_back(pseudoEncodingQualityLevel0 + qualityLevel);
 
+  encodings.push_back(pseudoEncodingExtendedMouseButtons);
```
CendioOssman, line 863: Why not with all the other stuff further up?

common/rfb/CMsgHandler.h
```diff
@@ -58,6 +58,7 @@ namespace rfb {
     virtual void fence(uint32_t flags, unsigned len, const uint8_t data[]);
     virtual void endOfContinuousUpdates();
     virtual void supportsQEMUKeyEvent();
+    virtual void supportExtendedMouseButtons();
```
CendioOssman, line 61: support**s**

CendioOssman (changes requested):
I guess this is not updated to the latest specification?

rudironsonijr:
@mdevaev is it possible to implement this functionality in PiKVM's vnc server?

mdevaev:
@rudironsonijr PiKVM emulates only 8 buttons now. I can add them when the patch will bi applied to TigerVNC upstream.

CendioHalim:
I've created a new PR with continuation of the work done in this PR:
* https://github.com/TigerVNC/tigervnc/pull/1826

CendioOssman:
Okay. Let's continue the work there then.

PR #1714, opened 2024-01-08
Xvnc: do not use dup2() to create a file descriptor for -inetd option

### LINKED ISSUE #1709: Don't use ' dup2(0, 3)' unconditionally
When a system is configured to use 'libnss_sss.so.2' as NSS provider (i.e. /etc/nsswitch.conf contains "passwd: sss files ...") this code - 
https://github.com/TigerVNC/tigervnc/blob/effd854bfd19654fa67ff3d39514a91a246b8ae6/unix/xserver/hw/vnc/xvnc.c#L369
 \- ruins internal state of 'libnss_sss.so.2' 

The flow is as follows: 
 - after all the fork/exec/etc, Xvnc calls one of NSS functions (most probably `getpwuid(getuid))` but not necessarily);
 - this initializes NSS stack, and calls to 'libnss_sss' 
 - this initializes internal data structures of 'libnss_sss', in particular opens some file descriptors
 - one of file descriptors obtained by 'libnss_sss' is 3 (first free)
 - Xvnc "hijacks" this fd with `dup2(0, 3);`
 - Xvnc call NSS again, 'libnss_sss' tries to work with this fd, discovers there is "something wrong", closes it (fd=3), reinits (and gets fd=3 again)
 - Xvnc keeps working with fd=3, discovers there is "something wrong", terminates

strace looks like:
```
fstat(3, {st_mode=S_IFSOCK|0777, st_size=0, ...}) = 0
munmap(0x7f3bf9792000, 9253600)         = 0
close(3)                                = 0
openat(AT_FDCWD, "/var/lib/sss/mc/passwd", O_RDONLY|O_CLOEXEC) = 3
...
setsockopt(3, SOL_TCP, TCP_CORK, [1], 4) = -1 ENOTSOCK (Socket operation on non-socket)
```

And Xvnc log:
```
Xvnc[76330]: XserverDesktop: client gone, sock 3
Xvnc[76330]: Selection: Remote clipboard lost, removing local ownership
Xvnc[76330]: VNCSConnST: closing [::ffff:...]::56002: read: Socket operation on non-socket (88)
```

**To Reproduce**
Configure system:
 - to use Xvnc with '-inetd'
 - nsswitch.conf to use "sss" before "files"
 - very recent SSSD (that has https://github.com/SSSD/sssd/commit/9c4ac1bddf77634634df21dd548b11d3ca62aad7)

and try to login: after successful authentication session terminates.

Doesn't matter.

Linux/TigerVNC (Xvnc)

What is the reason to use `dup2()` and assign `vncInetdSock=3` unconditionally instead of `dup()` and obtain assigned fd using `F_GETFD`?

Issue is worked around with the help of https://github.com/SSSD/sssd/pull/7085 but I think this is a bug in TigerVNC

grulja:
Use dup() instead to get an available file descriptor to avoid hijacking already existing descriptor.

Fixes #1709

### DIFF
--- a/unix/xserver/hw/vnc/xvnc.c
+++ b/unix/xserver/hw/vnc/xvnc.c
@@ -366,8 +366,10 @@ ddxProcessArgument(int argc, char *argv[], int i)
     if (strcmp(argv[i], "-inetd") == 0) {
         int nullfd;
 
-        dup2(0, 3);
-        vncInetdSock = 3;
+        if ((vncInetdSock = dup(0)) == -1)
+            FatalError
+                ("Xvnc error: failed to allocate a new file descriptor for -inetd: %s\n", strerror(errno));
+
 
         /* Avoid xserver >= 1.19's epoll-fd becoming fd 2 / stderr only to be
            replaced by /dev/null by OsInit() because the pollfd is not


CendioOssman (approved):
Great, thanks!

PR #1716, opened 2024-01-12
x0vncserver: Add support for systemd socket activation

MikeLooijmans:
systemd can pass in sockets as file descriptors 3 and beyond. Allows the server to use socket activation.

### DIFF
--- a/unix/x0vncserver/CMakeLists.txt
+++ b/unix/x0vncserver/CMakeLists.txt
@@ -21,6 +21,12 @@ target_include_directories(x0vncserver PUBLIC ${CMAKE_SOURCE_DIR}/unix)
 target_include_directories(x0vncserver PUBLIC ${CMAKE_SOURCE_DIR}/common)
 target_link_libraries(x0vncserver tx rfb network rdr unixcommon)
 
+# systemd support (socket activation)
+if (LIBSYSTEMD_FOUND)
+  add_definitions(-DHAVE_SYSTEMD_H)
+  target_link_libraries(x0vncserver ${LIBSYSTEMD_LIBRARIES})
+endif()
+
 if(X11_FOUND AND X11_XTest_LIB)
   add_definitions(-DHAVE_XTEST)
   target_link_libraries(x0vncserver ${X11_XTest_LIB})
--- a/unix/x0vncserver/x0vncserver.cxx
+++ b/unix/x0vncserver/x0vncserver.cxx
@@ -38,6 +38,9 @@
 #include <rfb/Timer.h>
 #include <network/TcpSocket.h>
 #include <network/UnixSocket.h>
+#ifdef HAVE_SYSTEMD_H
+#  include <systemd/sd-daemon.h>
+#endif
 
 #include <signal.h>
 #include <X11/X.h>
@@ -112,6 +115,25 @@ static void CleanupSignalHandler(int /*sig*/)
 {
   caughtSignal = true;
 }
+
+#ifdef HAVE_SYSTEMD_H
+static int createSystemdListeners(std::list<SocketListener*> *listeners)
+{
+  int count = sd_listen_fds(0);
+
+  for (int i = 0; i < count; ++i) {
+      /* systemd sockets start at FD 3 */
+      listeners->push_back(new TcpListener(3 + i));
+  }
+
+  return count;
+}
+#else
+static int createSystemdListeners(std::list<SocketListener*> *)
+{
+  return 0;
+}
+#endif
 
 
 class FileTcpFilter : public TcpFilter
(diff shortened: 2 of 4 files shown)


CendioOssman:
Nice. Socket activation is something we've mostly thought about for `vncserver`. The use case for  `x0vncserver` is not as apparent, as it needs an existing X11 server that it can talk to.

Would you mind describing how you envision this to be used?

unix/x0vncserver/CMakeLists.txt
```diff
@@ -21,6 +21,13 @@ target_include_directories(x0vncserver PUBLIC ${CMAKE_SOURCE_DIR}/unix)
 target_include_directories(x0vncserver PUBLIC ${CMAKE_SOURCE_DIR}/common)
 target_link_libraries(x0vncserver tx rfb network rdr unixcommon)
 
+# check for systemd support (socket activation)
+pkg_check_modules(LIBSYSTEMD libsystemd)
```
CendioOssman, line 25: We try to put all checks in the top `CMakeLists.txt`, to make it easier to see TigerVNC's dependencies.

On that note, do you think you could add this as an optional dependency to `BUILDING.txt`?

unix/x0vncserver/x0vncserver.cxx
```diff
@@ -113,6 +116,22 @@ static void CleanupSignalHandler(int /*sig*/)
...
 }
 
+static int createSystemdListeners(std::list<SocketListener*> *listeners)
+{
+#ifdef HAVE_SYSTEMD_H
+  int count = sd_listen_fds(0);
+
+  for (int i = 0; i < count; ++i) {
+      /* systemd sockets start at FD 3 */
+      listeners->push_back(new TcpListener(3 + i));
```
CendioOssman, line 126: There's a `SD_LISTEN_FDS_START` constant. Can't we use that to make things more clear?

unix/x0vncserver/x0vncserver.cxx
```diff
@@ -113,6 +116,22 @@ static void CleanupSignalHandler(int /*sig*/)
   caughtSignal = true;
 }
 
+static int createSystemdListeners(std::list<SocketListener*> *listeners)
+{
+#ifdef HAVE_SYSTEMD_H
+  int count = sd_listen_fds(0);
```
CendioOssman, line 122: Maybe it's best to check for errors explicit. Less risk for future bugs.

unix/x0vncserver/x0vncserver.cxx
```diff
@@ -300,7 +319,9 @@ int main(int argc, char** argv)
 
     VNCServerST server(desktopName, &desktop);
 
-    if (rfbunixpath.getValueStr()[0] != '\0') {
+    if (createSystemdListeners(&listeners)) {
+      vlog.info("Listening on systemd sockets");
+    } else if (rfbunixpath.getValueStr()[0] != '\0') {
```
CendioOssman, line 324: The `else` seems improper here. Let's respect all the options given.

CendioOssman (changes requested):

unix/x0vncserver/x0vncserver.cxx
MikeLooijmans: When using systemd, ALL listening sockets will be passed through filedescriptors. The server shouldn't attempt to open any further sockets on its own. So this flow is on purpose. Otherwise you'd be obligated to explicitly disable the other (default) sockets.

unix/x0vncserver/x0vncserver.cxx
MikeLooijmans: Yes. The systemd docs said "3", a constant is better than a magic number.

MikeLooijmans:
> Nice. Socket activation is something we've mostly thought about for `vncserver`. The use case for `x0vncserver` is not as apparent, as it needs an existing X11 server that it can talk to.
> 
> Would you mind describing how you envision this to be used?

Add a a socket and service file, e.g. 

$ cat x0vncserver.service

```
[Unit]
Description=Remote desktop service (VNC) for :0 display
Requires=display-manager.service
After=display-manager.service

[Service]
Type=simple
ExecStart=x0vncserver -display :0 -rfbauth /etc/vncpasswd -MaxDisconnectionTime 3 -ClientWaitTimeMillis 3000
```
$ cat x0vncserver.socket 
```
[Socket]
ListenStream=5900

[Install]
WantedBy=sockets.target
```

Connecting to port 5900 will automatically start the server, and multiple clients can connect. 3 seconds after the last client disconnects, the server will shut down. When a new client connects, systemd will start a new instance of the server.

This allows one to always have a VNC server on "standby", but it doesn't consume any resources (apart from the listening socket) when not in use.

CendioOssman:
Thanks. That makes it clear how things can be set up. Perhaps that should be added to the man page?

I'm still not clear how the whole system is supposed to work, though. How have you made sure there is a useful X server on :0? And that x0vncserver is authorized to talk to it?

Is it the display manager that's on :0? What happens once you log in at that point? Won't the desktop be started at :1?

MikeLooijmans:
Our particular use case is an embedded system. Since that only every runs one desktop with one user, we can get away with hardcoding the DISPLAY. Other people with other use cases may have to write their own solutions to all the problems you're mentioning.

My focus here is to get systemd support into the server, because I think it'd be useful to others as well, and upstreaming it will also avoid having to maintain it locally.

I'm willing to spend a reasonable bit of effort to get it into the project, but the response I'm getting so far isn't very encouraging.

unix/x0vncserver/x0vncserver.cxx
MikeLooijmans: There were some changes between this version and the one I tested with, I've updated the code to handle things as intended.

CendioOssman:
We are definitely open to merging this. But please keep in mind that TigerVNC is used by a lot of people. So we need to be careful to balance the needs of all our users. Which means a feature may need more polish than was needed in its initial use case.

CendioOssman (approved):
I think this should be mostly done. Let me have a look at some compatibility tweaks, and then I can get this merged.

PR #1718, opened 2024-01-16
VNCServerST: Add a timeout to pointer button ownership

MikeLooijmans:
When one clients holds down a button on the pointer device (probably dragging something), other clients' attempts at pointer operations are denied. This yields a sane user experience, but with limits.

When one clients starts dragging, and then his network connection fails, other clients are denied access to the pointer until the VNC server finally discovers that the connection is dead and closes it. This can take about 15 minutes.

Add a timeout to this policy: If we don't hear from the client for 3 seconds, other clients are allowed to control the pointer once more.

This solves the problem that one failing network could make the server completely deaf to other clients for a long time.

### DIFF
--- a/common/rfb/VNCServerST.cxx
+++ b/common/rfb/VNCServerST.cxx
@@ -83,6 +83,7 @@ VNCServerST::VNCServerST(const char* name_, SDesktop* desktop_)
   : blHosts(&blacklist), desktop(desktop_), desktopStarted(false),
     blockCounter(0), pb(0), ledState(ledUnknown),
     name(name_), pointerClient(0), clipboardClient(0),
+    pointerClientTime(0),
     comparer(0), cursor(new Cursor(0, 0, Point(), NULL)),
     renderedCursorInvalid(false),
     keyRemapper(&KeyRemapper::defInstance),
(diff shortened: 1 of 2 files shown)


LMattsson:
That sounds like an annoying issue, thank you for looking into this. 

From your description, I assume this is seen with "shared" connections. A colleague and I were able to reproduce the issue with input being locked if a viewer loses connection.

Steps we used to reproduce:
 1. Running Xvnc with `-alwaysshared=true`
 2. Start an application and a window manager, e.g. xterm and metacity. Also start e.g. glxgears to show continuous graphical updates.
 3. Connected viewer A and viewer B. Viewer A is run on a VM where we control the network connection
 4. Start dragging the xterm window in viewer A
 5. While dragging, disconnect the network from the VM.
 6. In viewer B, note that input (both keyboard and mouse) are blocked. However, we still get graphical updates.

We also tested the regular case when two viewers are connected and one user is dragging a window. On master, the second user cannot interfere while the first user is dragging.

LMattsson (changes requested):
We built and tested the regular case with your changes. Unfortunately, it seems something is off. When a user is dragging the window in viewer A, viewer B can immediately interfere with the mouse.

We get the impression that the timeout isn't working as intended. Generally, it seems reasonable to keep blocking user B unless we detect a period of inactivity from user A.

MikeLooijmans:
Yep, my attempt to avoid a time(0) syscall when not needed backfired... Better to keep the code path simple.

LMattsson (approved):
Thank you for the quick update, looks good now! 

Rebuilt with the updated commit and tested both the regular use case as well as when one of the clients lost network. Tested the same way as described earlier.

Lastly, we also tested when viewer A stopped moving the mouse but still had the mouse button down. Viewer B could not interfere until more than 3 seconds had passed since viewer A stopped moving.

PR #1722, opened 2024-01-26
Fix shared memory leak

adamhalim:
The [MIT-SHM documentation](https://www.x.org/releases/X11R7.7/doc/xextproto/shm.html ) says to run XShmDetach() first, and then to destroy the segment.


CendioOssman:
Oops. We fixed this for the viewer back in 1.8.0, but apparently forgot to check other XShm users.

PR #1729, opened 2024-02-20
Add support to build Xvnc with meson

joantolo:
Since xorg-xserver version 20, it is possible to build the xserver with meson.

This updates the patches of xserver120 and xserver21.1.1 to build Xvnc with meson. Add the required meson.build on unix/xserver/hw/vnc too.

This is a first step to build tigervnc with meson.

### DIFF
--- a/unix/xserver/hw/vnc/meson.build
+++ b/unix/xserver/hw/vnc/meson.build
@@ -1,0 +1,115 @@
+tigervnc_srcdir = join_paths(meson.source_root(), '../..')
+tigervnc_builddir = tigervnc_srcdir
+
+cpp = meson.get_compiler('cpp')
+
+vnccommon_deps = [
+    declare_dependency(
+        dependencies: cpp.find_library('rfb', dirs: join_paths(tigervnc_builddir, 'common/rfb'))),
+    declare_dependency(
+        dependencies: cpp.find_library('rdr', dirs: join_paths(tigervnc_builddir, 'common/rdr'))),
+    declare_dependency(
+        dependencies: cpp.find_library('os', dirs: join_paths(tigervnc_builddir, 'common/os'))),
+    declare_dependency(
+        dependencies: cpp.find_library('network', dirs: join_paths(tigervnc_builddir, 'common/network'))),
+    declare_dependency(
+        dependencies: cpp.find_library('unixcommon', dirs: join_paths(tigervnc_builddir, 'unix/common'))),
+    dependency('zlib'),
+    dependency('pam'),
+    dependency('gnutls'),
+    dependency('nettle'),
+    dependency('hogweed'),
+    dependency('gmp'),
+    dependency('libjpeg'),
+]
+
+srcs_vnccommon = [
+    'qnum_to_xorgevdev.c',
+    'qnum_to_xorgkbd.c',
+    'RandrGlue.c',
+    'RFBGlue.cc',
+    'RFBGlue.h',
+    'vncBlockHandler.c',
+    'vncBlockHandler.h',
+    'vncExt.c',
+    'vncExtInit.cc',
+    'vncHooks.c',
+    'vncInput.c',
+    'vncInput.h',
+    'vncInputXKB.c',
+    'vncSelection.c',
+    'vncSelection.h',
+    'xorg-version.h',
+    'XorgGlue.c',
+    'XorgGlue.h',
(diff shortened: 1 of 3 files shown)


CendioOssman:
Thanks for your contribution. This could be interesting as autoconf is on the way out.

But, if I'm reading things correctly, the latest release of the Xorg server still has autoconf. So it would be the next release which would require meson?

I'm cautious about putting an extra burden on developers to update two build systems if we don't have to. At least not yet.

Was this PR for future proofing, or do you have a requirement to use meson for Xorg already?

unix/xserver/hw/vnc/meson.build
```diff
@@ -0,0 +1,115 @@
...
+    declare_dependency(
+        dependencies: cpp.find_library('rfb', dirs: join_paths(tigervnc_builddir, 'common/rfb'))),
+    declare_dependency(
+        dependencies: cpp.find_library('rdr', dirs: join_paths(tigervnc_builddir, 'common/rdr'))),
+    declare_dependency(
+        dependencies: cpp.find_library('os', dirs: join_paths(tigervnc_builddir, 'common/os'))),
+    declare_dependency(
+        dependencies: cpp.find_library('network', dirs: join_paths(tigervnc_builddir, 'common/network'))),
+    declare_dependency(
+        dependencies: cpp.find_library('unixcommon', dirs: join_paths(tigervnc_builddir, 'unix/common'))),
```
CendioOssman, line 16: I'm afraid I'm unfamiliar with meson. Will this still use our `.la` files? Or do we need to come up with some new glue for meson?

It is important that it not only finds the `.a` files, but also includes any dependent libraries. Possibly also cflags, but I'm unsure how well that even works currently.

CendioOssman (changes requested):

unix/xserver120.patch
```diff
@@ -60,6 +60,53 @@ Index: xserver/hw/Makefile.am
...
++add_project_arguments('-DHAVE_DIX_CONFIG_H', language: ['c', 'cpp'])
+ cc = meson.get_compiler('c')
+ 
+-add_global_arguments('-fno-strict-aliasing', language : 'c')
+-add_global_arguments('-fvisibility=hidden', language : 'c')
++add_global_arguments('-fno-strict-aliasing', language : ['c', 'cpp'])
++add_global_arguments('-fvisibility=hidden', language : ['c', 'cpp'])
+ 
+-add_global_link_arguments('-fvisibility=hidden', language : 'c')
++add_global_link_arguments('-fvisibility=hidden', language : ['c', 'cpp'])
```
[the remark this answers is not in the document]
joantolo: This is the way of copying the cflags arguments from 'c' to 'cpp': using the `add_global_link_arguments`.
This function can only be called before any meson target (i.e. executable, library) is declared, this is at the root meson.build file. At the hw/vnc/meson.build specific cflags can be added for each target.

The other option is to drop these changes on the patches and manually hardcode the cflags on hw/vnc/meson.build for each target (i.e. Xvnc, libvnc...) which might not be in sync with the cflags a new xserver version uses.

unix/xserver/hw/vnc/meson.build
joantolo: Hmm, I don't think meson is capable of handle .la files.

This first attempt manually adds the dependencies, as you can see (e.g. `dependency('zlib')`).

Now that I see, this works only on my build setup. A workaround could be to do the dynamic library selection to build the common libs at hw/vnc/meson.build. 

If the common libs were built with meson too, the library dependencies could be known.

joantolo:
> Was this PR for future proofing, or do you have a requirement to use meson for Xorg already?

No, I don't have any requirement to use meson. 
I just wanted to help for the future and meson. But I understand that having two build systems could be an extra burden.

unix/xserver/hw/vnc/meson.build
CendioOssman: That duplication is not something we'll reasonably be able to maintain.

Moving everything from CMake to meson seems overkill. So we want to add some glue, like we have right now between CMake and libtool.

Please have a look at `CMakeMacroLibtoolFile.cmake` and see how we can do something similar for meson. I.e. automatically generate dependency information in a format meson can consume.

CendioOssman:
Closing in favour of #2064.

PR #1738, opened 2024-03-16
Make SELinux PAM modules `optional`

62832:
Currently, on Unix systems without SELinux enabled, `vncsession` will consistently fail and exit with status 71 due to the `pam_selinux.so` module being marked as `required`. To get around this, various distributions such as [Arch Linux](https://gitlab.archlinux.org/archlinux/packaging/packages/tigervnc/-/blob/main/remove-selinux.patch) (which doesn't officially support SELinux anyway) and its derivatives such as [Artix Linux](https://gitea.artixlinux.org/packages/tigervnc/src/branch/master/remove-selinux.patch) are forced to apply their own patch at build-time to completely remove the SELinux module from TigerVNC's PAM policy.

This PR eliminates the need for such a patch to be applied by distribution maintainers by simply marking the SELinux module as `optional`.

### DIFF
--- a/unix/vncserver/tigervnc.pam
+++ b/unix/vncserver/tigervnc.pam
@@ -1,8 +1,8 @@
 #%PAM-1.0
 # pam_selinux.so close should be the first session rule
--session   required     pam_selinux.so close
+-session   optional     pam_selinux.so close
 session    required     pam_loginuid.so
--session   required     pam_selinux.so open
+-session   optional     pam_selinux.so open
 session    required     pam_namespace.so
 session    optional     pam_keyinit.so force revoke
 session    required     pam_limits.so


CendioOssman:
On SELinux systems, it is likely dangerous to set this as `optional`.

The real issue seems to be a misunderstanding that this configuration is supposed to work on every distribution.

Rather, this file is just an example. It needs to be adapted (or even fully replaced) with a configuration that matches the specific distribution.

So, perhaps we need to add better documentation instead?

62832:
The documentation (e.g. `man` page) for `pam_selinux.so` itself shows example usage in which it is set as `optional`, so I assumed based on that that it wouldn't pose much issue. I could be very much mistaken, however.

CendioOssman:
Indeed it does. Interesting.

Both Fedora and Red Hat configure it as `required`, though. So I would trust that more as the norm. I've also seen issues with sessions misbehaving badly when pam_selinux fails, and they keep running as the service context (`initrc_t`).

CendioOssman:
I added a note in c17c163e1c542b4be67cc3fc247bf5294e3896f5 that this is an example. Let's see if that helps clarify and encourage users to not use the file as-is.

Issue #843, opened 2019-06-09
Invalid hotSpot


stefan-reich:
**Describe the bug**
The viewer says "Invalid hotSpot" and crashes.

**To Reproduce**

Connect from Linux client to Linux server with SSH tunneling using the TigerVNC Java client.

**Expected behavior**
Normal stuff

**Client (please complete the following information):**
 - TigerVNC Java 1.9.0

**Log**

```
Tunnel: Next authentication method: publickey
Tunnel: Authentication succeeded (publickey).
CConn: connected to host localhost port 5901
CConnection: Server supports RFB protocol version 3.8
CConnection: Using RFB protocol version 3.8
Exception in thread "AWT-EventQueue-0" java.lang.ClassCastException: java.desktop/javax.swing.JLayer cannot be cast to java.desktop/javax.swing.AbstractButton
	at java.desktop/javax.swing.plaf.synth.SynthButtonUI.getComponentState(SynthButtonUI.java:154)
	at java.desktop/javax.swing.plaf.synth.SynthButtonUI.getContext(SynthButtonUI.java:135)
	at java.desktop/javax.swing.plaf.synth.SynthBorder.paintBorder(SynthBorder.java:56)
	at java.desktop/javax.swing.JComponent.paintBorder(JComponent.java:967)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1075)
	at java.desktop/javax.swing.JLayer.paint(JLayer.java:475)
	at java.desktop/javax.swing.plaf.LayerUI.paint(LayerUI.java:80)
	at java.desktop/javax.swing.plaf.ComponentUI.update(ComponentUI.java:161)
	at java.desktop/javax.swing.JComponent.paintComponent(JComponent.java:797)
	at java.desktop/javax.swing.JLayer.paint(JLayer.java:470)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:907)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1083)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:907)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1083)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:907)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1083)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:907)
[...]

PR #1742, opened 2024-03-30
Fix for issue 843

bphinz:


bphinz:
Fixes issue #843

fnJeff:
Unfortunately, I don't think this is fixed in java client 1.14.0.  I just tried this version for the first time today and experienced the same crash described in #843.  Everything seemed fine until I clicked in a konsole and typed 1 character.  Viewer immediately disappeared and the Invalid hotspot error popup window showed up.  I don't know exactly what info you need to help debug, but if you let me know, I will try and get you what I can.

PR #1760, opened 2024-05-30
More XDG directory fixes

CendioOssman:
Mainly, make sure that pam_env can be used to influence where the log file ends up.


CendioOssman:
This is a continuation of #1737.

Issue #370, opened 2016-10-19
Service accepts different password


GereeNice:
Hey there!

I have a shared OS X / Win10 desktop pc, and I wanted to be able to reach Windows also, so I installed x64 1.7.0 . Set up Standard VNC authentication, and tested it with RealVNC from my iPhone, only to notice it connected without entering the password.

The phone has a 21 char long saved password for OS X, that starts, but is not the same as Windows'. Is the password being cut before verifying? If there is a limit to password length a user notification would be important when setting it.

Or if there is no such thing, then I have no idea why it lets me in and that's a worrying thought.

What could be the cause?

hildred:
The standard vnc authentication is a textbook example of how not to do passwords. It contains numerous design flaws including truncation, which would explain what you are seeing. Anyone trusting a plain or vnc password without some sort of channel encryption is unprepared to face an attacker more sophisticated than ... well words fail. At least with security=none you know how little security you have.

GereeNice:
And if you happen to have the time, do you have any suggestions what could make a safe and good solution that allows me to connect to the built in os x screen sharing, and vica versa? So far Tiger with standard is the only one I found to work. I wouldn't want to install another service since this is available on any machine around by default, and I also haven't made up my mind to set up vpn only for this either.

I know this might be kind of off topic, therefore I value your input even more.

CendioOssman:
Apple's screen sharing isn't really VNC compatible, so I would recommend a client that is specifically written for that server. No idea how the security is in the different varieties though.

COD3HUNT3R:
Current limitation for 8 character password is due to https://datatracker.ietf.org/doc/html/rfc6143#section-7.2.2.

```
 The client encrypts the challenge with DES, using a password supplied
 by the user as the key.  To form the key, the password is truncated
 to eight characters, or padded with null bytes on the right.  The
 client then sends the resulting 16-byte response
```

But every implementation out there probably missed this. https://datatracker.ietf.org/doc/html/rfc6143#section-6
```
New security types
      Adding a new security type gives full flexibility in modifying the
      behavior of the protocol without sacrificing compatibility with
      existing clients and servers.  A client and server that agree on a
      new security type can effectively talk whatever protocol they like
      after that -- it doesn't necessarily have to be anything like the
      RFB protocol.
```

You can obtain an ID from IANA for security type and it would still be a compliant VNC server https://datatracker.ietf.org/doc/html/rfc6143#section-8

```
[...]

PR #1762, opened 2024-06-02
vncpasswd add pwquality complexity rule check

prownd:
Add a password complexity strategy for VNC user authentication
enhance password quality verification checks, and enhance security

Additionally, Chinese translation has been added

### DIFF
--- a/unix/vncpasswd/CMakeLists.txt
+++ b/unix/vncpasswd/CMakeLists.txt
@@ -4,5 +4,9 @@ add_executable(vncpasswd
 target_include_directories(vncpasswd PUBLIC ${CMAKE_SOURCE_DIR}/common)
 target_link_libraries(vncpasswd tx rfb os)
 
+if(PWQUALITY_FOUND)
+  target_link_libraries(vncpasswd pwquality)
+endif()
+
 install(TARGETS vncpasswd DESTINATION ${CMAKE_INSTALL_FULL_BINDIR})
 install(FILES vncpasswd.man DESTINATION ${CMAKE_INSTALL_FULL_MANDIR}/man1 RENAME vncpasswd.1)
--- a/unix/vncpasswd/vncpasswd.cxx
+++ b/unix/vncpasswd/vncpasswd.cxx
@@ -149,6 +149,11 @@ static std::vector<uint8_t> readpassword() {
       continue;
     }
 
+    if (first.size() > 8) {
+      fprintf(stderr,"Password should not be greater than 8 characters\nBecause only 8 valid characters are used - try again\n");
+      continue;
+    }
+
 #ifdef HAVE_PWQUALITY
     //the function return score of password quality
     int r = check_passwd_pwquality(passwd);
--- a/unix/vncpasswd/vncpasswd.cxx
+++ b/unix/vncpasswd/vncpasswd.cxx
@@ -37,6 +37,9 @@
 
 #include <termios.h>
 
+#ifdef HAVE_PWQUALITY
+#include <pwquality.h>
+#endif
 
 using namespace rfb;
 
@@ -99,6 +102,36 @@ static int encrypt_pipe() {
   return 0;
 }
 
+#ifdef HAVE_PWQUALITY
+static int check_passwd_pwquality(const char *password)
+{
+	int r;
+	void *auxerror;
+	pwquality_settings_t *pwq;
+	pwq = pwquality_default_settings();
+	if (!pwq)
+		return -EINVAL;
+	r = pwquality_read_config(pwq, NULL, &auxerror);
+	if (r) {
+		printf("Cannot check password quality: %s \n",
+			pwquality_strerror(NULL, 0, r, auxerror));
+		pwquality_free_settings(pwq);
+		return -EINVAL;
+	}
+
+	r = pwquality_check(pwq, password, NULL, NULL, &auxerror);
+	if (r < 0) {
+		printf("Password quality check failed:\n %s \n",
+			pwquality_strerror(NULL, 0, r, auxerror));
+		r = -EPERM;
+	}
+	pwquality_free_settings(pwq);
+
+	//return the score of password quality
+	return r;
+}
+#endif
+
 static std::vector<uint8_t> readpassword() {
   while (true) {
     const char *passwd = getpassword("Password:");
@@ -115,6 +148,15 @@ static std::vector<uint8_t> readpassword() {
       fprintf(stderr,"Password must be at least 6 characters - try again\n");
       continue;
     }
+
+#ifdef HAVE_PWQUALITY
+    //the function return score of password quality
+    int r = check_passwd_pwquality(passwd);
+    if (r < 0){
+      printf("Password quality check failed, please set it correctly.\n");
+      continue;
+    }
+#endif
 
     passwd = getpassword("Verify:");
     if (passwd == NULL) {
--- a/BUILDING.txt
+++ b/BUILDING.txt
@@ -55,6 +55,9 @@ Build Requirements (Unix)
    * You might have to enable additional repositories for this. E.g.,
      on RHEL, EPEL and RPMFusion (free + nonfree) need to be enabled.
 
+-- If building vncpasswd with password quality check support:
+   * libpwquality
+
 ============================
 Build Requirements (Windows)
 ============================
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -324,6 +324,20 @@ if(UNIX AND NOT APPLE)
   endif()
 endif()
 
+# check for password pwquality check support
+option(ENABLE_PWQUALITY "Enable password pwquality check" ON)
+if(ENABLE_PWQUALITY)
+  if(UNIX)
+    find_package(PkgConfig)
+    if(PKG_CONFIG_FOUND)
+      pkg_check_modules(PWQUALITY pwquality)
+      if(PWQUALITY_FOUND)
+        add_definitions(-DHAVE_PWQUALITY)
+      endif()
+    endif()
+  endif()
+endif()
+
 # Generate config.h and make sure the source finds it
 configure_file(config.h.in config.h)
 add_definitions(-DHAVE_CONFIG_H)


CendioOssman:

I'm concerned how well a pwquality will work for VNC passwords, given that they are limited to 8 characters. #370 would probably need to be implemented first. Otherwise the user might come up with a long and secure to keep pwquality happy, but overlook the fact that the actual used password is much shorter.

It should probably also just be a warning when pwquality fails, not an error, given that it is difficult to get a good password with so few characters.

prownd:
Yes, due to the VNC protocol, the actual length of the VNC password used is 8 characters.
The first version I modified did indeed limit the maximum length to 8 characters. However, considering that there has always been no limit on the maximum length when setting passwords,  so I removed the limit of 8 characters.
According to your suggestion, when pwquality fails, there will only be an alarm prompt. Instead of terminating password settings?

prownd:
> Thank you for your contribution!
> 
> I'm concerned how well a pwquality will work for VNC passwords, given that they are limited to 8 characters. #370 would probably need to be implemented first. Otherwise the user might come up with a long and secure to keep pwquality happy, but overlook the fact that the actual used password is much shorter.
> 
> It should probably also just be a warning when pwquality fails, not an error, given that it is difficult to get a good password with so few characters.

According to your opinion, when setting a password, add an 8-character limit.
Can I modify and submit it again ?

prownd:
1.  Use libpwquality to check password quality and complexity, improving security.
    The validation rules for libpwquality are as follows:
    - The minimum password length is 6 and the maximum is 8
    -  Password characters repeat once
    -  At least 3 types of characters are required
2.  Add i18n internationalization support and Chinese localization translation.

prownd:
> 1. Use libpwquality to check password quality and complexity, improving security.
>    The validation rules for libpwquality are as follows:
>    
>    * The minimum password length is 6 and the maximum is 8
>    * Password characters repeat once
>    * At least 3 types of characters are required
> 2. Add i18n internationalization support and Chinese localization translation.

@CendioOssman Could you please take a look and see if there is anything else that needs to be added.

po/zh_CN.po
```diff
@@ -703,3 +703,31 @@ msgstr "正在监听端口 %d"
...
+msgid "Password should not be greater than 8 characters\nBecause only 8 valid characters are used - try again\n"
+msgstr "密码长度不应该大于8位,因为仅只有8位被使用,请重试\n"
+
+#: unix/vncpasswd/vncpasswd.cxx:169
+msgid "Password quality check failed, please set it correctly.\n"
+msgstr "密码复杂度检查失败,请设置正确的密码\n"
+
+#: unix/vncpasswd/vncpasswd.cxx:181
+msgid "Passwords don't match - try again\n"
+msgstr "密码不匹配,请重试\n"
```
CendioOssman, line 733: I'm afraid our translations are handled by the translation project. As a result, we cannot modify these files directly or they'll get overwritten. Please send your translations to them once a beta is out of TigerVNC with your changes in it.

unix/vncpasswd/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ add_executable(vncpasswd
   vncpasswd.cxx)
 
 target_include_directories(vncpasswd PUBLIC ${CMAKE_SOURCE_DIR}/common)
-target_link_libraries(vncpasswd tx rfb os)
+target_link_libraries(vncpasswd tx rfb os pwquality)
```
CendioOssman, line 5: How commont is pwquality? Is it available for all distributions? Otherwise it will need to be an optional dependency.

Could you please also add it to BUILDING.txt?

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -37,6 +37,15 @@
 
 #include <termios.h>
 
+#include <pwquality.h>
+#include <locale.h>
+#include <libintl.h>
```
CendioOssman, line 42: Please split adding translation support as a separate commit. It makes it easier to see what parts are associated.

CendioOssman, line 42: Let's try to use the same convenience wrappers that we have for vncviewer. Please refer to the files `i18n.h` and `gettext.h` there.

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -37,6 +37,15 @@
...
+
+#define PACKAGE_NAME "tigervnc"
+#define PACKAGE_NAME2 "libpwquality"
+#define LOCALE_DIR  "/usr/share/locale"
```
CendioOssman, line 46: This should come from CMake. Please refer to vncviewer on how to get this.

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -99,6 +108,39 @@ static int encrypt_pipe() {
...
+		printf(_("Cannot check password quality: %s \n"),
+			pwquality_strerror(NULL, 0, r, auxerror));
+		pwquality_free_settings(pwq);
+		return -EINVAL;
+	}
+
+	pwquality_set_int_value(pwq, PWQ_SETTING_MIN_LENGTH, 6);
+	pwquality_set_int_value(pwq, PWQ_SETTING_MAX_SEQUENCE, 8);
+	pwquality_set_int_value(pwq, PWQ_SETTING_MAX_REPEAT, 1);
+	pwquality_set_int_value(pwq, PWQ_SETTING_MIN_CLASS, 3);
```
CendioOssman, line 130: This doesn't seem proper to override. Isn't this configured by the system administrator in `/etc/security/pwquality.conf`?

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -109,10 +151,22 @@ static std::vector<uint8_t> readpassword() {
     std::string first = passwd;
     if (first.size() < 6) {
       if (first.empty()) {
-        fprintf(stderr,"Password not changed\n");
+        fprintf(stderr,_("Password not changed\n"));
         exit(1);
       }
-      fprintf(stderr,"Password must be at least 6 characters - try again\n");
+      fprintf(stderr,_("Password must be at least 6 characters - try again\n"));
```
CendioOssman, line 157: This is now policy that could conflict with what pwquality wants. It should be removed if we always use pwquality, or used as a fallback if pwquality ends up being optional.

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -109,10 +151,22 @@ static std::vector<uint8_t> readpassword() {
...
+      continue;
+    }
+
+    if (first.size() > 8) {
+      fprintf(stderr,_("Password should not be greater than 8 characters\nBecause only 8 valid characters are used - try again\n"));
```
CendioOssman, line 162: It's best to avoid formatting in the strings as it makes it difficult to translate.

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -109,10 +151,22 @@ static std::vector<uint8_t> readpassword() {
...
-        fprintf(stderr,"Password not changed\n");
+        fprintf(stderr,_("Password not changed\n"));
         exit(1);
       }
-      fprintf(stderr,"Password must be at least 6 characters - try again\n");
+      fprintf(stderr,_("Password must be at least 6 characters - try again\n"));
+      continue;
+    }
+
+    if (first.size() > 8) {
```
CendioOssman, line 161: Nice! But could you split this to a separate commit? It's independent of using pwquality.

unix/vncpasswd/vncpasswd.cxx
```diff
@@ -133,6 +187,11 @@ static std::vector<uint8_t> readpassword() {
 
 int main(int argc, char** argv)
 {
+  setlocale(LC_ALL, "");
+  bindtextdomain(PACKAGE_NAME, LOCALE_DIR);
+  bind_textdomain_codeset(PACKAGE_NAME2, "UTF-8");
```
CendioOssman, line 192: This seems wrong. Why is this needed? We should not need to mess with a library's translation handling. And UTF-8 might not be the correct character set for the terminal.

CendioOssman (changes requested):
Looks really good! There are some tweaks that are needed, though. Please have a look at my comments.

po/zh_CN.po
prownd: Okay, I understand. Maybe I didn't consider it before.

unix/vncpasswd/CMakeLists.txt
prownd: I checked and tested the pwqaulity.pc of the rpm and deb based distribution (centos/debian)
The compilation link parameters are all "- l pwquality"
It should be universal, and I will test and check other distributions later.

unix/vncpasswd/vncpasswd.cxx
prownd: Okay, it really should be separated.

prownd: Okay, I'll try to modify it.

unix/vncpasswd/vncpasswd.cxx
prownd: I was not very familiar with CMake before
Okay, I will make the necessary modifications according to your suggestions

unix/vncpasswd/vncpasswd.cxx
prownd: The default rules are in pwquality.conf
If specified in the code, the rules will be rewritten.

unix/vncpasswd/CMakeLists.txt
prownd: Compiling on Debian requires installing the libpwquality-dev package.

unix/vncpasswd/vncpasswd.cxx
prownd: Perhaps they should be submitted separately

unix/vncpasswd/vncpasswd.cxx
prownd: sorry，
Is it adding a new commit in the current PR. 
Or create a new PR to handle translation.

prownd:
@CendioOssman  
I have made modifications according to your suggestions. 
Could you please review it to see if it is correct.

unix/vncpasswd/vncpasswd.cxx
CendioOssman: It can be part of this PR as long as it's a separate commit.

CendioOssman (changes requested):

unix/vncpasswd/vncpasswd.cxx
CendioOssman: A separate commit, yes. But you don't need to open a separate PR for such a small fix.

prownd:
Could you please take a moment to take a  review  again .  @CendioOssman

prownd:
@CendioOssman Please review again.
Is your suggestion to use the default/etc/security/pwquality.conf configuration rules?
But as mentioned earlier, the actual maximum length of a password is only 8, which may have generated a very long and complex password, but only 8 lengths were actually used.
So it is very necessary to specify some rules in the code.

Looking forward to your reply and suggestions.

CendioOssman:
Sorry for the late response. I've been away for a few weeks.

There are still some open requests that haven't been addressed. Please have a look at them and see if we can get the last bits fixed.

CendioOssman:
> Is your suggestion to use the default/etc/security/pwquality.conf configuration rules?

Yes, that is correct.

> But as mentioned earlier, the actual maximum length of a password is only 8, which may have generated a very long and complex password, but only 8 lengths were actually used.
> So it is very necessary to specify some rules in the code.

If the system policy doesn't allow passwords that are compatible with VNC, then the system should refuse to continue. It shouldn't silently change the policy and allow things that the user thinks have been blocked.

As for only 8 characters being used, your previous suggestion had a check for that. But it seems to have been dropped?

prownd:
> @CendioOssman Please review again. Is your suggestion to use the default/etc/security/pwquality.conf configuration rules? But as mentioned earlier, the actual maximum length of a password is only 8, which may have generated a very long and complex password, but only 8 lengths were actually used. So it is very necessary to specify some rules in the code.
> 
> Looking forward to your reply and suggestions.



> > Is your suggestion to use the default/etc/security/pwquality.conf configuration rules?
> 
> Yes, that is correct.
> 
> > But as mentioned earlier, the actual maximum length of a password is only 8, which may have generated a very long and complex password, but only 8 lengths were actually used.
> > So it is very necessary to specify some rules in the code.
> 
> If the system policy doesn't allow passwords that are compatible with VNC, then the system should refuse to continue. It shouldn't silently change the policy and allow things that the user thinks have been blocked.
> 
> As for only 8 characters being used, your previous suggestion had a check for that. But it seems to have been dropped?



> > Is your suggestion to use the default/etc/security/pwquality.conf configuration rules?
> 
> Yes, that is correct.
Okay, then use the default pwquality-oriented. conf rule
> 
> > But as mentioned earlier, the actual maximum length of a password is only 8, which may have generated a very long and complex password, but only 8 lengths were actually used.
> > So it is very necessary to specify some rules in the code.
> 
> If the system policy doesn't allow passwords that are compatible with VNC, then the system should refuse to continue. It shouldn't silently change the policy and allow things that the user thinks have been blocked.
> 
> As for only 8 characters being used, your previous suggestion had a check for that. But it seems to have been dropped?

Because it's independent of using pwquality
Your suggestion is to submit a separate commit


> > Is your suggestion to use the default/etc/security/pwquality.conf configuration rules?
> 
> Yes, that is correct.
> 
> > But as mentioned earlier, the actual maximum length of a password is only 8, which may have generated a very long and complex password, but only 8 lengths were actually used.
> > So it is very necessary to specify some rules in the code.
> 
> If the system policy doesn't allow passwords that are compatible with VNC, then the system should refuse to continue. It shouldn't silently change the policy and allow things that the user thinks have been blocked.
> 
> As for only 8 characters being used, your previous suggestion had a check for that. But it seems to have been dropped?

prownd:
> > Is your suggestion to use the default/etc/security/pwquality.conf configuration rules?
> 
> Yes, that is correct.

Okay, then use the default pwquality.conf rule

>
> > But as mentioned earlier, the actual maximum length of a password is only 8, which may have generated a very long and complex password, but only 8 lengths were actually used.
> > So it is very necessary to specify some rules in the code.
> 
> If the system policy doesn't allow passwords that are compatible with VNC, then the system should refuse to continue. It shouldn't silently change the policy and allow things that the user thinks have been blocked.
> 
> As for only 8 characters being used, your previous suggestion had a check for that. But it seems to have been dropped?

Because it's independent of using pwquality
Your suggestion is to submit a separate commit

prownd:
@CendioOssman  According to your suggestion, it has been modified. 
Could you please review it again

CendioOssman (approved):
Looks good! Thank you for getting this done.

PR #1767, opened 2024-06-19
Updates for package builds

bphinz:


contrib/packages/deb/ubuntu-noble/debian/control
```diff
@@ -0,0 +1,193 @@
+Source: tigervnc
+Section: x11
+Priority: optional
+Maintainer: Brian P. Hinz <bphinz@users.sourceforge.net>
+Standards-Version: 3.8.4
+Build-Depends:
+# TigerVNC dependencies
```
CendioOssman, line 7: I see you did some cleanups in this list. Perhaps best to do in the other variants as well to make things easy to keep in sync?

On that note, gcrypt is not that commonly used any more. Are we sure Ubuntu still uses that for Xorg?

contrib/packages/deb/ubuntu-noble/debian/copyright
```diff
@@ -0,0 +1,116 @@
+This package was packaged for Debian by Brian P. Hinz <bphinz@users.sourceforge.net>
+on Tue, 18 Jun 2024 18:54:03 -0400 using the tightvnc package as a base.
```
CendioOssman, line 2: I hadn't noticed this file before. We're obviously not keeping this copy in sync properly. Can we automate that? Or better yet, just put a notice here to check the README?

CendioOssman (approved):
Looks good to me! Just squash the commits and it should be good to merge.

I made some notes of general improvements we might consider for all Debian packages.

contrib/packages/deb/ubuntu-noble/debian/control
bphinz: libgcrypt is still listed as a build requirement in Ubuntu's xorg-server package

contrib/packages/deb/ubuntu-noble/debian/copyright
bphinz: I don't know if there is an automated way to keep it in sync but I will take a look at what debian's packaging guidelines are regarding this file.  For now I've updated them all manually.

CendioOssman (changes requested):
I think something went wrong here. The merge should not be in to this branch.

CendioOssman (approved):

CendioOssman:
I would suggest resolving it manually on the command line as part of merging to `master`. Or rebase your original commit after bionic was removed.

PR #1783, opened 2024-07-22
Use std::min replace min

KangLin:
Use std::min replace min

### DIFF
--- a/common/rdr/HexInStream.cxx
+++ b/common/rdr/HexInStream.cxx
@@ -21,13 +21,12 @@
 #include <config.h>
 #endif
 
+#include <algorithm>
 #include <rdr/HexInStream.h>
 #include <rdr/Exception.h>
 #include <rfb/util.h>
 
 using namespace rdr;
-
-static inline int min(int a, int b) {return a<b ? a : b;}
 
 HexInStream::HexInStream(InStream& is)
 : in_stream(is)
@@ -37,12 +36,11 @@ HexInStream::HexInStream(InStream& is)
 HexInStream::~HexInStream() {
 }
 
-
 bool HexInStream::fillBuffer() {
   if (!in_stream.hasData(2))
     return false;
 
-  size_t length = min(in_stream.avail()/2, availSpace());
+  size_t length = std::min(in_stream.avail()/2, availSpace());
   const uint8_t* iptr = in_stream.getptr(length*2);
 
   uint8_t* optr = (uint8_t*) end;
--- a/common/rdr/HexOutStream.cxx
+++ b/common/rdr/HexOutStream.cxx
@@ -20,13 +20,11 @@
 #ifdef HAVE_CONFIG_H
 #include <config.h>
 #endif
-
+#include <algorithm>
 #include <rdr/HexOutStream.h>
 #include <rfb/util.h>
 
 using namespace rdr;
-
-static inline size_t min(size_t a, size_t b) {return a<b ? a : b;}
 
 HexOutStream::HexOutStream(OutStream& os)
   : out_stream(os)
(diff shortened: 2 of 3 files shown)


common/rdr/HexOutStream.cxx
```diff
@@ -41,7 +40,7 @@ bool HexOutStream::flushBuffer()
 {
   while (sentUpTo != ptr) {
     uint8_t* optr = out_stream.getptr(2);
-    size_t length = min(ptr-sentUpTo, out_stream.avail()/2);
+    size_t length = std::min(ptr-sentUpTo, (int64_t)out_stream.avail()/2);
```
CendioOssman, line 43: `int64_t` seems like an odd choice here as neither the left or right side as that type. Why not `size_t` or `ptrdiff_t`?

common/rdr/HexInStream.cxx
```diff
@@ -24,11 +24,10 @@
 #include <rdr/HexInStream.h>
 #include <rdr/Exception.h>
 #include <rfb/util.h>
+#include <algorithm>
```
CendioOssman, line 27: We try to sort system includes before the local ones, and a blank line to separate them. Makes it easier to get an overview of the includes.

CendioOssman (changes requested):
The PR looks mostly fine, but I'm unclear to the motivation. You mention "cross-platform", but how does this change improve that?

common/rdr/HexInStream.cxx
```diff
@@ -21,13 +21,16 @@
...
 
+#include <algorithm>
 #include <rdr/HexInStream.h>
 #include <rdr/Exception.h>
 #include <rfb/util.h>
 
-using namespace rdr;
+#if defined(_MSC_VER)
+#undef min
+#endif
```
CendioOssman, line 31: This is a surprising new addition. Are you saying that std::min cannot be safely used in Visual Studio?

common/rdr/HexInStream.cxx
```diff
@@ -37,12 +40,11 @@ HexInStream::HexInStream(InStream& is)
 HexInStream::~HexInStream() {
 }
 
-
 bool HexInStream::fillBuffer() {
   if (!in_stream.hasData(2))
     return false;
-
-  size_t length = min(in_stream.avail()/2, availSpace());
+
```
CendioOssman, line 46: Some whitespace snuck in here.

CendioOssman (changes requested):

common/rdr/HexInStream.cxx
KangLin: In fact, the mix macro is defined in the minwindef.h header file, while Windows.h references minwindef.h. The mix macro is defined as follows:
```
#ifndef NOMINMAX
 
#ifndef max
#define max(a,b)            (((a) > (b)) ? (a) : (b))
#endif
 
#ifndef min
#define min(a,b)            (((a) < (b)) ? (a) : (b))
#endif
 
#endif  /* NOMINMAX */
```

Maybe it's better to define  NOMINMAX

CendioOssman (changes requested):
I think the code looks good now. But I'm still not clear on the rationale. What was the reason you started working on this fix?

KangLin:
I am will use to msvc.

在 2024-07-29 22:06:42，"Pierre Ossman (ThinLinc team)" ***@***.***> 写道：


@CendioOssman requested changes on this pull request.

I think the code looks good now. But I'm still not clear on the rationale. What was the reason you started working on this fix?

CendioOssman (approved):
> - There are standard functions available, so why define them yourself?

Generally, that's a good principle. But `std::min()` has some bad behaviours based on the fact that it is implemented as a template, which means it isn't always a good fit.

We also don't want to avoid creating unnecessary history, as that makes `git blame` annoying to use. In other words, there should be a decently sized cleanup.

That said, this didn't change too many lines. So I can accept this if it makes things easier for you.

> - I am will use to msvc.

Ah. That's not one of our normally used compilers, but it's always a plus if we're more compatible.

KangLin:
> But `std::min()` has some bad behaviours based on the fact that it is implemented as a template, which means it isn't always a good fit.

Please describe any bad behaviours？

PR #1784, opened 2024-07-22
H264Decoder: Fix context reset

any1:
This fixes a regression introduced by
12b3f4021641537b90727b23d42de5dff59006cd.


CendioOssman:
Crap. I missed one usage. Thanks for the fix!

PR #1785, opened 2024-07-22
h264: Use consistent size of scaling buffer

any1:
This fixes out-of-bounds writes when there is a mismatch between the pixel buffer's stride and the rect's width or between the frame's height and the rect's height.

### DIFF
--- a/common/rfb/H264LibavDecoderContext.cxx
+++ b/common/rfb/H264LibavDecoderContext.cxx
@@ -43,7 +43,6 @@ bool H264LibavDecoderContext::initCodec() {
   os::AutoMutex lock(&mutex);
 
   sws = nullptr;
-  swsBuffer = nullptr;
   h264WorkBuffer = nullptr;
   h264WorkBufferLength = 0;
 
@@ -87,9 +86,6 @@ bool H264LibavDecoderContext::initCodec() {
     return false;
   }
 
-  int numBytes = av_image_get_buffer_size(AV_PIX_FMT_RGB32, rect.width(), rect.height(), 1);
-  swsBuffer = new uint8_t[numBytes];
-
   initialized = true;
   return true;
 }
@@ -101,8 +97,8 @@ void H264LibavDecoderContext::freeCodec() {
     return;
   av_parser_close(parser);
   avcodec_free_context(&avctx);
+  av_frame_free(&rgbFrame);
   av_frame_free(&frame);
-  delete[] swsBuffer;
   free(h264WorkBuffer);
   initialized = false;
 }
@@ -220,11 +216,22 @@ void H264LibavDecoderContext::decode(const uint8_t* h264_in_buffer,
                              frame->width, frame->height, AV_PIX_FMT_RGB32,
                              0, nullptr, nullptr, nullptr);
 
-  int stride;
-  pb->getBuffer(rect, &stride);
-  int dst_linesize = stride * pb->getPF().bpp/8;  // stride is in pixels, linesize is in bytes (stride x4). We need bytes
+  if (rgbFrame && (rgbFrame->width != frame->width || rgbFrame->height != frame->height)) {
+    av_frame_free(&rgbFrame);
 
-  sws_scale(sws, frame->data, frame->linesize, 0, frame->height, &swsBuffer, &dst_linesize);
+  }
 
-  pb->imageRect(rect, swsBuffer, stride);
+  if (!rgbFrame) {
+    rgbFrame = av_frame_alloc();
+    // TODO: Can we really assume that the pixel format will always be RGB32?
+    rgbFrame->format = AV_PIX_FMT_RGB32;
+    rgbFrame->width = frame->width;
+    rgbFrame->height = frame->height;
+    av_frame_get_buffer(rgbFrame, 0);
+  }
+
+  sws_scale(sws, frame->data, frame->linesize, 0, frame->height, rgbFrame->data,
+            rgbFrame->linesize);
+
+  pb->imageRect(rect, rgbFrame->data[0], rgbFrame->linesize[0] / 4);
 }
--- a/common/rfb/H264LibavDecoderContext.cxx
+++ b/common/rfb/H264LibavDecoderContext.cxx
@@ -99,6 +99,7 @@ void H264LibavDecoderContext::freeCodec() {
   avcodec_free_context(&avctx);
   av_frame_free(&rgbFrame);
   av_frame_free(&frame);
+  sws_freeContext(sws);
   free(h264WorkBuffer);
   initialized = false;
 }
--- a/common/rfb/H264LibavDecoderContext.h
+++ b/common/rfb/H264LibavDecoderContext.h
@@ -47,8 +47,8 @@ namespace rfb {
       AVCodecContext *avctx;
       AVCodecParserContext *parser;
       AVFrame* frame;
+      AVFrame* rgbFrame = nullptr;
       SwsContext* sws;
-      uint8_t* swsBuffer;
       uint8_t* h264WorkBuffer;
       uint32_t h264WorkBufferLength;
   };


common/rfb/H264LibavDecoderContext.cxx
```diff
@@ -222,9 +222,19 @@ void H264LibavDecoderContext::decode(const uint8_t* h264_in_buffer,
 
   int stride;
   pb->getBuffer(rect, &stride);
-  int dst_linesize = stride * pb->getPF().bpp/8;  // stride is in pixels, linesize is in bytes (stride x4). We need bytes
 
-  sws_scale(sws, frame->data, frame->linesize, 0, frame->height, &swsBuffer, &dst_linesize);
+  int height = frame->height;
 
-  pb->imageRect(rect, swsBuffer, stride);
+  // The server may send a frame with a height greater than that of the rect,
```
CendioOssman, line 228: I understand why this is desirable, but it is unfortunately not something mentioned in the specification for this extension. So it can be argued that it is a violation of the protocol to do so.

So at the minimum, we should make it clear that this is legal. And to do so, I think we need some background research of the current implementations of this extensions:

1. Which servers are already sending H.264 frames larger than the RFB rect
2. How are current and historical clients tolerating this. Evidently TigerVNC doesn't. How are other implementations?

common/rfb/H264LibavDecoderContext.cxx
```diff
@@ -222,9 +222,19 @@ void H264LibavDecoderContext::decode(const uint8_t* h264_in_buffer,
...
+  // but we don't want to write outside of the rect.
+  // The server may choosed to do this due to hardware encoder constraints.
+  if (height > rect.height())
+    height = rect.height();
```
CendioOssman, line 232: Why not width as well?

CendioOssman (changes requested):
Thanks for your PR.

There are two fixes in this commit, which are conceptually quite different. One is also a pure bug, whilst the other is more controversial. Please split them to separate commits.

common/rfb/H264LibavDecoderContext.cxx
any1: Because this fixed the crashing and adding the width too was a bit more work and it was time to go to bed.

any1: Actually, I think that this might be handled by the linesize arguments to `sws_scale`, but I'd have to take a closer look at the implementation of `sws_scale` to be sure.

common/rfb/H264LibavDecoderContext.cxx
CendioOssman: Okay, sounds like we might get away with an adjustment. Can you submit a clarification for rfbproto?

In that spirit, perhaps best specify that a smaller frame is invalid? Otherwise, it would need to be specific what to do with the missing pixels.

common/rfb/H264LibavDecoderContext.cxx
CendioOssman: Please check. It would be very odd to specify that height can be larger, but not width.

any1: Well, `sws_scale` is quite a complex beast. It can go into multiple code paths depending on the pixel formats and there are different implementations for different platforms.

It's probably safest to go by the guarantees given by the API, which are none.

A better approach would be to make the scaling buffer as wide as the frame and handle the cropping in `pb->imageRect` instead. E.g. by setting the stride.

common/rfb/H264LibavDecoderContext.cxx
any1: Done. I agree that smaller frames should be invalid.

common/rfb/H264LibavDecoderContext.cxx
CendioOssman: Either way is fine by me.

CendioOssman (changes requested):
I guess the Windows backend also needs to be able to cope with oversized frames?

common/rfb/H264LibavDecoderContext.cxx
any1: Current iteration takes care of width, height and alignment.

any1:
> I guess the Windows backend also needs to be able to cope with oversized frames?

Sure, but I don't use Windows and I do not have any computer set up with Windows, so I'm not your guy for that.

common/rfb/H264LibavDecoderContext.cxx
cgm999: Well I do not think the current proposed version resolves the swsBuffer size issue that I see. 

sws_scale wants that there is room after end of the buffer(I found this by experimentation) , which we need to ignore .. I assume as you said related to SIMD code that is optimized for CPU/platforms and it needs to write in chunks of data , so it needs to have room at end for that garbage data.

new proposed code:

  int dstLineSize = FFALIGN(frame->width * dstBytesPerPixel , av_cpu_max_align());                                                                             
  swsBuffer_test.resize(dstLineSize * frame->height, 0);                                                                                                       
  printf("swsBuffer_test.capacity=%d\n",swsBuffer_test.capacity());                                                                                            

result: swsBuffer_test.capacity=4202496 
4202496 is exactly 1368*4*768 (1368x768 is my frame)

When I allocate the buffer with av_frame_get_buffer on a AVFrame , the buffer have extra room (visible with mtrace/muntrace, I was not able to find a better way to figure size of allocated buffer ):

         mtrace();                                                                                                                                             
         av_frame_get_buffer( pic_out, av_cpu_max_align() );                                         
         muntrace();                                                                                                                                           

LD_PRELOAD=/usr/lib/libc_malloc_debug.so MALLOC_TRACE=/tmp/x ./test
cat /tmp/x
= Start
@ /usr/lib/libavutil.so.59:(av_malloc+66)[0x499c6] + 0x741a595c8040 0x402080

0x402080 == 1368*4*768 + 128  . 
The issue that I am facing is .. sws_scale writes garbage data after end of the buffer so we  need some extra room for that. Not sure how much space we need (I guess it depends on CPU/etc) ,we could create a frame  with av_frame_alloc  / av_frame_get_buffer and use that as the dest buffer for sws_scale then free it ? Or just allocate that once and keep it untill the frame size changes?

CendioOssman:
Windows needs to be figured out, as I'm not ready to claim compatibility with neatvnc if it doesn't work on all platforms. :/

@mmozeiko, @xornet-sl, do you think you can have a look? There looks to be some cropping code already in place?

common/rfb/H264LibavDecoderContext.cxx
any1: Yeah, using `AVFrame` for the scaling buffer is probably a good idea as it handles the alignment for us.

The calculation of the frame size a slightly more complex than what we have, but it adds something called `plane_padding`, which is `(16 + 16) * 4 = 128`.

See: https://ffmpeg.org/doxygen/trunk/frame_8c_source.html#l00169

There is one problem with the current approach (and using `AVFrame`). Technically, after the stride is aligned, it might not be divisible by bytes-per-pixel. The solution for that would be to either align on the width or to add a new overload for `imageRect` that takes stride in bytes rather than row pixel count.

any1:
> There looks to be some cropping code already in place?

That code is manually applying the cropping region specified by the SPS. It does not seem to take into account the size of the rect.

any1:
Come to think of it, and looking at the linux kernel implementation of the `bcm2835-codec` driver, it does actually look like it's setting the cropping region. I haven't verified this by actually looking at the SPS though.

Thus, this is really just a matter of proper alignment and padding, and not misbehaving encoders, although they should definitely be guarded against.

common/rfb/H264LibavDecoderContext.cxx
cgm999: At a glance it looks good to me. Will test it a bit on my setup ..

Also to get rid of the memleak of sws from sws_getContext  , need to freeContext(sws) after sws_scale* (valgrind shows it)

cgm999: So far testing looks good. doing more testing

cgm999: OK so this sws_freeContext(sws) needs to be done in H264LibavDecoderContext::freeCodec()

cgm999: So far no crash in my testing. However there is some big memleak(when h264 resets),trying to figure why

any1: I added a commit to clean up `sws`.

cgm999: One more thing, for sws_getCachedContext , we passing flags 0 , however in ffmpeg code that will be converted to SWS_BICUBIC , which in turn means every time a new SwsContext will be allocated unless we pass flags SWS_BICUBIC like this..

  sws = sws_getCachedContext(sws, frame->width, frame->height, avctx->pix_fmt,                                                                                 
                             frame->width, frame->height, AV_PIX_FMT_RGB32,                                                                                    
                             SWS_BICUBIC, nullptr, nullptr, nullptr);

any1: Well, I'd say that it shouldn't be using any interpolation at all; just use nearest neighbour, but it seems that this is not available as an option, so I suppose `SWS_FAST_BILINEAR` would be the least bad option.

In any case, this is a separate issue.

cgm999: Right it is a separate issue,however right now we are using sws_getCachedContext and it does not work as expected, is always going to return a new context and free previous one because 0 is translated to 4 (SWS_BICUBIC) inside sws_init_single_context . 

I understand this needs to be fixed in a diff PR.

BTW So far all my testing shows good results. I see there is an issue with the build(sws_scale_frame was introduced in 2021 Sep based on ffmpeg doc/APIchanges)

cgm999:
@CendioOssman what is the next step to merge the PR ?  I think this will benefit for some users (I have PiKVM in mind now)

CendioOssman:
It's not currently building properly, so that would need to be fixed. Otherwise, I don't think there are any open issues.

any1:
> It's not currently building properly, so that would need to be fixed. Otherwise, I don't think there are any open issues.

We can disable h264 depending on whether `av_scale_frame` exists or not. Would that be good enough?

CendioOssman:
I thought we already checked for swscale? Or is that some feature that requires a very new ffmpeg?

CendioOssman:
That sounds like it might not be included in EPEL 8 and Ubuntu 20.04. So if it's possible to support those as well, then it would be appreciated.

any1:
This feature was added in version 1.13.0, so I suppose that this fix should be backported to 1.13.

Jammy has 1.12.0: https://launchpad.net/ubuntu/jammy/+package/tigervnc-viewer

Older version of Ubuntu are bound to have older releases. The Ubuntu versions on which the build fails are not going to receive this update unless a user wants to compile manually.

I don't see the point in chasing old software, but if that's how you want things, I can make this work with older ffmpeg. Some time might pass until I find the time to do so.

any1:
It should compile on those old Ubuntus now.

I also have a colour space fix lined up for this, but I can make a separate PR for that when this is merged.

cgm999:
Looks good to me.

Issue #1786, opened 2024-07-26
VNC service does not start if user does not have a shell set in /etc/passwd


casantos:
**Describe the bug**
The VNC service does no start if the configured user has an empty shell field in the password entry, either in /etc/passwd or via AD/IPA/NIS. An empty shell field in the password file is valid, although not common. In this case, /bin/sh is used, as documented in the passwd(5) manual page.

**To Reproduce**
1. Add a local user and use vipw to empty the last field in the password file:
 ```
# getent passwd testuser
testuser:x:1000:1000:Test User:/home/testuser:
```
2. Configure a vnc service using that local user
3. Attempt tp start the service

**Expected behavior**
The service should start normally.

**Client (please complete the following information):**
 - OS: Red Hat Enterprise Linux 9.4
 - VNC client: TigerVNC (vncviewer)
 - VNC client version: 1.13.1
 - Client downloaded from: Linux distribution

**Server (please complete the following information):**
 - OS: Red Hat Enterprise Linux 8.10
 - VNC server: TigerVNC
 - VNC server version: 1.14.0
 - Server downloaded from: built from source
 - Server was started using: `systemctl start vncservice@:4`]

**Additional context**
Actual problem was reported by a customer that uses LDAP but the problem is easy to reproduce with a local user.
The problem happens because the SHELL environment variable is set by vncsession before running the vncserver script:
```
struct passwd *pwent;
...
pwent = getpwnam(username);
...
setenv("SHELL", *pwent->pw_shell);
```
So if the shell field is empty the script will run with an empty SHELL environment variable. In order to prevent this error, the code should be like this:
```
setenv("SHELL", *pwent->pw_shell != '\0' ? pwent->pw_shell : "/bin/sh", 1);
```

PR #1787, opened 2024-07-26
vncsession: use /bin/sh if the user shell is not set

casantos:
An empty shell field in the password file is valid, although not common. Use /bin/sh in this case, as documented in the passwd(5) man page, since the vncserver script requires a non-empty SHELL environment variable.

Fixes issue #1786.


CendioOssman:
Ooops. Surprised this wasn't noticed by your enterprise users earlier. :)

PR #1792, opened 2024-07-29
Add option allowing to connect only the user owning the running session

grulja:
Checks, whether the user who is trying to authenticate is already logged into the running session in order to allow or reject the connection. This is expected to be used with 'plain' security type in combination with 'PlainUsers=*' option allowing everyone to connect to the session.

### DIFF
--- a/common/rfb/VNCServerST.cxx
+++ b/common/rfb/VNCServerST.cxx
@@ -720,13 +720,6 @@ void VNCServerST::queryConnection(VNCSConnectionST* client,
     return;
   }
 
-  // - Are we configured to do queries?
-  if (!rfb::Server::queryConnect &&
-      !client->getSock()->requiresQuery()) {
-    approveConnection(client->getSock(), true, nullptr);
-    return;
-  }
-
   // - Does the client have the right to bypass the query?
   if (client->accessCheck(AccessNoQuery))
   {
--- a/unix/x0vncserver/XDesktop.cxx
+++ b/unix/x0vncserver/XDesktop.cxx
@@ -33,6 +33,7 @@
 #include <network/Socket.h>
 
 #include <rfb/ScreenSet.h>
+#include <rfb/ServerCore.h>
 
 #include <x0vncserver/XDesktop.h>
 
@@ -323,6 +324,13 @@ void XDesktop::queryConnection(network::Socket* sock,
                                const char* userName)
 {
   assert(isRunning());
+
+  // - Are we configured to do queries?
+  if (!rfb::Server::queryConnect &&
+      !sock->requiresQuery()) {
+    server->approveConnection(sock, true, nullptr);
+    return;
+  }
 
   // Someone already querying?
   if (queryConnectSock) {
--- a/unix/xserver/hw/vnc/XserverDesktop.cc
+++ b/unix/xserver/hw/vnc/XserverDesktop.cc
@@ -55,6 +55,11 @@
 #include "XorgGlue.h"
 #include "vncInput.h"
 
+#if HAVE_SYSTEMD_DAEMON
+#  include <pwd.h>
+#  include <systemd/sd-login.h>
+#endif
+
 extern "C" {
 void vncSetGlueContext(int screenIndex);
 void vncPresentMscEvent(uint64_t id, uint64_t msc);
@@ -72,6 +77,15 @@ core::IntParameter
                       "connection' dialog before rejecting the "
                       "connection", 10, 0, INT_MAX);
 
+#ifdef HAVE_SYSTEMD_DAEMON
+core::BoolParameter approveLoggedUserOnly
+("ApproveLoggedUserOnly",
+ "Approve only the user who is currently logged into the session."
+ "This is expected to be combined with 'plain' security type and with "
+ "'PlainUsers=*' option allowing everyone to connect to the session."
+ "Default is off.",
+  false);
+#endif
 
 XserverDesktop::XserverDesktop(int screenIndex_,
                                std::list<network::SocketListener*> listeners_,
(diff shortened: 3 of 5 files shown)


grulja:
Marking this as a draft as I would like to get some feedback on this first. This feature has been requested by our customer. Their workflow is to configure GDM to accept XDMCP in `/etc/gdm/custom.conf`:
```
[xdmcp]
Enable=true
```
And create and start a systemd unit file as `/etc/systemd/system/xvnc.service` with:
```
[Unit]
Description=VNC XDMCP Daemon

[Service]
ExecStart=-/usr/bin/Xvnc -query localhost -once -Log *:stderr:100 rfbport=5900 securitytypes=TLSPlain PlainUsers=* pam_service=gdm-password AlwaysShared=1
User=root
StandardError=syslog

[Install]
WantedBy=multi-user.target
```
This now allows everyone to connect and log in GDM, however, they want to have an option to only authenticate the same user who logged in through GDM.

There are some things I'm not sure about:
1) Name of this option
2) Where this is now implemented. I think it would make more sense to be implemented in `XServerDesktop` (maybe), but this is tricky, because it uses Makefile and I don't know how to add this optional systemd support. From there I could also get the display number and also use `sd_session_get_display()` for an additional check.

CendioOssman:
I'm afraid I don't quite understand the user case. Even more so given the fact that modern systems (of which Red Hat has been a driving force) don't work properly if you are logged in locally and via TigerVNC at the same time.

I would more have expected a patch doing the opposite, blocking users that are already logged in elsewhere.

Can you explain more why the customer wants this policy, and what they expect to actually happen?

grulja:
I probably failed at explanation. The workflow is:

1) Let's say we have a system with users `vnc1` and `vnc2`.
2) Configure the system as mentioned above
3) Connect with `vncviewer -SecurityTypes TLSPlain server`
4) Authenticate as user `vnc1` and log into e.g. GNOME session using GDM
5) Connect again with `vncviewer -SecurityTypes TLSPlain server`
6) Try to authenticate as user `vnc2`, where this user will be able to connect and see the session of user `vnc1`.

The goal is to have an option that will not allow user `vnc2` to be connected (as a viewer) if there is already a different user logged into a graphical session.

CendioOssman:
Ah, this is not an inetd setup. That's usually where XDMCP is used.

That complicates things.

I don't think the proposed implementation scales. You'd be blocked by any other user login, right? Ideally, we should check the status of this session only.

Does the XDMCP-protocol provide any feedback we might use?

grulja:
> I don't think the proposed implementation scales. You'd be blocked by any other user login, right? Ideally, we should check the status of this session only.

Right, any other running session would be also blocking you.

> Does the XDMCP-protocol provide any feedback we might use?

I don't know, I will look and ask around. That's why I mentioned use of `sd_session_get_display()` which we could use to check this session only, but I don't know how to get the display number in `VNCServerST`. At least this is how I think it should be possible to check the session we want.

CendioOssman:
The core is not aware of X11 things. And it shouldn't be. So this sounds very much like a frontend thing that should be in Xvnc somewhere. It might be `XServerDesktop`, as you suggested, but might be elsewhere. Probably depends on which part has access to the best information.

grulja:
> The core is not aware of X11 things. And it shouldn't be. So this sounds very much like a frontend thing that should be in Xvnc somewhere. It might be `XServerDesktop`, as you suggested, but might be elsewhere. Probably depends on which part has access to the best information.

I was looking into adding it to `XServerDesktop`, where I think it makes more sense as it's X11-aware, but I struggled with the use of Makefile there and the optional use of systemd library. I guess it can be done through CMake modifying the makefile similar to `config.h.in`. If you think it makes more sense to be there, I will give it a try.

CendioOssman:
CMake cannot do much about Xorg's autoconf build. If you need to adjust things there, I'm afraid you'll need to update the patch files that modifes Xorg's `configure.ac`. There is some systemd support there already. Could be enough?

grulja:
What about something like this? To enable this, Xvnc has to be built with `--enable-systemd-logind` and `--enable-config-udev` which is a dependency for systemd-logind support in Xorg (unless we patch it).

CendioOssman:
Slightly better. But I'm concerned that this is fundamentally built on making assumptions about what the display manager will do. I would prefer if we kept to things we have control over. Or push those decisions out of TigerVNC.

Did you examine what you can get from XDMCP? I'm thinking one model would be that Xvnc keeps track of which user authenticated. Then locks that in as long as XDMCP indicates that a session is active. Once the session ends, Xvnc can forget about the user and go back to accepting everyone.

Another approach is to delegate this to something external. Perhaps using the query system. Write a custom program there that can implement the logic of which connections to accept or not.

grulja:
> Did you examine what you can get from XDMCP? I'm thinking one model would be that Xvnc keeps track of which user authenticated. Then locks that in as long as XDMCP indicates that a session is active. Once the session ends, Xvnc can forget about the user and go back to accepting everyone.
>

Not yet.
 
> Another approach is to delegate this to something external. Perhaps using the query system. Write a custom program there that can implement the logic of which connections to accept or not.

Something similar to `vncconfig` and `QueryConnect` option in `Xvnc`? Can `vncconfig` be extended with this functionality?  Or would you prefer a separate one, but implemented in a similar way as `vnconfig`.

CendioOssman:
> Something similar to `vncconfig` and `QueryConnect` option in `Xvnc`?

Yes. I'm hoping there is enough information there to act on.

> Can `vncconfig` be extended with this functionality? Or would you prefer a separate one, but implemented in a similar way as `vnconfig`.

Might be best to make this a separate tool. I'm still unsure if it's too niche to include in the standard TigerVNC. Being a separate tool would definitely make it easier to include.

(Related, vncconfig should probably be split in to multiple tools)

Neustradamus:
To follow this PR :)

grulja:
> > Did you examine what you can get from XDMCP? I'm thinking one model would be that Xvnc keeps track of which user authenticated. Then locks that in as long as XDMCP indicates that a session is active. Once the session ends, Xvnc can forget about the user and go back to accepting everyone.
> 

I've been looking into this and in the XDMCP implementation I can see it regularly checks whether the session is still alive, BUT I don't know how to propagate or connect this with the code I have in `XServerDesktop.cc`. One thing that could work would be to just add an additional public function that will simply return the state of XDMCP and use a timer to check this state regularly. Another approach might be to add a hook (pass a function pointer) which we can call once the session is no longer active and remove the locked user at that point. 

> > Can `vncconfig` be extended with this functionality? Or would you prefer a separate one, but implemented in a similar way as `vnconfig`.
> 
> Might be best to make this a separate tool. I'm still unsure if it's too niche to include in the standard TigerVNC. Being a separate tool would definitely make it easier to include.
> 

I've been also looking into this and it would be way too much work for something like this. I would prefer spend less time on this. I guess it's fine if you are not going to accept this upstream, still at least some feedback and review will be helpful. Even though it's not the way I prefer, we can at least carry this as a downstream patch.

casantos:
I was able (somewhat) to solve the problem by means of this PAM module:

~~~
/*
 * pam_usermatch.c
 *
 * Author: Carlos Santos <casantos@redhat.com>
 *
 * Copyright 2025 Red Hat, Inc.
 *
 * Redistribution and use in source and binary forms, with or without
 * modification, are permitted provided that the following conditions
 * are met:
 * 1. Redistributions of source code must retain the above copyright
 *    notice, and the entire permission notice in its entirety,
 *    including the disclaimer of warranties.
 * 2. Redistributions in binary form must reproduce the above copyright
 *    notice, this list of conditions and the following disclaimer in the
 *    documentation and/or other materials provided with the distribution.
 * 3. The name of the author may not be used to endorse or promote
 *    products derived from this software without specific prior
 *    written permission.
 *
 * Build and install it with:
 *
 *   $ gcc -c -fPIC -o pam_usermatch.o pam_usermatch.c
 *	 $ gcc -shared -o pam_usermatch.so -l pam -l selinux pam_usermatch.o
 *   $ sudo cp pam_usermatch.so /usr/lib/security    (32-bit systems)
 *   $ sudo cp pam_usermatch.so /usr/lib64/security  (64-bit systems)
 *
 * Ensure that the pam service file loads the module, e.g.
 *
 *   #%PAM-1.0
 *   auth  substack  system-auth
 *   auth  include   postlogin
 *   auth  required  pam_usermatch.so
 *
 */

#include <pwd.h>
#include <string.h>
#include <stdlib.h>
#include <syslog.h>
#include <security/pam_ext.h>
#include <security/pam_modules.h>
#include <systemd/sd-login.h>

static void report_error(pam_handle_t *pamh, const char *error)
{
    pam_syslog(pamh, LOG_ERR, error);
    pam_error(pamh, error);
}

PAM_EXTERN int pam_sm_authenticate(pam_handle_t *pamh, int flags, int argc, const char **argv)
{
    const void *display;
    struct passwd *pw;
    int i, retval, num_sessions;
    const void *user;
    char **sessions;
    uid_t uid;
    char *attrib;

    retval = pam_get_item(pamh, PAM_XDISPLAY, &display);
    if (retval != PAM_SUCCESS || display == NULL || *(const char*) display == '\0') {
        display = getenv("DISPLAY");
        if (display == NULL || *(const char*) display == '\0') {
            report_error(pamh, "Could not obtain process display.");
            return PAM_AUTH_ERR;
        }
    }

    retval = pam_get_item(pamh, PAM_USER, &user);
    if (retval != PAM_SUCCESS || user == NULL || *(const char*) user == '\0') {
        report_error(pamh, "Unknown user.");
        return PAM_USER_UNKNOWN;
    }

    pw = getpwnam(user);
    if (pw == NULL) {
        report_error(pamh, "User not found.");
        return PAM_USER_UNKNOWN;
    }

    num_sessions = sd_get_sessions(&sessions);
    if (num_sessions < 0) {
        report_error(pamh, "Could not obtain current login sessions.");
        return PAM_AUTH_ERR;
    }

    if (num_sessions == 0 || sessions == NULL) {
        report_error(pamh, "No user session found.");
        return PAM_SUCCESS;
    }

    retval = PAM_SUCCESS;
    for (i = 0; i < num_sessions; i++) {

        if (sd_session_get_type(sessions[i], &attrib) < 0) {
            report_error(pamh, "Could not obtain session type.");
            retval = PAM_AUTH_ERR;
            break;
        }
        if (strcmp(attrib, "x11") != 0) {
            free(attrib);
            continue;
        }
        free(attrib);

        if (sd_session_get_display(sessions[i], &attrib) < 0) {
            report_error(pamh, "Could not obtain session display.");
            retval = PAM_AUTH_ERR;
            break;
        }
        if (strcmp(attrib, display) != 0) {
            free(attrib);
            continue;
        }
        free(attrib);

        if (sd_session_get_class(sessions[i], &attrib) < 0) {
            report_error(pamh, "Could not obtain session class.");
            retval = PAM_AUTH_ERR;
            break;
        }
        if (strcmp(attrib, "user") != 0) {
            free(attrib);
            continue;
        }
        free(attrib);

        if (sd_session_get_state(sessions[i], &attrib) < 0) {
            report_error(pamh, "Could not obtain session state.");
            retval = PAM_AUTH_ERR;
            break;
        }
        if (strcmp(attrib, "active") != 0) {
            free(attrib);
            continue;
        }
        free(attrib);

        if (sd_session_get_uid(sessions[i], &uid) < 0) {
            report_error(pamh, "Could not obtain session UID.");
            break;
        }
        /* Found a session with type "x11",same display, class "user" and state "active". */
        if (uid != pw->pw_uid) {
            /* Another user is already logged in. */
            report_error(pamh, "Another user already onws the running session.");
            retval = PAM_AUTH_ERR;
            break;
        }
    }

    if (sessions) {
        for (int i = 0; sessions[i]; i++) {
            free(sessions[i]);
        }

        free(sessions);
    }

    return retval;
}

PAM_EXTERN int pam_sm_setcred(pam_handle_t *pamh, int flags, int argc, const char **argv)
{
    return PAM_SUCCESS;
}
~~~
I was unable to make Xvnc set the PAM_XDISPLAY item or the DISPLAY environment variable, however. As a workaround, I've set the environment variable in the systemd unit, `/etc/systemd/system/vnc-xdmcp.service`:
~~~
[Unit]
Description=VNC XDMCP Daemon

[Service]
ExecStart=-/usr/bin/Xvnc :0 -query localhost rfbport=5900 securitytypes=TLSPlain PlainUsers=* PAMService=tigervnc AlwaysShared=1
Environment=DISPLAY='::1:0'
User=root
StandardError=syslog
Restart=on-success

[Install]
WantedBy=multi-user.target
~~~
Notice that due to using `-query` the display will become either `127.0.0.1:0` or `::1:0`, depending on wether IPv6 is used or not.

My `/etc/pam.d/tigervnc`is a slightly modified version of `/etc/pam.d/login`:
~~~
#%PAM-1.0
auth       substack     system-auth
auth       include      postlogin
auth       required     pam_usermatch.so
account    required     pam_nologin.so
account    include      system-auth
password   include      system-auth
# pam_selinux.so close should be the first session rule
session    required     pam_selinux.so close
session    required     pam_loginuid.so
# pam_selinux.so open should only be followed by sessions to be executed in the user context
session    required     pam_selinux.so open
session    required     pam_namespace.so
session    optional     pam_keyinit.so force revoke
session    include      system-auth
session    include      postlogin
-session   optional     pam_ck_connector.so
~~~

casantos:
@CendioOssman @grulja please check https://github.com/TigerVNC/tigervnc/pull/1957

CendioOssman:
With #1957 merged, I guess this can be closed?

PR #1796, opened 2024-07-30
vncsession-start: Add place admin can drop in args

jcpunk:
If the system service manager could benefit from calling `vncsession` with `-D`, then there should be an easy way to make it the default for folks.

In this way it is easy for people to quickly select what they would prefer.  They'll still need to set an override for `vncserver@.service` -> `Type=simple` if they want to use systemd, but this could be used to gather more information.

### DIFF
--- a/unix/vncserver/vncsession-start.in
+++ b/unix/vncserver/vncsession-start.in
@@ -19,11 +19,18 @@
 #
 
 USERSFILE="@CMAKE_INSTALL_FULL_SYSCONFDIR@/tigervnc/vncserver.users"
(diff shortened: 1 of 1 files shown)


CendioOssman:
You're not supposed to configure things via arguments to vncsession, hence why we haven't built any mechanism for it.

What problem are you trying to solve?

jcpunk:
This is really the first step towards making https://github.com/TigerVNC/tigervnc/pull/1799 something I can configure.

CendioOssman:
Since #1799 is not the direction we want to take things, I'm afraid there is no current use case for this proposed change.

PR #1797, opened 2024-07-31
Include DLL when compiling installer in MSYS2

lurenJBD:
When building in Windows using MSYS2, some DLLs were missing during the packaging of the installer. This modification will locate the relevant DLLs and package them together with the installer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -347,6 +347,8 @@ add_subdirectory(common)
 
 if(WIN32)
   add_subdirectory(win)
+  find_path(TEMP_PATH libunistring-5.dll)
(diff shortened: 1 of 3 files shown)


lurenJBD:
Refer to this issue https://github.com/TigerVNC/tigervnc/issues/1798 for detailed information.

CendioOssman (changes requested):
If we want to start to include DLLs in the installer, then I'm afraid this is not a robust way to do so.

The list of files specified here may be right for your system, but is not right for other systems.

Something dynamic would need to be put in place in that case that detects which files are needed, and where they are located.

PR #1807, opened 2024-08-12
Fix typos

algonell:
Thinking out loud, I would also rewrite this:

"taking the values of the local port number, the remote host, the port number on the remote host, and the gateway machine respectively."

Into:

"taking values of local port number, remote host, port number on the remote host, and gateway machine respectively."


CendioOssman (approved):
Thanks!

PR #1808, opened 2024-08-16
CSecurityTLS: change the variable that x509 authentication CA and CRL file from global to local

KangLin:
CSecurityTLS: change the variable that x509 authentication CA and CRL file from global to local.
Use different CA files for multiple connections.

### DIFF
--- a/common/rfb/CConnection.h
+++ b/common/rfb/CConnection.h
@@ -121,6 +121,13 @@ namespace rfb {
 
     // Methods to be overridden in a derived class
 
+    /*!
+     * \brief Get x509 authentication CA and CRL file, the file format is pem.
+     * \param ca: certificate authority, the file format is pem.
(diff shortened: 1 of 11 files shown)


CendioOssman:
I'm unsure about this approach. Feels like we're papering over a more fundamental problem. User's shouldn't have to fiddle with this setting every time they change which server they connect to.

Does GnuTLS read just a single certificate from that file? Maybe the correct usage model is to add all your trusted certificates to it, rather than having many files?

CendioOssman:
I had a look at the documentation, and it claims that you can add multiple certificates to it:

https://www.gnutls.org/manual/html_node/Core-TLS-API.html#gnutls_005fcertificate_005fset_005fx509_005ftrust_005ffile

Is that sufficient for your use case?

Neustradamus:
To follow the @KangLin PR :)

PR #1813, opened 2024-08-19
Apple only: exit FL_Window fullscreen if active in order to minimize.

rudironsonijr:


LMattsson:
Thank you for both finding and making a fix for this issue, it doesn't seem to have been reported previously.

I looked into this in detail, here's a summary:

Confirmed the bug with TigerVNC 1.14.0 on macOS 14.5. While in fullscreen, nothing happens when I try to minimise from the F8-menu. In windowed mode, minimise works fine. Also confirmed that minimise work after the fix.

One theory was that macOS native fullscreen was involved in this issue, since TigerVNC doesn't work well with that. I tested a few different things that gave me conflicting information:

 1) Tested building vncviewer with macOS native fullscreen off, but I could still see the minimise bug. One indication that native fullscreen is off is that the green button in the top bar of the window shows a '+' when hovered, instead of 2 arrows pointing away from each other. Could see that the green button had shifted to a '+' for this new build.

 2) Tested building vncviewer on another machine with FLTK v. 1.3.7 and macOS SDK v. 10.6. When running this build on the original machine, minimise worked when fullscreen was entered from the F8-menu. When I entered fullscreen natively (clicking the green button with 2 arrows in top bar), minimise did not work. 

Since minimise did work in case 2, this fix seems to act more like a workaround than targeting the core issue. As a side note, the vncviewer ends up in windowed mode after un-minimising the window, not fullscreen mode, but this is still an improvement from the current behaviour.

It's probably better to wait finding the core issue until we get more clues about it. I'll go ahead and merge this.

PR #1824, opened 2024-08-27
vncviewer: add missing include for struct timeval

0-wiz-0:
The header file uses `struct timeval` but doesn't include the necessary header.
This breaks the build on NetBSD.

### DIFF
--- a/vncviewer/BaseTouchHandler.h
+++ b/vncviewer/BaseTouchHandler.h
@@ -22,6 +22,8 @@
 
 #include "GestureEvent.h"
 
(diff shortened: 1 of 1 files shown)


CendioOssman (changes requested):
My man page claims that it is in `sys/time.h`, not `time.h`. Does that work for you as well?

0-wiz-0:
Sure, that's fine too. Updated the patch.

CendioOssman:
Great! Thanks!

PR #1826, opened 2024-08-28
Add support for forward and back mouse buttons

CendioHalim:
This is a continuation of the work initially done by [manny33](https://github.com/manny33) in #1711:
* https://github.com/TigerVNC/tigervnc/pull/1711

This PR intends to add support for forward and back mouse buttons by extending RFB with the pseudo-encoding `ExtendedMouseButtons`, which is IANA registered and has an open draft PR on rfbproto:
* https://github.com/rfbproto/rfbproto/pull/57

In its current state, this PR has a working implementation that works on Windows, Mac and Linux (with some issues that are being worked on).

I'm quite content with the RFB implementation on the VNC side, but the client-side implementation in [Viewport.cxx](https://github.com/TigerVNC/tigervnc/blob/master/vncviewer/Viewport.cxx) is currently being re-written quite extensively so that **all** mouse events are handled with native code instead of using a combination of FLTK and native code.

This is still a work in progress, and commits will be cleaned up / rebased.

### DIFF
--- a/common/rfb/CMsgWriter.cxx
+++ b/common/rfb/CMsgWriter.cxx
@@ -22,6 +22,7 @@
 #endif
 
 #include <stdio.h>
+#include <assert.h>
 
 #include <rdr/OutStream.h>
 #include <rdr/MemOutStream.h>
@@ -173,18 +174,47 @@ void CMsgWriter::writeKeyEvent(uint32_t keysym, uint32_t keycode, bool down)
 }
 
 
-void CMsgWriter::writePointerEvent(const Point& pos, uint8_t buttonMask)
+void CMsgWriter::writePointerEvent(const Point& pos, uint16_t buttonMask)
 {
   Point p(pos);
+  bool extendedMouseButtons;
+
   if (p.x < 0) p.x = 0;
   if (p.y < 0) p.y = 0;
   if (p.x >= server->width()) p.x = server->width() - 1;
   if (p.y >= server->height()) p.y = server->height() - 1;
 
+  /* The highest bit in buttonMask is never sent to the server */
+  assert(!(buttonMask & 0x8000));
+
+  /* Only send extended pointerEvent message when needed */
+  extendedMouseButtons = buttonMask & 0x7f80;
+
   startMsg(msgTypePointerEvent);
-  os->writeU8(buttonMask);
-  os->writeU16(p.x);
-  os->writeU16(p.y);
+  if (server->supportsExtendedMouseButtons && extendedMouseButtons) {
+    int higherBits;
+    int lowerBits;
+
+    higherBits = (buttonMask  >> 7) & 0xff;
+    assert(!(higherBits & 0xfc)); /* Bits 2-7 are reserved */
+
+    lowerBits = buttonMask & 0x7f;
+    lowerBits |= 0x80; /* Set marker bit to 1 */
+
+    os->writeU8(lowerBits);
+    os->writeU16(p.x);
+    os->writeU16(p.y);
+    os->writeU8(higherBits);
+  } else {
+    /* Marker bit must be set to 0, otherwise the server might confuse
+     * the marker bit with the highest bit in a normal PointerEvent
+     * message.
+    */
+    buttonMask &= 0x7f;
+    os->writeU8(buttonMask);
+    os->writeU16(p.x);
+    os->writeU16(p.y);
+  }
   endMsg();
 }
 
--- a/common/rfb/SMsgReader.cxx
+++ b/common/rfb/SMsgReader.cxx
@@ -272,11 +272,32 @@ bool SMsgReader::readKeyEvent()
 
 bool SMsgReader::readPointerEvent()
 {
+  int mask;
+  int x;
+  int y;
+
   if (!is->hasData(1 + 2 + 2))
     return false;
-  int mask = is->readU8();
-  int x = is->readU16();
-  int y = is->readU16();
+
+  is->setRestorePoint();
+
+  mask = is->readU8();
+  x = is->readU16();
+  y = is->readU16();
+
+  if (handler->client.supportsExtendedMouseButtons() && mask & 0x80 ) {
+    int highBits;
+    int lowBits;
+
+    if (!is->hasDataOrRestore(1))
+      return false;
+
+    highBits = is->readU8();
+    lowBits = mask & 0x7f; /* Clear marker bit */
+    mask = (highBits << 7) | lowBits;
+  }
+
+  is->clearRestorePoint();
   handler->pointerEvent(Point(x, y), mask);
   return true;
 }
--- a/unix/xserver/hw/vnc/vncInput.c
+++ b/unix/xserver/hw/vnc/vncInput.c
@@ -50,7 +50,7 @@ extern const unsigned int code_map_qnum_to_xorgevdev_len;
 extern const unsigned short code_map_qnum_to_xorgkbd[];
 extern const unsigned int code_map_qnum_to_xorgkbd_len;
 
-#define BUTTONS 7
+#define BUTTONS 9
 
 DeviceIntPtr vncKeyboardDev;
 DeviceIntPtr vncPointerDev;
@@ -206,6 +206,23 @@ static int vncPointerProc(DeviceIntPtr pDevice, int onoff)
 		btn_labels[4] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_WHEEL_DOWN);
 		btn_labels[5] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_LEFT);
 		btn_labels[6] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_RIGHT);
+
+		/*
+		 * The labels BTN_LABEL_PROP_BTN_SIDE and BTN_LABEL_PROP_BTN_EXTRA
+		 * represent the side buttons on mice that are typically used to
+		 * navigate back/forward respectively in web browsers.
+		 *
+		 * In X11, these labels are mapped to the BTN_SIDE and BTN_EXTRA
+		 * input codes, which are mapped in the Linux HID driver. These
+		 * are not to be confused with the BTN_FORWARD and BTN_BACK input
+		 * codes, which some applications also use for back/forward
+		 * navigation.
+		 *
+		 * It seems like most mice have their side buttons mapped to
+		 * BTN_SIDE and BTN_EXTRA.
+		 */
+		btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_SIDE);
+		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_EXTRA);
 
 		axes_labels[0] = XIGetKnownProperty(AXIS_LABEL_PROP_REL_X);
 		axes_labels[1] = XIGetKnownProperty(AXIS_LABEL_PROP_REL_Y);
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -600,22 +600,22 @@ int Viewport::handle(int event)
   case FL_MOUSEWHEEL:
     buttonMask = 0;
     if (Fl::event_button1())
-      buttonMask |= 1;
+      buttonMask |= 1 << 0;
     if (Fl::event_button2())
-      buttonMask |= 2;
+      buttonMask |= 1 << 1;
     if (Fl::event_button3())
-      buttonMask |= 4;
+      buttonMask |= 1 << 2;
 
     if (event == FL_MOUSEWHEEL) {
       wheelMask = 0;
       if (Fl::event_dy() < 0)
-        wheelMask |= 8;
+        wheelMask |= 1 << 3;
       if (Fl::event_dy() > 0)
-        wheelMask |= 16;
+        wheelMask |= 1 << 4;
       if (Fl::event_dx() < 0)
-        wheelMask |= 32;
+        wheelMask |= 1 << 5;
       if (Fl::event_dx() > 0)
-        wheelMask |= 64;
+        wheelMask |= 1 << 6;
 
       // A quick press of the wheel "button", followed by a immediate
       // release below
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -606,6 +606,20 @@ int Viewport::handle(int event)
     if (Fl::event_button3())
       buttonMask |= 1 << 2;
 
+  // The back/forward buttons are not supported by FTLK 1.3 and require
+  // a patch which adds these buttons to the FLTK API. These buttons
+  // will be part of the upcoming 1.4 API:
+  //   * https://github.com/fltk/fltk/pull/1081
+  //
+  // A backport for branch-1.3 is available here:
+  //   * https://github.com/fltk/fltk/pull/1083
+#if defined(FL_BUTTON4) && defined(FL_BUTTON5)
+    if (Fl::event_button4())
+      buttonMask |= 1 << 7;
+    if (Fl::event_button5())
+      buttonMask |= 1 << 8;
+#endif
+
     if (event == FL_MOUSEWHEEL) {
       wheelMask = 0;
       if (Fl::event_dy() < 0)
@@ -660,7 +674,7 @@ int Viewport::handle(int event)
   return Fl_Widget::handle(event);
 }
 
-void Viewport::sendPointerEvent(const rfb::Point& pos, uint8_t buttonMask)
+void Viewport::sendPointerEvent(const rfb::Point& pos, uint16_t buttonMask)
 {
   if (viewOnly)
       return;
@@ -790,7 +804,7 @@ void Viewport::flushPendingClipboard()
 }
 
 
-void Viewport::handlePointerEvent(const rfb::Point& pos, uint8_t buttonMask)
+void Viewport::handlePointerEvent(const rfb::Point& pos, uint16_t buttonMask)
 {
   filterPointerEvent(pos, buttonMask);
 }
@@ -937,6 +951,8 @@ int Viewport::handleSystemEvent(void *event, void *data)
       (msg->message == WM_RBUTTONUP) ||
       (msg->message == WM_MBUTTONDOWN) ||
       (msg->message == WM_MBUTTONUP) ||
+      (msg->message == WM_XBUTTONDOWN) ||
+      (msg->message == WM_XBUTTONUP) ||
       (msg->message == WM_MOUSEWHEEL) ||
       (msg->message == WM_MOUSEHWHEEL)) {
     // We can't get a mouse event in the middle of an AltGr sequence, so
--- a/vncviewer/Viewport.h
+++ b/vncviewer/Viewport.h
@@ -70,7 +70,7 @@ public:
   int handle(int event) override;
 
 protected:
-  void sendPointerEvent(const rfb::Point& pos, uint8_t buttonMask) override;
+  void sendPointerEvent(const rfb::Point& pos, uint16_t buttonMask) override;
 
 private:
   bool hasFocus();
@@ -81,7 +81,7 @@ private:
 
   void flushPendingClipboard();
 
-  void handlePointerEvent(const rfb::Point& pos, uint8_t buttonMask);
+  void handlePointerEvent(const rfb::Point& pos, uint16_t buttonMask);
   static void handlePointerTimeout(void *data);
 
   void resetKeyboard();
@@ -111,7 +111,7 @@ private:
   PlatformPixelBuffer* frameBuffer;
 
   rfb::Point lastPointerPos;
-  uint8_t lastButtonMask;
+  uint16_t lastButtonMask;
 
   typedef std::map<int, uint32_t> DownMap;
   DownMap downKeySym;
--- a/win/rfb_win32/SDisplay.cxx
+++ b/win/rfb_win32/SDisplay.cxx
@@ -312,7 +312,7 @@ void SDisplay::handleClipboardData(const char* data) {
 }
 
 
-void SDisplay::pointerEvent(const Point& pos, uint8_t buttonmask) {
+void SDisplay::pointerEvent(const Point& pos, uint16_t buttonmask) {
   if (pb->getRect().contains(pos)) {
     Point screenPos = pos.translate(screenRect.tl);
     // - Check that the SDesktop doesn't need restarting
--- a/common/rfb/CConnection.cxx
+++ b/common/rfb/CConnection.cxx
@@ -835,6 +835,7 @@ void CConnection::updateEncodings()
   encodings.push_back(pseudoEncodingContinuousUpdates);
   encodings.push_back(pseudoEncodingFence);
   encodings.push_back(pseudoEncodingQEMUKeyEvent);
+  encodings.push_back(pseudoEncodingExtendedMouseButtons);
 
   if (Decoder::supported(preferredEncoding)) {
     encodings.push_back(preferredEncoding);
--- a/common/rfb/CMsgHandler.cxx
+++ b/common/rfb/CMsgHandler.cxx
@@ -75,6 +75,11 @@ void CMsgHandler::endOfContinuousUpdates()
   server.supportsContinuousUpdates = true;
 }
 
+void CMsgHandler::supportsExtendedMouseButtons()
+{
+  server.supportsExtendedMouseButtons = true;
+}
+
 void CMsgHandler::supportsQEMUKeyEvent()
 {
   server.supportsQEMUKeyEvent = true;
--- a/common/rfb/CMsgHandler.h
+++ b/common/rfb/CMsgHandler.h
@@ -57,6 +57,7 @@ namespace rfb {
     virtual void fence(uint32_t flags, unsigned len, const uint8_t data[]);
     virtual void endOfContinuousUpdates();
     virtual void supportsQEMUKeyEvent();
+    virtual void supportsExtendedMouseButtons();
     virtual void serverInit(int width, int height,
                             const PixelFormat& pf,
                             const char* name) = 0;
--- a/common/rfb/CMsgReader.cxx
+++ b/common/rfb/CMsgReader.cxx
@@ -202,6 +202,10 @@ bool CMsgReader::readMsg()
       handler->supportsQEMUKeyEvent();
       ret = true;
       break;
+    case pseudoEncodingExtendedMouseButtons:
+      handler->supportsExtendedMouseButtons();
+      ret = true;
+      break;
     default:
       ret = readRect(dataRect, rectEncoding);
       break;
--- a/common/rfb/CMsgWriter.h
+++ b/common/rfb/CMsgWriter.h
@@ -54,7 +54,7 @@ namespace rfb {
     void writeFence(uint32_t flags, unsigned len, const uint8_t data[]);
 
     void writeKeyEvent(uint32_t keysym, uint32_t keycode, bool down);
-    void writePointerEvent(const Point& pos, uint8_t buttonMask);
+    void writePointerEvent(const Point& pos, uint16_t buttonMask);
 
     void writeClientCutText(const char* str);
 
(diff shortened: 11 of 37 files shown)


Neustradamus:
To follow this new PR :)

CendioHalim:
I've now updated `Viewport.cxx` so that all* mouse handling is done using native platform code: https://github.com/TigerVNC/tigervnc/pull/1826/commits/30e4e9efc0685e4f415e276cbca1c99679968d09

(*) There is still a known bug where `FL_ENTER` and `FL_LEAVE` in `Viewport::handle()` are still reached. I think that these should be removed completely, as they are pointer events that should be handled entirely in `handleSystemPointerEvent()`.

CendioHalim:
Seems like the close button is not respected on macOS (minimize/maximize works), will have to do some digging to find out why that is.

CendioHalim:
I've fixed the issued and updated my branch. Most of the work was done in https://github.com/TigerVNC/tigervnc/pull/1826/commits/cfe40cc637f21d9d52146af50a37f1830eb4b185, which has a very hacky solution to fix resizing on macOS.

CendioHalim:
There appears to still be some issues:

When I start vncviewer on Linux, the cursor is invisible until I enter and leave the window once.

On Windows, the window buttons (minimize, maximize, close) don't work (we should return 0 instead of consuming the events). Also, the viewer doesn't detect WM_MOUSELEAVE / mouse-enter unless I lose focus and get window focus again.

On macOS, the resizing is still a bit buggy and can leave the window in a broken state where the viewer can't be resized / loses input.

CendioHalim:
There is also a bug on all platforms where pressing the forward button while moving the mouse will trigger a push/release/push instantly. Haven't looked into why this happens, but I've seen it on all platforms with three different mice.

CendioHalim:
> On Windows, the window buttons (minimize, maximize, close) don't work (we should return 0 instead of consuming the events). Also, the viewer doesn't detect WM_MOUSELEAVE / mouse-enter unless I lose focus and get window focus again.
Fixed in https://github.com/TigerVNC/tigervnc/pull/1826/commits/2f8fe06f80a00754fb1aceff6ccdeed620427e4c and https://github.com/TigerVNC/tigervnc/pull/1826/commits/f82a0741a186f4e79b00ad050083c43cc90e4eb0.

CendioHalim:
I've opened a PR to FLTK to fix forward/back mouse button support for macOS and Windows:
* https://github.com/fltk/fltk/pull/1068

With this fix in FLTK, the TigerVNC code can be immensely simplified. I've updated my branch to reflect the changes done in the PR.

vncviewer/Viewport.cxx
```diff
@@ -598,32 +600,52 @@ int Viewport::handle(int event)
...
     if (Fl::event_button2())
-      buttonMask |= 2;
+        mouseButtonMask |= 1 << 1;
     if (Fl::event_button3())
-      buttonMask |= 4;
+        mouseButtonMask |= 1 << 2;
+
+    switch (event) {
+    case FL_PUSH:
+    case FL_DRAG:
```
CendioOssman, line 616: Why is `FL_DRAG` needed? Won't it always be preceded by a `FL_PUSH`?

vncviewer/Viewport.h
```diff
@@ -35,6 +35,10 @@ class CConn;
 class PlatformPixelBuffer;
 class Surface;
 
+// ExtendedMouseButtons
+const int MOUSE_FORWARD = 8;
+const int MOUSE_BACK = 9;
```
CendioOssman, line 40: Aren't these part of the new FLTK API? Seems strange to have them here in TigerVNC.

vncviewer/Viewport.cxx
```diff
@@ -589,6 +590,7 @@ int Viewport::handle(int event)
 
   case FL_LEAVE:
     window()->cursor(FL_CURSOR_DEFAULT);
+    mouseButtonMask = 0;
```
CendioOssman, line 593: Have you checked for cases where the system might steal things for us? E.g. Alt+Tab? Or screenshots?

Do we get a `FL_LEAVE` in those cases? Or do we need to check for something else?

CendioOssman (changes requested):
Looks really nice! Great work!

There are some minor tweaks, but we are pretty much done.

Please also organise the commit history to something more final.

vncviewer/Viewport.cxx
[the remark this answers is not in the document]
CendioHalim: Correct, I can keep a button pressed and move outside the viewport and get an `FL_LEAVE` when the button is released.

vncviewer/Viewport.h
CendioHalim: Good point, I've removed these now.

common/rfb/SMsgReader.cxx
```diff
@@ -272,11 +272,32 @@ bool SMsgReader::readKeyEvent()
...
+
+    if (!is->hasData(1))
+      return false;
+
+    highBits = is->readU8();
+    lowBits = mask & 0x7f; // Clear marker bit
+    mask = (highBits << 7) | lowBits;
+  } else {
+    // Ignore marker bit
+    mask &= ~(0x7f);
```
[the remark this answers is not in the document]
CendioHalim: That is true. We should still read the 8th bit, as we've done previously. This will make our server still support the back button, even without the pseudo-encoding enabled with legacy clients. This is probably how RealVNC works currently.

vncviewer/Viewport.cxx
```diff
@@ -598,32 +600,52 @@ int Viewport::handle(int event)
   case FL_DRAG:
   case FL_MOVE:
   case FL_MOUSEWHEEL:
-    buttonMask = 0;
+
+    // Reset state for left, right, middle
+    mouseButtonMask &= ~(0x7);
```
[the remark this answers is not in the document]
CendioHalim: Unfortunately, FLTK is quite buggy and I couldn't reliably keep the button states in sync:
  * https://github.com/fltk/fltk/issues/1076
  * https://github.com/fltk/fltk/issues/1077
  
So we either have to fix these issues upstream, or work around this issue by looking at the button states from LMB/MMB/RMB for each event (current implementation).

The bug that is in FLTK means that we can get in a scenario where the state for the forward/back buttons get out of sync in some edge cases. I could reproduce this by holding the back button, then tabbing to xterm and releasing the back button. Hovering back to the viewport, the `mouseButtonMask` had the back button bit set. I had to click the back button once to reset it.

unix/xserver/hw/vnc/vncInput.c
```diff
@@ -206,6 +206,8 @@ static int vncPointerProc(DeviceIntPtr pDevice, int onoff)
 		btn_labels[4] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_WHEEL_DOWN);
 		btn_labels[5] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_LEFT);
 		btn_labels[6] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_RIGHT);
+		btn_labels[7] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_SIDE);
+		btn_labels[8] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_EXTRA);
```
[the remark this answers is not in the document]
CendioHalim: I've added a comment now. It could probably be a bit more detailed/specific, but I don't really feel like digging through that rabbit hole right now.

vncviewer/Viewport.cxx
CendioHalim: https://github.com/TigerVNC/tigervnc/pull/1826#discussion_r1773230996

CendioHalim:
I've addressed most of the comments and rebased/cleaned up my commits.

vncviewer/Viewport.cxx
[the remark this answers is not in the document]
Albrecht-S: @CendioHalim wrote:
> Unfortunately, FLTK is quite buggy and I couldn't reliably keep the button states in sync ...

I disagree with **this wording** (_"quite buggy"_).

I commented on both above mentioned FLTK issues. Fact is that FLTK **does not support** "mouse buttons 8 and 9" (yet). These buttons are AFAICT not at all documented in the Xorg docs and in any X11 headers! I'm also not aware of any other docs (conventions) that define the meaning and internals of these extended mouse buttons.

Dragging the mouse while holding an unsupported mouse button **and** using `Alt+Tab` - which is usually intercepted/handled by the OS (WM) - can't be considered a bug as well. As you, @CendioHalim, wrote elsewhere, X11 does not keep the button states itself, and FLTK relies on X11 sending the current button states.

Of course we're interested in extending FLTK to support these additional mouse buttons, but calling the current state "quite buggy" is not appropriate.

unix/xserver/hw/vnc/vncInput.c
```diff
@@ -207,6 +207,24 @@ static int vncPointerProc(DeviceIntPtr pDevice, int onoff)
...
+     *
+     * There are also two similar labels:
+     *		BTN_LABEL_PROP_BTN_BACK and BTN_LABEL_PROP_BTN_FORWARD,
+     * which are also used by *some* X applications to indicate
+     * back/forward navigation. The SIDE/EXTRA labels seem to be more
+     * widely spread by X applications, and some applications even
+     * detect both labels.
+     *
+     * Most modern mice seem to label the side buttons
+     * (back / forward) with SIDE/EXTRA.
```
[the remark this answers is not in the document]
CendioOssman: Not sure if it needs to be added to the comment, but it's probably also good to keep in mind that many (most?) applications do not look at these labels, but rather just the button index. E.g. FLTK?

unix/xserver/hw/vnc/vncInput.c
```diff
@@ -207,6 +207,24 @@ static int vncPointerProc(DeviceIntPtr pDevice, int onoff)
 		btn_labels[5] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_LEFT);
 		btn_labels[6] = XIGetKnownProperty(BTN_LABEL_PROP_BTN_HWHEEL_RIGHT);
 
+    /*
```
CendioOssman, line 210: Indentation is off here.

unix/xserver/hw/vnc/vncInput.c
```diff
@@ -50,7 +50,7 @@ extern const unsigned int code_map_qnum_to_xorgevdev_len;
 extern const unsigned short code_map_qnum_to_xorgkbd[];
 extern const unsigned int code_map_qnum_to_xorgkbd_len;
 
-#define BUTTONS 7
+#define BUTTONS 9
```
CendioOssman, line 53: Does this implicitly also make unknown bits ignored? Or are they just unlabeled and perhaps wrong?

common/rfb/CMsgWriter.cxx
```diff
@@ -176,15 +176,37 @@ void CMsgWriter::writeKeyEvent(uint32_t keysym, uint32_t keycode, bool down)
...
   Point p(pos);
+  bool extendedMouseButtons;
+
   if (p.x < 0) p.x = 0;
   if (p.y < 0) p.y = 0;
   if (p.x >= server->width()) p.x = server->width() - 1;
   if (p.y >= server->height()) p.y = server->height() - 1;
 
+  // Only send extended pointerEvent message when needed
+  extendedMouseButtons = buttonMask & 0x180;
```
CendioOssman, line 187: Future-proof and use `0x7f80`?

CendioOssman, line 187: Worth adding a `assert(!(buttonMask & 0x8000))`?

common/rfb/CMsgWriter.cxx
```diff
@@ -176,15 +176,37 @@ void CMsgWriter::writeKeyEvent(uint32_t keysym, uint32_t keycode, bool down)
...
+    lowerBits |= 0x80; // Set marker bit to 1
+
+    higherBits &= 0x03; // Clear reserved bits
+
+    os->writeU8(lowerBits);
+    os->writeU16(p.x);
+    os->writeU16(p.y);
+    os->writeU8(higherBits);
+  } else {
+    buttonMask &= 0x7f; // Set marker bit to 0
```
CendioOssman, line 205: We probably need a comment why this is needed, or someone might easily remove it.

common/rfb/CMsgWriter.cxx
```diff
@@ -176,15 +176,37 @@ void CMsgWriter::writeKeyEvent(uint32_t keysym, uint32_t keycode, bool down)
...
-  os->writeU16(p.y);
+  if (server->supportsExtendedMouseButtons && extendedMouseButtons) {
+    int higherBits;
+    int lowerBits;
+
+    higherBits = (buttonMask  >> 7) & 0xff;
+    lowerBits = buttonMask & 0x7f;
+    lowerBits |= 0x80; // Set marker bit to 1
+
+    higherBits &= 0x03; // Clear reserved bits
```
CendioOssman, line 198: I think asserts are better at this layer.

But mapping is handled by the caller, so should we have this here at all?

vncviewer/Viewport.cxx
```diff
@@ -115,9 +115,10 @@ enum { ID_DISCONNECT, ID_FULLSCREEN, ID_MINIMIZE, ID_RESIZE,
 static const WORD SCAN_FAKE = 0xaa;
 #endif
 
+
 Viewport::Viewport(int w, int h, const rfb::PixelFormat& /*serverPF*/, CConn* cc_)
   : Fl_Widget(0, 0, w, h), cc(cc_), frameBuffer(nullptr),
-    lastPointerPos(0, 0), lastButtonMask(0),
+    lastPointerPos(0, 0), mouseButtonMask(0), lastButtonMask(0),
```
CendioOssman, line 121: Please split the mask tracking to a separate commit. It's complex enough that it's best if we get a clear diff of what changes it entails.

vncviewer/Viewport.cxx
CendioOssman, line 593: A short comment here perhaps to explain why this is needed?

vncviewer/Viewport.cxx
```diff
@@ -598,13 +600,40 @@ int Viewport::handle(int event)
   case FL_DRAG:
   case FL_MOVE:
   case FL_MOUSEWHEEL:
-    buttonMask = 0;
+
+    // FIXME: FLTK is not consistent in how it sends FL_PUSH/FL_RELEASE
+    // events:
+    //   * https://github.com/fltk/fltk/issues/1076
+    //   * https://github.com/fltk/fltk/issues/1077.
```
CendioOssman, line 607: Avoid the dot. It can mess up the link

vncviewer/Viewport.cxx
```diff
@@ -598,13 +600,40 @@ int Viewport::handle(int event)
   case FL_DRAG:
   case FL_MOVE:
   case FL_MOUSEWHEEL:
-    buttonMask = 0;
+
+    // FIXME: FLTK is not consistent in how it sends FL_PUSH/FL_RELEASE
+    // events:
+    //   * https://github.com/fltk/fltk/issues/1076
```
CendioOssman, line 606: This only seems to be FLTK 1.4, and we use 1.3.

Do we still need this workaround for 1.3? It's not apparent why.

vncviewer/Viewport.cxx
CendioOssman, line 607: This seems to be more about a missing FL_LEAVE. So a workaround for just the first three buttons seems insufficient?

Can't we work around this more easily by looking at the FL_UNFOCUS?

vncviewer/Viewport.cxx
```diff
@@ -598,13 +600,40 @@ int Viewport::handle(int event)
...
+    //   * https://github.com/fltk/fltk/issues/1077.
+    // For LMB/MMB/RMB, we can get the button states consistently.
+    // We can't get the state for back/forward, and have to keep track
+    // of them ourselves. Unfortunately as FLTK has these
```
CendioOssman, line 610: This is an important note even without FLTK bugs. Perhaps it should be separated out to be made more clear?

vncviewer/Viewport.cxx
```diff
@@ -598,13 +600,40 @@ int Viewport::handle(int event)
...
+    // inconsistencies, we could end up in a scenario where the
+    // mouseButtonMask state is out of sync with reality
+    // (for back/forward). There isn't much we can do until the bug is
+    // resolved in FLTK.
+    mouseButtonMask &= ~(0x07);
     if (Fl::event_button1())
-      buttonMask |= 1;
+        mouseButtonMask |= 1;
```
CendioOssman: Indentation got messed up here.

vncviewer/Viewport.h
```diff
@@ -111,7 +111,8 @@ class Viewport : public Fl_Widget, public EmulateMB {
   PlatformPixelBuffer* frameBuffer;
 
   rfb::Point lastPointerPos;
-  uint8_t lastButtonMask;
+  uint16_t mouseButtonMask;
```
CendioOssman, line 114: This grouping can be a bit confusing. It has nothing to do with `lastPointerPos` or `lastButtonMask`.

vncviewer/Viewport.cxx
```diff
@@ -598,13 +600,40 @@ int Viewport::handle(int event)
   case FL_DRAG:
   case FL_MOVE:
   case FL_MOUSEWHEEL:
-    buttonMask = 0;
+
+    // FIXME: FLTK is not consistent in how it sends FL_PUSH/FL_RELEASE
+    // events:
```
CendioOssman, line 605: You could consider making the workaround separate commits, to make things more clear. Assuming it works okayish without the workarounds in place. We don't want the commit without a workaround to be completely broken.

CendioOssman (changes requested):
For the commit message; full URLs are preferred to references that only work in GitHub. Or perhaps omit that reference to the old PR and just like it here in the new PR?

Same with using GitHub usernames. If it's a commit message, then a name (and possibly email) is best. Just like for git's author/committer.

vncviewer/Viewport.cxx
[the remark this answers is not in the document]
CendioHalim: > @CendioHalim wrote:
> 
> > Unfortunately, FLTK is quite buggy and I couldn't reliably keep the button states in sync ...
> 
> I disagree with **this wording** (_"quite buggy"_).

I apologize for my wording.  

FLTK sends input events for back/forward, and I've found some cases where this happens incorrectly. Whether this is a bug in FLTK or the WM/OS, I'm not sure and here is not much I can do from my perspective.

FWIW, at one point, I added my own system handler to handle all mouse events with native code and not rely on FLTK. With that setup, I did not have the issue described in my earlier comment. So there is some kind of weird interaction going on in that particular case.

> Dragging the mouse while holding an unsupported mouse button and using Alt+Tab - which is usually intercepted/handled by the OS (WM) - can't be considered a bug as well. 

To clarify, this happens when using LMB as well. The bits I'm talking about in my comment is my own button state tracker, not anything from FLTK. Both `Fl::event_state()` and `fl_xevent->xbutton.state` are set correctly, but I still get `FL_PUSH` when no buttons are pressed.

unix/xserver/hw/vnc/vncInput.c
CendioHalim: Whoops, looks like the file uses tabs and not spaces.

unix/xserver/hw/vnc/vncInput.c
CendioHalim: We've explicitly labelled and initialized 9 buttons. We then loop through the mask, bits set in the higher range are ignored:

```c
void vncPointerButtonAction(int buttonMask)
{
	int i;
	ValuatorMask mask;

	for (i = 0; i < BUTTONS; i++) {
		if ((buttonMask ^ oldButtonMask) & (1 << i)) {
			int action = (buttonMask & (1<<i)) ?
				     ButtonPress : ButtonRelease;
			valuator_mask_set_range(&mask, 0, 0, NULL);
			QueuePointerEvents(vncPointerDev, action, i + 1,
					   POINTER_RELATIVE, &mask);
		}
	}

	oldButtonMask = buttonMask;
}
```

vncviewer/Viewport.cxx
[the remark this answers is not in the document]
Albrecht-S: @CendioHalim wrote:
> I apologize for my wording.

Thanks, appreciated.

Regarding status of back/forward mouse buttons when switching context (`Alt+Tab` or `unfocus/focus`): ISTR that FLTK sets up the current keyboard and mouse button status whenever we [lose and] get focus [back]. On X11 I'm pretty sure that we read the current status from the X11 (`FOCUS` or any other) event which would preclude getting the back/forward button status since X11 doesn't keep the status and thus doesn't send it in any event.

OTOH, if we kept the status internally and the user pushed and held one of these buttons before context switch (i.e. focus loss) we wouldn't be able to determine the status when we get the focus back because the user could have released the button after the switch. I think this is one scenario you mentioned previously here or elsewhere.

This is something that needs investigation and can maybe not be solved at all unless someone finds out how to query the current button status when we get the focus back.

PS: I received my new mouse and did a first few tests. Back/forward buttons work in firefox as expected. I'll post my results later at one of the FLTK issues or PR's.

vncviewer/Viewport.cxx
CendioHalim: Indeed, the issue still persists for the back/forward buttons. Resetting the state at `FL_UNFOCUS` would be a cleaner workaround.

CendioHalim: This is no longer an issue as FLTK is in the progress of adding support for back/forward in their API:

* https://github.com/fltk/fltk/pull/1081
* https://github.com/fltk/fltk/pull/1083

common/rfb/CMsgWriter.cxx
CendioHalim: Good point, I've changed it to an assert instead. Not sure if it fits better here or in the Viewport code (or somewhere in-between)

vncviewer/DesktopWindow.cxx
```diff
@@ -174,7 +174,7 @@ DesktopWindow::DesktopWindow(int w, int h, const char *name,
 #ifdef __APPLE__
   // On OS X we can do the maximize thing properly before the
   // window is showned. Other platforms handled further down...
-  if (maximize) {
+  if (false) {
```
CendioOssman, line 177: Oops? :)

common/rfb/CMsgWriter.cxx
```diff
@@ -173,18 +174,48 @@ void CMsgWriter::writeKeyEvent(uint32_t keysym, uint32_t keycode, bool down)
...
+    int higherBits;
+    int lowerBits;
+
+    higherBits = (buttonMask  >> 7) & 0xff;
+    assert(!(higherBits & 0xfc)); /* Bits 2-7 are reserved */
+
+    lowerBits = buttonMask & 0x7f;
+    lowerBits |= 0x80; /* Set marker bit to 1 */
+
+
```
CendioOssman, line 204: Double empty lines.

vncviewer/Viewport.cxx
```diff
@@ -115,6 +115,7 @@ enum { ID_DISCONNECT, ID_FULLSCREEN, ID_MINIMIZE, ID_RESIZE,
 static const WORD SCAN_FAKE = 0xaa;
 #endif
 
+
```
CendioOssman, line 118: Stray change.

vncviewer/Viewport.cxx
```diff
@@ -1234,20 +1250,20 @@ void Viewport::initContextMenu()
 
   fltk_menu_add(contextMenu, p_("ContextMenu|", "&Full screen"),
                 0, nullptr, (void*)ID_FULLSCREEN,
-                FL_MENU_TOGGLE | (window()->fullscreen_active()?FL_MENU_VALUE:0));
+                2 | (window()->fullscreen_active()?4:0));
```
CendioOssman, line 1253: More oops?

CendioOssman (changes requested):

vncviewer/DesktopWindow.cxx
CendioHalim: Oops indeed, hardcoded some things to build with fltk-1.4

CendioOssman:
Almost complete. x0vncserver doesn't work correctly for some reason. Only the back button works, not the forward button.

CendioHalim:
> Almost complete. x0vncserver doesn't work correctly for some reason. Only the back button works, not the forward button.

https://github.com/TigerVNC/tigervnc/pull/1826/commits/1c2f07f3a9c97178d0f84c80d60105be0384d6f9 fixes this, but this would have to be updated if we want to add support for more buttons in the future.

win/rfb_win32/SDisplay.cxx
```diff
@@ -312,7 +312,7 @@ void SDisplay::handleClipboardData(const char* data) {
 }
 
 
-void SDisplay::pointerEvent(const Point& pos, uint8_t buttonmask) {
+void SDisplay::pointerEvent(const Point& pos, uint16_t buttonmask) {
```
CendioOssman, line 315: This and `SInput` are for the Windows server, so they are misplaced in the client commit.

Perhaps move these changes, and the x0vncserver changes, to the Xvnc commit and simply have that change all servers?

CendioOssman (changes requested):
Seems to work correctly now. Good work!

But I think a tweak is needed on which change is in which commit.

CendioHalim:
> When can we expect [fltk/fltk#1083](https://github.com/fltk/fltk/pull/1083) to be merged? I don't want to risk the API changing.

The FLTK fix has already been merged into master: https://github.com/fltk/fltk/pull/1081 and should be part of the upcoming 1.4.0 API. 

Regarding the backport to 1.3 , it looks like it might be a while: https://github.com/fltk/fltk/pull/1083#discussion_r1808730440

CendioOssman:
We are targeting FLTK 1.3, not 1.4, so we need to have some confirmation that we can start relying on this API. Otherwise we might end up in a mess where there are several version with subtle differences that we somehow need to handle.

Albrecht-S:
> We are targeting FLTK 1.3, not 1.4, so we need to have some confirmation that we can start relying on this API. Otherwise we might end up in a mess where there are several version with subtle differences that we somehow need to handle.

Sorry, our release process for 1.4.0 will take a while, depending on the number of RC's we need to publish. RC2 will certainly follow because we do already have one essential fix. Each RC is scheduled one week after the previous one and the final release will typically be two weeks after the last RC, i.e. the final release will be published if we don't have any issues for two weeks.

I can't backport the changes to 1.3.10 (i.e. merge the existing PR) before 1.4.0 has been released which will not be before about 3 weeks from now (or later if more RC's are required, see above).

CendioOssman:
We don't have to wait for a 1.3.10 release. We can apply a patch to our builds until then. But we need to know that you are happy with the API, and we won't have to change things once 1.3.10 is out.

pfactum:
Is it supposed to work with `x0vncserver`?

I've got 1.15.0 now on both client and server.

On server: `x0vncserver -display=:0 -Geometry=2560x1440-1920+0 -SecurityTypes=None -UseIPv6=0 -AcceptSetDesktopSize=0`

On client: `vncviewer x.x.x.x:0 -FullScreen=on -FullscreenSystemKeys=off -MenuKey=F12 -AcceptClipboard=off -SendClipboard=off -SendPrimary=off -SetPrimary=off`

Running `xev` on server and clicking "forward"/"backward" mouse buttons yields `MotionNotify` events but not `ButtonPress`/`ButtonRelease` events.

CendioHalim:
@correia55

What VNC client and OS are you using? I'm guessing forward works as expected locally on the client you're using?

CendioHalim:
> > @correia55
> > What VNC client and OS are you using? I'm guessing forward works as expected locally on the client you're using?
> 
> The client is a Remote Desktop Connection (not sure of the version), in a Citrix workspace (22.3.2000.2105) in the browser, from a Windows Machine. Everything works well in the Windows machine.

I've got no experience with that client, but my guess is that they simply haven't implemented the extension that is needed for the forward/back buttons to work.

Are you able to use the TigerVNC client, or maybe noVNC using a chromium browser? Those should work with the buttons. The latter does require a little more setup to work:
* https://github.com/novnc/noVNC?tab=readme-ov-file#quick-start

PR #1830, opened 2024-09-05
Add developer documents

KangLin:


CendioOssman:
Thank you for your suggestion. Unfortunately, this is something that will need to be maintained as the code changes. And I'm afraid I don't see that there is enough we gain here for that extra work (and risk of confusion when someone fails to keep this documentation updated).

Issue #529, opened 2017-10-24
x0vncserver does not support clipboard


astrand:
Currently, x0vncserver has no support for sending clipboard changes to the client, or accept clipboard updates from the client.

Racansky:
Hi,
for x0vncserver I used to run the parameters `-SendCutText -AcceptCutText` - then my clipboard works in both directions

CendioOssman:
I'm afraid that can't be our `x0vncserver` in that case. This is the complete current code for clipboard handling right now:

```C++
void XDesktop::clientCutText(const char* str, int len) {
}
```

:)

TheQuantumPhysicist:
I thought I found the perfect VNC server... but now this... I seriously can't catch a break with VNC!

Mannshoch:
Is that problem still available?

CendioOssman:
I'm afraid no one has had a look at this yet.

ivorob:
Hi. Is the bounty active for that issue? I want to try resolve it to start working with project.

CendioOssman:
Yes, the bounty is still active so feel free to start hacking. :)

ivorob:
Guys, I am close to provide code to code review. But I have a several questions. Do you have some code style standard? Where can I find it. Sorry for offtopic.

CendioOssman:
Sorry, no. You'll have to look at existing files and try to follow the same style.

mathewng:
After establishing a connection to `X0tigervnc`, I launched `x11vnc -nofb -clip 4x4+0+0` to synchronise the clipboard.
This works for me.

markd89:
@mathewng can you post more details of your workaround. When I try that x11vnc listens on 5901 and I don't get any clipboard transfer in tigervncviewer. I'm thinking that maybe I need to run a second viewer to get the clipboard transfer?

@ivorob how are you doing with your code? You may be able to find something in the x11vnc source code that you can modify to add to x0vncserver. I would be willing to add to the bountysource if you could get this done soon.

mathewng:
> @mathewng can you post more details of your workaround. When I try that x11vnc listens on 5901 and I don't get any clipboard transfer in tigervncviewer. I'm thinking that maybe I need to run a second viewer to get the clipboard transfer?
> 
> @ivorob how are you doing with your code? You may be able to find something in the x11vnc source code that you can modify to add to x0vncserver. I would be willing to add to the bountysource if you could get this done soon.

Yes, you are right. 

[...]

PR #1831, opened 2024-09-05
Add clipboard support to x0vncserver

gujjwal00:
Most of the groundwork related to selections is already present in `TxWindow`. And `XDesktop` already makes use of XFixes, which can provide a notification whenever selection changes. I have added `XSelection` to handle the rest (took some notes from `vncSelection.c` in Xvnc). 

- Only `CLIPBOARD` selection is handled for now (because I don't personally use `PRIMARY`). Is this sufficient?
- When notified about selection change, current implementation will immediately ask for clip-data from selection owner, and will then notify VNC clients. It simplifies things a bit and in most configurations net outcome is the same , but  `vncSelection.c` seems to only request data from selection owner when explicitly asked by clients in `handleClipboardRequest()`.

I didn't want to make it too complicated before getting some feedback.

Issue: #529

### DIFF
--- a/unix/x0vncserver/XDesktop.cxx
+++ b/unix/x0vncserver/XDesktop.cxx
@@ -45,6 +45,7 @@
 #endif
 #ifdef HAVE_XFIXES
 #include <X11/extensions/Xfixes.h>
+#include <X11/Xatom.h>
 #endif
 #ifdef HAVE_XRANDR
 #include <X11/extensions/Xrandr.h>
@@ -83,7 +84,7 @@ static const char * ledNames[XDESKTOP_N_LEDS] = {
 
 XDesktop::XDesktop(Display* dpy_, Geometry *geometry_)
   : dpy(dpy_), geometry(geometry_), pb(nullptr), server(nullptr),
-    queryConnectDialog(nullptr), queryConnectSock(nullptr),
+    queryConnectDialog(nullptr), queryConnectSock(nullptr), selection(dpy_, this),
     oldButtonMask(0), haveXtest(false), haveDamage(false),
     maxButtons(0), running(false), ledMasks(), ledState(0),
     codeMap(nullptr), codeMapLen(0)
@@ -181,10 +182,15 @@ XDesktop::XDesktop(Display* dpy_, Geometry *geometry_)
   if (XFixesQueryExtension(dpy, &xfixesEventBase, &xfixesErrorBase)) {
     XFixesSelectCursorInput(dpy, DefaultRootWindow(dpy),
                             XFixesDisplayCursorNotifyMask);
+
+    XFixesSelectSelectionInput(dpy, DefaultRootWindow(dpy), XA_PRIMARY,
+                               XFixesSetSelectionOwnerNotifyMask);
+    XFixesSelectSelectionInput(dpy, DefaultRootWindow(dpy), xaCLIPBOARD,
+                               XFixesSetSelectionOwnerNotifyMask);
   } else {
 #endif
     vlog.info("XFIXES extension not present");
-    vlog.info("Will not be able to display cursors");
+    vlog.info("Will not be able to display cursors or monitor clipboard");
 #ifdef HAVE_XFIXES
   }
 #endif
@@ -891,6 +897,20 @@ bool XDesktop::handleGlobalEvent(XEvent* ev) {
       return false;
 
     return setCursor();
+  }
+  else if (ev->type == xfixesEventBase + XFixesSelectionNotify) {
+    XFixesSelectionNotifyEvent* sev = (XFixesSelectionNotifyEvent*)ev;
+
+    if (!running)
+      return true;
+
+    if (sev->subtype != XFixesSetSelectionOwnerNotify)
+      return false;
+
+    selection.handleSelectionOwnerChange(sev->owner, sev->selection,
+                                         sev->timestamp);
+
+    return true;
 #endif
 #ifdef HAVE_XRANDR
   } else if (ev->type == Expose) {
@@ -1038,3 +1058,28 @@ bool XDesktop::setCursor()
   return true;
 }
 #endif
+
+// X selection availability changed, let VNC clients know
+void XDesktop::handleXSelectionAnnounce(bool available) {
+  server->announceClipboard(available);
+}
+
+// A VNC client wants data, send request to selection owner
+void XDesktop::handleClipboardRequest() { 
+  selection.requestSelectionData(); 
+}
+
+// Data is available, send it to clients
+void XDesktop::handleXSelectionData(const char* data) {
+  server->sendClipboardData(data);
+}
+
+// When a client says it has clipboard data, request it 
+void XDesktop::handleClipboardAnnounce(bool available) {
+   if(available) server->requestClipboard();
+}
+
+// Client has sent the data
+void XDesktop::handleClipboardData(const char* data) {
+  if (data) selection.handleClientClipboardData(data);
+}
--- a/unix/x0vncserver/XSelection.cxx
+++ b/unix/x0vncserver/XSelection.cxx
@@ -1,0 +1,195 @@
+/* Copyright (C) 2024 Gaurav Ujjwal.  All Rights Reserved.
+ *
+ * This is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this software; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ * USA.
+ */
+
+#include <X11/Xatom.h>
+#include <rfb/Configuration.h>
+#include <rfb/LogWriter.h>
+#include <rfb/util.h>
+#include <x0vncserver/XSelection.h>
+
+rfb::BoolParameter setPrimary("SetPrimary",
+                              "Set the PRIMARY as well as the CLIPBOARD selection",
+                              true);
+rfb::BoolParameter sendPrimary("SendPrimary",
+                               "Send the PRIMARY as well as the CLIPBOARD selection",
+                               true);
+
+static rfb::LogWriter vlog("XSelection");
+
+XSelection::XSelection(Display* dpy_, XSelectionHandler* handler_)
+    : TXWindow(dpy_, 1, 1, nullptr), handler(handler_), announcedSelection(None)
+{
+  probeProperty = XInternAtom(dpy, "TigerVNC_ProbeProperty", False);
+  transferProperty = XInternAtom(dpy, "TigerVNC_TransferProperty", False);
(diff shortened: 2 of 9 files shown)


unix/x0vncserver/XDesktop.cxx
```diff
@@ -181,10 +181,13 @@ XDesktop::XDesktop(Display* dpy_, Geometry *geometry_)
   if (XFixesQueryExtension(dpy, &xfixesEventBase, &xfixesErrorBase)) {
     XFixesSelectCursorInput(dpy, DefaultRootWindow(dpy),
                             XFixesDisplayCursorNotifyMask);
+
+    XFixesSelectSelectionInput(dpy, DefaultRootWindow(dpy), xaCLIPBOARD,
+                               XFixesSetSelectionOwnerNotifyMask);
```
CendioOssman, line 186: How difficult would it be to also support the selection clipboard?

unix/x0vncserver/XDesktop.cxx
```diff
@@ -1038,3 +1055,23 @@ bool XDesktop::setCursor()
...
+}
+
+// A client wants clipboard data
+void XDesktop::handleClipboardRequest() {
+  server->sendClipboardData(selection.getClipboardData().c_str());
+}
+
+// When a client says it has clipboard data, request it 
+void XDesktop::handleClipboardAnnounce(bool available) {
+  if (available) server->requestClipboard();
```
CendioOssman, line 1071: Could we wait until another application requests it? It would avoid transferring things needlessly.

unix/x0vncserver/XSelection.cxx
```diff
@@ -0,0 +1,144 @@
...
+Time XSelection::getXServerTime()
+{
+  XEvent ev;
+  uint8_t data = 0;
+
+  // Trigger a PropertyNotify event to extract server time
+  XChangeProperty(dpy, win(), xaTimestampProperty, XA_STRING, 8,
+                  PropModeReplace, &data, sizeof(data));
+  XIfEvent(dpy, &ev, &PropertyEventMatcher, (XPointer)&xaTimestampProperty);
+  return ev.xproperty.time;
```
CendioOssman, line 51: Why not just `CurrentTime`?

CendioOssman (commented):
This is awesome! Nice work!

unix/x0vncserver/XDesktop.cxx
gujjwal00: Sorry for not being clear, but in PR summary I was talking about how current code issues `XConvertSelection()` commands immediately on a clipboard change on server side, without first calling `announceClipboard(true)` and waiting for `handleClipboardRequest()` call from VNC client. But I already have a solution to do it the 2nd way.

As for the `server->requestClipboard();` here, my main motivation was to start with a simple working solution. Retrieving the clipboard  from VNC client _after_ another app requests it means that we need to queue that request (and potentially others) and fulfill it later when data is available. As you said, it would would avoid avoid needless transfers, and I am hoping to implement it. 

Although, requesting it immediately means that when user selects 'Paste', the data is available immediately without potential lag. It also avoids the problem mentioned above about owning multiple selections, as the data will always be cached.

unix/x0vncserver/XSelection.cxx
gujjwal00: I was following the advice from https://x.org/releases/X11R7.7/doc/xorg-docs/icccm/icccm.html#Peer_to_Peer_Communication_by_Means_of_Selections : 

`Clients attempting to acquire a selection must set the time value of the SetSelectionOwner request to the timestamp of the event triggering the acquisition attempt, not to CurrentTime. A zero-length append to a property is a way to obtain a timestamp for this purpose; the timestamp is in the corresponding PropertyNotify event. `

I did use `CurrentTime` initially, and it worked without an issue. But since this time is also used [here](https://github.com/TigerVNC/tigervnc/blob/2fe9dca45f7d0a519fef5053ba61e2db7c1ff2b1/unix/tx/TXWindow.cxx#L472-L475) , I changed it to current implementation just to be safe. 
I can switch it back if you want.

francoisp:
👍 👍

unix/x0vncserver/XDesktop.cxx
gujjwal00: I tried to implement this lazy loading, but it really complicates things. In addition to the issues mentioned above:
- Eventually transferred data can be larger then MaxCutText, so selection ownership needs to be released in that case
- Can't rely on `TXWindow` to generate `SelectionNotify` events, because `selectionRequest()` will only add the request to queue.

So, I plan to keep this simple for now. Because in many cases, data will be immediately transferred anyway (needing to cache data if one selection ownership is lost, clients doing unsolicited transfer, clients not supporting extended clipboard).

CendioOssman: Okay. We can have that as a possible future improvement.

unix/x0vncserver/XSelection.cxx
CendioOssman: Ah, we need it for `TIMESTAMP`. That complicates things.

*Normally*, you take the event time that caused us to grab the selection. E.g. when the user pressed "Ctrl+C". Unfortunately, we are not normal. We get our events from RFB, not from X11, so we don't have a readily available time. "Right now" is as good as it gets, and that is precisely what "CurrentTime" means.

But that only solves the call for `XSetSelectionOwner()`. So I think your suggestion is the best approach.

CendioOssman (approved):

unix/x0vncserver/x0vncserver.cxx
```diff
@@ -281,11 +281,6 @@ int main(int argc, char** argv)
 
   Configuration::enableServerParams();
 
-  // FIXME: We don't support clipboard yet
-  Configuration::removeParam("AcceptCutText");
-  Configuration::removeParam("SendCutText");
-  Configuration::removeParam("MaxCutText");
```
CendioOssman, line 287: Could you update the man page as well now that these are available?

gujjwal00:
Updates:
- Added support for `PRIMARY` selection
- Data from selection owners is retrieved lazily (when client asks via `XDesktop::handleClipboardRequest()`) 
- Improved error handling

unix/x0vncserver/x0vncserver.cxx
gujjwal00: Done.

CendioOssman (approved):
Fantastic! Thank for your hard work!

unix/x0vncserver/XDesktop.cxx
```diff
@@ -214,6 +214,15 @@ XDesktop::XDesktop(Display* dpy_, Geometry *geometry_)
...
 #endif
 
+  // Need to set an error handler for XLib because most errors are reported
+  // asynchronously and default handler terminates the process on any error.
+  XSetErrorHandler([](Display* d, XErrorEvent* ev) -> int {
+    char buff[1024];
+    XGetErrorText(d, ev->error_code, buff, sizeof(buff));
+    vlog.error("XLib error: %s", buff);
+    return 0;
+  });
```
CendioOssman, line 224: This is an unrelated change IMO, and a controversial one. All the existing code is written with the default error handler in mind.

What issue was this meant to resolve?

CendioOssman (changes requested):

unix/x0vncserver/XDesktop.cxx
gujjwal00: It was meant to avoid server crash due to possible XLib errors. During testing, I occasionally closed the app which had sent selection requests to x0vncserver, while server was paused in debugger. This would lead to `BadWindow` error once server was resumed and it would terminate. This is very unlikely to happen in real life, but I added the handler just in case.

I didn't realize this would be controversial, sorry about that. I will remove that commit.

CendioOssman: No worries. It's not expected that everyone keeps track of all of X11's evil quirks. :)

CendioOssman (approved):

bitcoinmeetups:
Ok, I want to test it. How do I install this version where clipboard works for x0vncserver on Debian 12?

cjnaz:
I see this feature as merged and this issue closed.  The latest release is 1.14.1, a bug fix release, and does not have this enhancement (probably since its not a bug).

My RHEL9 system is running 1.14.1, built Nov 8, 2024.  

When might we expect a release with this feature included?
Is there a pointer to how to build the package locally?

Thanks much,
cjn

CendioOssman:
> When might we expect a release with this feature included?

When the maintainers have sufficient free time. :)

But you're in luck. I'm trying to get a new release packaged up right now.

> Is there a pointer to how to build the package locally?

That would be the `BUILDING.txt` document. But since you are on RHEL9 I would suggest testing the nightly builds we have on our home page. There are RPMs for RHEL9 available there.

cjnaz:
I installed 
- tigervnc-server-1.14.80-1.el9.x86_64.rpm
- tigervnc-server-minimal-1.14.80-1.el9.x86_64.rpm

and ran `x0vncserver -PasswordFile=/home/cjn/.config/tigervnc/passwd -AlwaysShared=1 -Display=:1 -rfbport=5901`

The startup messages...
```
Tue Jan 14 15:19:42 2025
 Geometry:    Desktop geometry is set to 3840x2160+0+0
 XDesktop:    Using evdev codemap
 XDesktop:    
 XDesktop:    XTest extension present - version 2.2
 Main:        Listening for VNC connections on all interface(s), port 5900
```
Notice `port 5900`.  My desktop is on DISPLAY = :1, and I ssh (putty) in with tunnel localhost:5901 > remote 5901.  Changing the tunnel to remote 5900, Windows Real VNC Viewer thinks about it for a couple seconds, then "Sorry, loopback connections are not enabled".  
>> Shouldn't the Main:  line port reflect the command line -rfbport number?

I then uninstalled V14 and tried to install V15:
- tigervnc-server-1.15.80-1.el8.x86_64.rpm
- tigervnc-server-minimal-1.15.80-1.el8.x86_64.rpm

with these install errors:
```
$ sudo dnf install tigervnc-server-minimal-1.15.80-1.el8.x86_64.rpm tigervnc-server-1.15.80-1.el8.x86_64.rpm
Updating Subscription Management repositories.
Last metadata expiration check: 3:06:27 ago on Tue 14 Jan 2025 12:35:45 PM MST.
Error: 
 Problem 1: conflicting requests
  - nothing provides libhogweed.so.4()(64bit) needed by tigervnc-server-minimal-1.15.80-1.el8.x86_64 from @commandline
  - nothing provides libhogweed.so.4(HOGWEED_4)(64bit) needed by tigervnc-server-minimal-1.15.80-1.el8.x86_64 from @commandline
  - nothing provides libnettle.so.6()(64bit) needed by tigervnc-server-minimal-1.15.80-1.el8.x86_64 from @commandline
  - nothing provides libnettle.so.6(NETTLE_6)(64bit) needed by tigervnc-server-minimal-1.15.80-1.el8.x86_64 from @commandline
 Problem 2: conflicting requests
  - nothing provides libhogweed.so.4()(64bit) needed by tigervnc-server-1.15.80-1.el8.x86_64 from @commandline
  - nothing provides libhogweed.so.4(HOGWEED_4)(64bit) needed by tigervnc-server-1.15.80-1.el8.x86_64 from @commandline
  - nothing provides libnettle.so.6()(64bit) needed by tigervnc-server-1.15.80-1.el8.x86_64 from @commandline
  - nothing provides libnettle.so.6(NETTLE_6)(64bit) needed by tigervnc-server-1.15.80-1.el8.x86_64 from @commandline
```

Due to (my ignorant) install problems, no testing results to report

CendioOssman:
This does not seem to have anything to do with this pull request. Please use the discussion group to get assistance.

cjnaz:
I posted [here](https://groups.google.com/g/tigervnc-devel/c/aVknbKhF9EY).  Correct place?

PR #1833, opened 2024-09-10
Split SystemException to handle Windows

CendioOssman:
Windows has (at least) two error namespaces, both errno and GetLastResult(). These overlap, so it is important we keep track of which one we are dealing with.
    
To make things extra problematic, the BSD socket API normally uses errno, but on Windows it has been mapped in to the GetLastResult() namespace.
    
Try to keep better control of this by using separate classes for the namespaces.

### DIFF
--- a/common/network/Socket.cxx
+++ b/common/network/Socket.cxx
@@ -53,7 +53,7 @@ void network::initSockets() {
   WSADATA initResult;
   
   if (WSAStartup(requiredVersion, &initResult) != 0)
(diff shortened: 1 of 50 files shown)


samhed (approved):
Looks good! I think this makes things clearer and easier to understand. Nice to hear that it fixes a few incorrect errors as well.

One small thing; Looks like a typo in the commit message "Mocket SocketException to rdr", I guess it should be "Move ...".

Issue #787, opened 2019-01-14
X509CA option on server side


mestag-a:
Is there a technical reason why the X509CA configuration option is available on the client side, but not the server side ?
I wanted to use this so that a VNC server would only accept client connections from users with a certificate signed from the configured CA.
It is possible to configure OpenVPN and SSH to trust a given CA, so I was wondering why this "trusting feature" was implemented the other way around in TigerVNC.

CendioOssman:
X509 is currently only used to authenticate the server, not the client. So basically how most web servers work.

It should theoretically be possible to use client certificates as well, but this is not implemented at the moment. It could also require a protocol extension to work really well.

pprindeville:
I have crypto experience.  I could work on this...

CendioOssman:
Go right ahead. We have the development mailing list if you want to discuss technical details.

opentissandy:
[client_cert.patch.txt](https://github.com/user-attachments/files/17268987/client_cert.patch.txt)

I have made one patch to tag 1.14.0 that can use with qemu.
Please merge to tag 1.14.0. Thanks.

pprindeville:
> [client_cert.patch.txt](https://github.com/user-attachments/files/17268987/client_cert.patch.txt)
> 
> I have made one patch to tag 1.14.0 that can use with qemu. Please merge to tag 1.14.0. Thanks.

Can you create a PR so it can be reviewed?

pprindeville:
> X509 is currently only used to authenticate the server, not the client. So basically how most web servers work.

Not exactly true.  Apache has [`SSLVerifyClient require`](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslverifyclient) for exactly this purpose.  And [`SSLVerifyDepth n`](https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslverifydepth) controls how far up the certificate has to "ladder up" to a trusted root CA.

> It should theoretically be possible to use client certificates as well, but this is not implemented at the moment. It could also require a protocol extension to work really well.

KangLin:
This is a nice feature. but because gnutls does not support cross-platform (e.g., msvc compilation is not supported), it is recommended to use openSSL for this feature

opentissandy:
> > [client_cert.patch.txt](https://github.com/user-attachments/files/17268987/client_cert.patch.txt)
[...]

PR #1842, opened 2024-10-07
X509 client authentication #787

opentissandy:
Add feature: X509 client authentication.

### DIFF
--- a/common/rfb/CSecurityTLS.cxx
+++ b/common/rfb/CSecurityTLS.cxx
@@ -56,11 +56,11 @@ StringParameter CSecurityTLS::X509CA("X509CA", "X509 CA certificate",
 StringParameter CSecurityTLS::X509CRL("X509CRL", "X509 CRL file",
                                      configdirfn("x509_crl.pem"),
                                      ConfViewer);
-StringParameter CSecurityTLS::X509CERT("X509CERT", "X509 client certificate",
-                                     configdirfn("x509_cert.pem"),
+StringParameter CSecurityTLS::X509Cert("X509Cert", "X509 client certificate",
+                                     configdirfn("x509_client_cert.pem"),
                                      ConfViewer);
-StringParameter CSecurityTLS::X509KEY("X509KEY", "X509 client private key",
-                                     configdirfn("x509_key.pem"),
+StringParameter CSecurityTLS::X509Key("X509Key", "X509 client private key",
+                                     configdirfn("x509_client_key.pem"),
                                      ConfViewer);
 
 static LogWriter vlog("TLS");
@@ -287,7 +287,7 @@ void CSecurityTLS::setParam()
     if (gnutls_certificate_set_x509_crl_file(cert_cred, X509CRL, GNUTLS_X509_FMT_PEM) < 0)
       vlog.error("Could not load user specified certificate revocation list");
 
-    if (gnutls_certificate_set_x509_key_file (cert_cred, X509CERT, X509KEY, GNUTLS_X509_FMT_PEM) < 0)
+    if (gnutls_certificate_set_x509_key_file (cert_cred, X509Cert, X509Key, GNUTLS_X509_FMT_PEM) < 0)
       vlog.error("Could not load user specified client certificate");
 
     ret = gnutls_credentials_set(session, GNUTLS_CRD_CERTIFICATE, cert_cred);
--- a/common/rfb/CSecurityTLS.cxx
+++ b/common/rfb/CSecurityTLS.cxx
@@ -56,6 +56,12 @@ StringParameter CSecurityTLS::X509CA("X509CA", "X509 CA certificate",
 StringParameter CSecurityTLS::X509CRL("X509CRL", "X509 CRL file",
                                      configdirfn("x509_crl.pem"),
                                      ConfViewer);
+StringParameter CSecurityTLS::X509CERT("X509CERT", "X509 client certificate",
+                                     configdirfn("x509_cert.pem"),
+                                     ConfViewer);
+StringParameter CSecurityTLS::X509KEY("X509KEY", "X509 client private key",
+                                     configdirfn("x509_key.pem"),
+                                     ConfViewer);
 
 static LogWriter vlog("TLS");
 
@@ -280,6 +286,9 @@ void CSecurityTLS::setParam()
 
     if (gnutls_certificate_set_x509_crl_file(cert_cred, X509CRL, GNUTLS_X509_FMT_PEM) < 0)
       vlog.error("Could not load user specified certificate revocation list");
+
+    if (gnutls_certificate_set_x509_key_file (cert_cred, X509CERT, X509KEY, GNUTLS_X509_FMT_PEM) < 0)
+      vlog.error("Could not load user specified client certificate");
 
     ret = gnutls_credentials_set(session, GNUTLS_CRD_CERTIFICATE, cert_cred);
     if (ret != GNUTLS_E_SUCCESS)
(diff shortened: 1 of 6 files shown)


common/rfb/CSecurityTLS.cxx
```diff
@@ -56,6 +56,12 @@ StringParameter CSecurityTLS::X509CA("X509CA", "X509 CA certificate",
 StringParameter CSecurityTLS::X509CRL("X509CRL", "X509 CRL file",
                                      configdirfn("x509_crl.pem"),
                                      ConfViewer);
+StringParameter CSecurityTLS::X509CERT("X509CERT", "X509 client certificate",
```
CendioOssman, line 59: The other two are all upper case because they are abbreviations. So please use `X509Cert` here.

common/rfb/CSecurityTLS.cxx
```diff
@@ -56,6 +56,12 @@ StringParameter CSecurityTLS::X509CA("X509CA", "X509 CA certificate",
...
+                                     configdirfn("x509_cert.pem"),
```
CendioOssman, line 60: The config directory is also shared with the server, so I'm thinking we might need a more unique default name to avoid confusion. Perhaps `x509_client_cert.pem`?

vncviewer/vncviewer.man
```diff
@@ -165,6 +165,20 @@ Path to certificate revocation list to use in conjunction with
 \fI~/.config/tigervnc/x509_crl.pem\fP.
 .
 .TP
+.B \-X509CERT \fIpath\fP
+Path to client certificate to use when authenticating client using any
+of the X509 security schemes (X509None, X509Vnc, etc.). Must be in PEM
```
CendioOssman, line 170: I don't think just adding it to the existing security types will be the best user experience. The client has no way of knowing if the server will ask for a client certificate with this approach. That means the client cannot use a different security type, in case it doesn't have a certificate.

The easiest approach is probably to add a new `X509Cert` type. It would fit in the theme of specifying how the user authenticates separately.

@berrange, you've done a lot of the VeNCrypt stuff. Any insight?

common/rfb/CSecurityTLS.cxx
```diff
@@ -281,6 +287,9 @@ void CSecurityTLS::setParam()
     if (gnutls_certificate_set_x509_crl_file(cert_cred, X509CRL, GNUTLS_X509_FMT_PEM) < 0)
       vlog.error("Could not load user specified certificate revocation list");
 
+    if (gnutls_certificate_set_x509_key_file (cert_cred, X509CERT, X509KEY, GNUTLS_X509_FMT_PEM) < 0)
+      vlog.error("Could not load user specified client certificate");
```
CendioOssman, line 291: I wonder if we need to have a bit more explicit error handling. What is the user experience if there is some issue with the certificate or key file?

vncviewer/OptionsDialog.cxx
```diff
@@ -728,6 +732,20 @@ void OptionsDialog::createSecurityPage(int tx, int ty, int tw, int th)
...
+                           _("Path to X509 client certificate"));
+    certInput->align(FL_ALIGN_LEFT | FL_ALIGN_TOP);
+    ty += INPUT_HEIGHT + TIGHT_MARGIN;
+
+    ty += INPUT_LABEL_OFFSET;
+    keyInput = new Fl_Input(tx + INDENT, ty,
+                            width - INDENT * 2, INPUT_HEIGHT,
+                            _("Path to X509 client private key"));
+    keyInput->align(FL_ALIGN_LEFT | FL_ALIGN_TOP);
+    ty += INPUT_HEIGHT + TIGHT_MARGIN;
```
CendioOssman, line 748: These are authentication settings, not encryption. So I think they fit better in the other section.

CendioOssman (changes requested):
Thanks for your contribution! This looks nice and clean.

But there are some user experience issues I think we need to discuss and resolve before we can merge this.

CendioOssman:
These changes are just for the client. What server have you used for this?

opentissandy:
Thanks reviewed by @CendioOssman. I try to fixed the code.
I deployed libvirt/qemu server with Client Certificate VNC based on https://wiki.libvirt.org/VNCTLSSetup.html.
I like use tigervnc nor Vinagre/remmina. So I try to find the way to remove tigervnc from "VNC clients known to NOT work with TLS" completely. Then I find tigervnc Framework is so flexible and gnutls api is simply. In fact ONLY use one gnutls_certificate_set_x509_key_file API to add the last feature to support client TLS fully.
I find remmina UI with CA and client certificate in one page. So I agree move these to new section. But I have no ability to do so big UI change. Can you make a planning? @CendioOssman.
Screenshot_20241008_213637

CendioOssman:
Thanks for the link. So it seems QEMU has had this behaviour for a while, which makes things more problematic.

Which means I'm even more keen on getting some insight from @berrange about these design choices.

Effectively, this means double authentication, which is not the norm in VNC.

vncviewer/vncviewer.man
berrange: > The client has no way of knowing if the server will ask for a client certificate with this approach. That means the client cannot use a different security type, in case it doesn't have a certificate.

That is correct, but at the same time I don't think it is a real world problem. 

If a server has gone to the trouble of enabling VeNCrypt with X509 certs, and decided to mandate client certs, IMHO, it is highly unlikely that the server will also offer other VNC auth options in parallel which don't require certs.  IOW, I don't think the idea that the client could choose a different security type is actually going to happen in practice.

Never say never, of course, but it is highly unlikely IMHO.

opentissandy:
I don't know the VNC or TLS technology details. But I can use "With x509 certificates, client verification and passwords" of QEMU (https://www.qemu.org/docs/master/system/vnc-security.html) with my simply patch code. So I share my idea.

If the project tigervnc leaders can promote the feature moving to better with general effectively design, that is fortunate of users like me.

CendioOssman:
> I find remmina UI with CA and client certificate in one page. So I agree move these to new section. But I have no ability to do so big UI change. Can you make a planning? @CendioOssman.

What issue are you seeing moving the settings? Doesn't it work just moving the code a bit further down to the other section?

berrange:
> > As a general design concept, the way VeNCrypt auth was designed is kind of frowned upon as a way for doing TLS upgrades on a live connection. Best practice design these days would be to either
> > ```
> > * Mandate TLS from the very moment the connection is opening, with no negotiation
> > ```
> 
> I've been thinking about that as well. Partly to handle noVNC clients more cleanly. I'm not sure how to present things in a clean way to the user, though, since we don't work with URLs.

If we do this, effectively it really ceases to be the RFB protocol as we've known it and it is painful to make it compatible with existing clients in a seemless way.

ie if you are told to connect to a port, you don't know if its RFB or RFB-over-TLS. As a client you would have to wait to see if you receive "RFB" bytes from the server, which identifies it as a traditional RFB server, and if none arrive, then start a TLS handshake. This is quite unpleasant as it requires the client wait a non-negligible time in case the initial "RFB" byte signature was slow to arrive.

If you're going to have such a break with the historical protocol, then honestly it would be better to just go all the way to implementing RFB-over-websock fully, which is what the noVNC proxy would do, and thus get TLS via use of HTTPS.

Perhaps it would be worth formally documenting RFB-over-websock in the RFB specification, or as an companion specification ?

> > ```
> > * Support a simple "STARTTLS" mechanism, which negotiates TLS, and then _restarts_ the entire VNC protocol handshake from the beginning, but under TLS.
> > ```
> 
> What practical benefits would we get from that? There's very little VNC protocol before we start VeNCrypt.

The key benefit is that the protocol is still RFB, so a client connecting can automatically "do the right thing" based on what auth is advertized.

clarksx12:
> ly, this means double authentication, whic

x11vnc has also had this feature for years.

CendioOssman:
No response. Closing.

PR #1852, opened 2024-10-18
Support VK_PACKET in vncviewer on Windows

### LINKED ISSUE #1847: VK_PACKET ignored by VNC viewer for Windows
**Describe the bug**
MS Windows has a special virtual keycode (VK) called `VK_PACKET` that indicates that the scan code is to be interpreted as a character code instead. This keycode is frequently used by applications that inject key events. It appears that VNC viewer ignores this keycode and sends no key events to the server. This bug likely also caused #1818.

**To Reproduce**
1. In Keepass 2, create an entry with password ```"a'a^a`a~a```.
2. In Windows, select Swedish keyboard layout.
3. Connect to a remote server using Tiger VNC client and, on the remote server, open a text editor.
4. Use Keepass 2 autotype feature to type the password into the VNC client.
5. Notice the text `aaaaa` appearing in the remote server's text editor.

**Expected behavior**
The text ```"a'a^a`a~a``` should appear.

 - OS: Windows 10 Pro
 - VNC client: TigerVNC
 - VNC client version: 1.12.0

 - OS: Debian
 - VNC server: x11vnc
 - VNC server version: 0.9.16
 - Server downloaded from: NixOS
 - Server was started using: `x11vnc -N -forever -loop100 -usepw -nevershared -clear_keys -v`

**Additional context**
* My guess is that Keepass 2 will use VK_PACKET on any keyboard layout when auto-typing characters not found in that keyboard layout.
* You may add `-pipeinput "reopen:cat - > /tmp/keys.in.fifo"` to the `x11vnc` arguments to get a text stream of the key events received by the server.

svenssonaxel:
Fixes #1847

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -120,6 +120,7 @@ Viewport::Viewport(int w, int h, const rfb::PixelFormat& /*serverPF*/, CConn* cc
     lastPointerPos(0, 0), lastButtonMask(0),
 #ifdef WIN32
     altGrArmed(false),
+    vkPacketHighSurrogate(0),
 #endif
     firstLEDState(true), pendingClientClipboard(false),
     menuCtrlKey(false), menuAltKey(false), cursor(nullptr)
@@ -971,6 +972,12 @@ int Viewport::handleSystemEvent(void *event, void *data)
       keyCode = 0x38;
     }
 
+    // VK_PACKET handling: Translate to WM_*CHAR message, handled below.
+    if (vKey == VK_PACKET) {
+      TranslateMessage(msg);
+      return 1;
+    }
+
     // Windows doesn't have a proper AltGr, but handles it using fake
     // Ctrl+Alt. However the remote end might not be Windows, so we need
     // to merge those in to a single AltGr event. We detect this case
@@ -1086,6 +1093,12 @@ int Viewport::handleSystemEvent(void *event, void *data)
       keyCode = 0x38;
     }
 
+    // VK_PACKET handling: Translate to WM_*CHAR message, handled below.
+    if (vKey == VK_PACKET) {
+      TranslateMessage(msg);
+      return 1;
+    }
+
     // We can't get a release in the middle of an AltGr sequence, so
     // abort that detection
     if (self->altGrArmed)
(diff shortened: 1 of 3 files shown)


vncviewer/Viewport.cxx
```diff
@@ -971,6 +972,12 @@ int Viewport::handleSystemEvent(void *event, void *data)
       keyCode = 0x38;
     }
 
+    // VK_PACKET handling: Translate to WM_*CHAR message, handled below.
+    if (vKey == VK_PACKET) {
+      TranslateMessage(msg);
```
CendioOssman, line 977: I'd rather avoid the risk of intercepting another message type. Was there not enough information in the `WM_KEYDOWN` event? `TranslateMessage()` is able to make sense of it somehow, so there should be something there.

vncviewer/Viewport.cxx
```diff
@@ -1118,6 +1131,55 @@ int Viewport::handleSystemEvent(void *event, void *data)
...
+             (msg->message == WM_SYSCHAR) || (msg->message == WM_SYSDEADCHAR)) {
+    // Windows will send VK_PACKET key events if it doesn't have a physical key
+    // associated with the event (e.g. from a mobile device virtual keyboard).
+    // After translating a keydown/keyup event pair with VK_PACKET as vKey, a
+    // WM_*CHAR message will be issued containing a UTF-16 code unit. Unicode
+    // characters outside of the range U+0000 - U+ffff (Basic Multilingual
+    // Plane, BMP), are encoded as a pair of UTF-16 code units that will need to
+    // be synthesized into one Unicode code point.
+    uint32_t ucsCode = msg->wParam;
+    if ((ucsCode & 0xfc00) == 0xd800) {
```
CendioOssman, line 1145: Could the surrogate handling be a separate commit? Makes it easier to see which parts are connected.

vncviewer/Viewport.cxx
```diff
@@ -1118,6 +1131,55 @@ int Viewport::handleSystemEvent(void *event, void *data)
...
+      if (self->vkPacketHighSurrogate) {
+        vlog.error(_("Unmatched UTF-16 surrogate pair through VK_PACKET, "
+                     "codes: 0x%04x, U+%04x"),
+                   self->vkPacketHighSurrogate, ucsCode);
+        self->vkPacketHighSurrogate = 0;
+      }
+      codePoint = ucsCode;
+    }
+    uint32_t keySym = ucs2keysym(codePoint);
+    uint32_t keyCode = 0x100 + keySym; // Fake key code
```
CendioOssman, line 1180: Keysym is 32-bit, so this is not going to work. Perhaps something similar to the universal keysyms? I.e. `0x01000000 | codePoint`?

vncviewer/keysym2ucs.c
```diff
@@ -167,8 +167,8 @@ unsigned ucs2keysym(unsigned ucs)
   if (keysym != NoSymbol)
     return keysym;
 
-  /* us the directly encoded 24-bit UCS character */
-  if ((ucs & 0xff000000) == 0)
+  /* ucs is a directly encoded 21-bit Unicode character */
+  if (ucs <= 0x10ffff && ((ucs & 0xfff800) != 0x00d800))
```
CendioOssman, line 171: I guess this is a safety net? Good idea, but I think it is a bit hidden here. Perhaps instead:

```diff
diff --git a/vncviewer/keysym2ucs.c b/vncviewer/keysym2ucs.c
index 6607e3065..ad9b7ebf3 100644
--- a/vncviewer/keysym2ucs.c
+++ b/vncviewer/keysym2ucs.c
@@ -167,6 +167,14 @@ unsigned ucs2keysym(unsigned ucs)
   if (keysym != NoSymbol)
     return keysym;
 
+  /* surrogates? */
+  if (ucs >= 0xd800 && ucs <= 0xdfff)
+    return NoSymbol;
+
+  /* private use? */
+  if (ucs >= 0xe000 && ucs <= 0xf8ff)
+    return NoSymbol;
+
   /* us the directly encoded 24-bit UCS character */
   if ((ucs & 0xff000000) == 0)
     return ucs | 0x01000000;

```

CendioOssman (changes requested):

vncviewer/Viewport.cxx
svenssonaxel: I analyzed a hex dump of the entire `WM_KEYDOWN` message, and there is nothing there. `TranslateMessage()` is a Windows API call and need not be limited to the information contained in the parameter. Perhaps there is something in adjacent memory, but even if so it'd be riskier to access that than to use the provided API.

vncviewer/Viewport.cxx
svenssonaxel: Good point, will change to what you suggested. Another option is actually a constant, e.g. `keyCode = 0x123`, since we expect only one concurrent `VK_PACKET` key down.

vncviewer/keysym2ucs.c
svenssonaxel: Will make more explicit as you suggest, but I will leave the 0x10ffff limit since code points above that values are not valid.

vncviewer/Viewport.cxx
svenssonaxel: I could, but only by making a first commit that exhibits wrong behavior w.r.t. surrogate pairs. Would you like that?

svenssonaxel:
@CendioOssman Addressed your review comments and pushed a fix. Once you decide what you want me to do with commit splitting, I'll clean up the commit history.

vncviewer/Viewport.cxx
CendioOssman: I played around with it here and I could not find anything in the event either.

But I was getting proper symbols back from `win32_vkey_to_keysym()`. The complexity from `TranslateMessage()` might not be needed.

Did you test that path?

CendioOssman: Ping @svenssonaxel

CendioOssman:
No response. Closing.

vncviewer/Viewport.cxx
svenssonaxel: Correct, `TranslateMessage()` was not needed. Rewritten as you suggested.

vncviewer/Viewport.cxx
svenssonaxel: ping @CendioOssman

svenssonaxel:
@CendioOssman @Neustradamus Rebased and rewrote as suggested without `TranslateMessage()`. Pushed commit `b617739b` to branch `svenssonaxel/vkpacket`, perhaps you have to reopen the PR to see it.

svenssonaxel:
Pushed `3f0f7bbb`

Issue #949, opened 2020-01-29
Cannot build vncviewer without NLS


Peter2121:
I'm trying to build tigervnc from Git on Mac OS X 10.13.6. It builds correctly with the default options, but when I try to build without NLS - it's failed.

CMAKE command:
`cmake -G "Unix Makefiles" -DENABLE_NLS=false ..`

Trying to build I get many errors like this:
```
/Users/peter/Programming/tigervnc/vncviewer/Viewport.cxx:1240:30: error: use of undeclared identifier 'LC_MESSAGES'
  fltk_menu_add(contextMenu, p_("ContextMenu|", "&Options..."),
```

CendioOssman:
I can reproduce the issue here as well. It seems gettext's standard header assumes that the locale defines are present even if gettext is missing.

Will need to do some form of workaround.

PR #1854, opened 2024-10-20
Add Nix flake

svenssonaxel:
After much failure trying to follow `BUILDING.txt` instructions, I only managed to build using Nix. Contributing a `flake.nix` in case others have similar trouble.

* Supports one-line build for
  * linux x86_64
  * linux aarch64 (not tested)
  * darwin x86_64 (not tested)
  * darwin aarch_64 (not tested)
  * java
  * windows 64-bit
  * windows 32-bit

Marking as **draft** due to a couple of questions:
* NLS support for Windows (separate commit) creates a `vncviewer.exe` with a hard-coded path pointing to `gettext` language support files. Not sure why.
* Not sure if we want to support release builds using Nix.
* Windows builds are almost entirely static, using no `*.dll` files except for `wm_hooks.dll`. Do we want the other builds to also be static?

### DIFF
--- a/BUILDING.txt
+++ b/BUILDING.txt
@@ -404,3 +404,32 @@ Distribution-Specific Packaging
 
 Configuration for building packages for current versions of RHEL/CentOS
 and Ubuntu LTS can be found under contrib/packages.
+
+
+===============
+Build using Nix
+===============
+
+The perhaps easiest way to build TigerVNC is using Nix.
+
+To install Nix on Linux, macOS or WSL:
+
+  curl --proto '=https' --tlsv1.2 -sSf -L \
+    https://install.determinate.systems/nix | sh -s -- install --determinate
+
+To build native binaries for the current system (supports linux and darwin on
+x86_64 and aarch64):
+
+  nix build .
+
+To build for 64-bit Windows:
+
+  nix build .#windows
+
+To build for 32-bit Windows:
+
+  nix build .#windows32
+
+To build the Java version of the TigerVNC Viewer:
+
+  nix build .#java
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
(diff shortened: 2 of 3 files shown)


CendioOssman:
I'm not too keen on another build script to maintain. So first I'd like to explore where you got stuck, and see if we can ease that hurdle?

svenssonaxel:
Sure. I've already provided a Dockerfile explaining where I get stuck. Attached again here:

[Dockerfile-debian.txt](https://github.com/user-attachments/files/17405660/Dockerfile-debian.txt)

Run it as `docker build -f Dockerfile-debian.txt .`. If you can edit this file so that building works, perhaps even simplify it, I think you'll have produced better documentation already.

I suppose my point is that this PR is not "another build script", that's in `CMakeLists.txt`. This is more like an executable description of dependencies, something that currently does not exist at all. The Dockerfile fulfills (or would, if fixed) the same function, in less pure but perhaps more accessible way.

CendioOssman:
The Dockerfile is based on Debian, which does explain a lot of the difficulties. Unfortunately, they don't have a good MinGW environment to use. So you have to set up one from scratch, with all the difficulties that entails.

Using Arch or Fedora makes the process much easier as you can just install all the needed tools and libraries from their repositories.

MXE is available for Debian/Ubuntu, though, which could make things easier.

The main issue with the given Dockerfile seems to be that pkg-config isn't properly used. Which is how the build system finds most things. Usually, you specify the cross-compiler pkg-config to CMake (just like you specify the cross-compiler). But since there isn't one, you'll need to use the normal pkg-config with a different search path:

```
RUN cd tigervnc && \
    mkdir build && \
    cd build && \
    PKG_CONFIG_LIBDIR=/usr/x86_64-w64-mingw32/lib/pkgconfig:/usr/x86_64-w64-mingw32/share/pkgconfig \
    $CMAKE -G "Unix Makefiles" \
        ..
```

At this point it works much better. It still fails for me because we apparently have a bug on `master` if GnuTLS is missing. And we have #949 if gettext is missing. If those two are added, then I would assume things build just fine.

PR #1864, opened 2024-11-04
Don't use login as a verb in Xvnc.man

ThinLinc-Zeijlon:
The title says it all :smile:


CendioOssman:
Thanks! It looks like the section might need a reflow after the change, though?

ThinLinc-Zeijlon:
I agree! Reworded the paragraph a little bit in my latest commit.

PR #1868, opened 2024-11-12
Move general common code to "core" library

CendioOssman:
The `common/` structure has been a bit messy in some cases because some general infrastructure was located in `common/rfb`. Besides the confusion about the purpose of each library, it also created dependency loops, creating issues such as #1691.

Solve this by moving generic support code to a new library called "core" that serves as a base for everything else.

I've tried to perform this move in a way that minimises the diffs, and hence damage to `git blame`.


KangLin:
I think it's okay to just move the cross-platform part and keep the structure of the other RFBs

common/core/Thread.h
KangLin, line 1: I'd recommend replacing the existing Thread with std::thread.

common/core/xdgdirs.h
KangLin, line 1: This is used by the utilities, not core. Therefore, it is advisable to create a separate directory: Utilities

common/core/Mutex.h
KangLin, line 1: It is recommended to use std::c++11

common/core/Thread.h
CendioOssman: We ran in to some compatibility problems with older MinGW and gcc. So it's on its way, but not yet.

common/core/xdgdirs.h
CendioOssman: The core library already has a bunch of general utility things. E.g. the string helpers. So I think this fits in well enough to avoid more complexity.

CendioOssman:
The latest shuffling caused the commit "Move configuration to core library" to fail to build. The problem is the dependency loops, and that commit just got unlucky.

We might get lucky shuffling the order around, but we also might not. I don't know if it's worth the extra effort or if we can live with one broken commit.

common/core/Thread.h
KangLin: stdc++11 has been out for ten years. It doesn't make sense to support compilers older than that. Just add complexity to the code.

KangLin:
@CendioOssman 
It is recommended use [C++11]((https://cplusplus.com/reference/)) atomics and threading library.
- Replace thread with [std::thread](https://cplusplus.com/reference/thread/)
- Replace Mutex with [std::Mutex](https://cplusplus.com/reference/mutex/)
- Replace Condition with [std::condition_variable](https://cplusplus.com/reference/condition_variable/)

PR #1878, opened 2024-11-24
Update it.po

albanobattistella:


CendioOssman:

Unfortunately, we cannot accept translation changes directly. Our translation files are managed by the translation project:

https://translationproject.org/domain/tigervnc.html

Please submit your updates to their Italian team and it should reach us the proper way:

https://translationproject.org/team/it.html

Issue #931, opened 2020-01-02
keyboard grab not released on secondary monitors on macOS


qijialiu:
TigerVNC Viewer 64-bit v1.10.1
macOS  Catalina version 10.15.2

I have two monitors. With VNC viewer running on one of the monitors in full-screen mode and taking system keys directly to server.

When my mouse goes to the other monitor without VNC, the mouse clip works there just instantaneously, but the focus remains at the VNC viewer, as any keyboard input will be directed into the VNC viewer.

In another thread below, a similar issue (on ubuntu but not mentioned on macOS) was resolved. But I'm still having the issue with v1.10.1.
https://github.com/TigerVNC/tigervnc/issues/447

CendioOssman:
I can reproduce this here as well. Not sure what the issue is right now.

Unfortunately #447 was a bit specific to how X11 behaves, so it's not surprising that it didn't have any effect on macOS.

CendioOssman:
Hmm.. This worked fine when testing on macOS 11. Perhaps it was a OS bug rather than something in TigerVNC?

@qijialiu, are you still seeing this problem?

qijialiu:
Pierre,

I'm still with 10.57. Will check about the issue after updating to 11.

Thanks for updating. It sounds good if the issue can be resolved by new macOS.

Thanks,
Qijia
________________________________

CendioOssman:
This should be fixed by 72e94ca93d3ada8fb7f518fa65644d2bc172bd4b, which is also in the new 1.12.0 release.

williamsjoblom:
Despite the alleged fix in 72e94ca I can still reproduce the issue on macOS 12.0.1.

I have tried the following build variations with seemingly identical broken behavior:
- The TigerVNC 1.12.0 release build
- HEAD of the master branch built with FLTK 1.3.7.
- At the alleged fix (72e94ca) with FLTK 1.3.7 and 1.3.5.
- The commit before the alleged fix with FLTK 1.3.7 and 1.3.5.

So, as far as I can tell, this fix externally looks like a NOOP, so we should probably reopen this issue @CendioOssman. Worth noting is that I can only reproduce the issue with vncviewer full-screened on the *primary* monitor (applies to all builds above).

samhed:
[...]

PR #1884, opened 2024-12-08
Use event taps for keyboard grab on macOS

CendioOssman:
These are the slightly more proper way to grab the keyboard and macOS and is what similar applications do. It avoids a lot of issues we have, e.g., problems with multiple monitors.
    
Unfortunately we need to have the user explicitly approve this (which really is a good thing, security wise), and Apple have chosen to mark this feature as only for accessibility.

Pre-requisite for windowed keyboard grabbing in #1375.

### DIFF
--- a/vncviewer/DesktopWindow.cxx
+++ b/vncviewer/DesktopWindow.cxx
@@ -698,10 +698,6 @@ void DesktopWindow::resize(int x, int y, int w, int h)
 
     repositionWidgets();
   }
-
-  // Some systems require a grab after the window size has been changed.
-  // Otherwise they might hold on to displays, resulting in them being unusable.
-  maybeGrabKeyboard();
 }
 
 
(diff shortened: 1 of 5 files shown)


CendioOssman:
One user visible change from this is that touch pad gestures will no longer be blocked. E.g. three finger swipe to switch virtual desktop. This is because we now decide each event type to intercept and block/redirect.

The swipe gesture isn't documented, but I've experimentally discovered it is using type `30`.

There would not be any added functionality right now by blocking this, as we don't do anything useful with this event. On the other hand, in some utopian future we'd like to pass these events on to the remote side. E.g. controlling another macOS system, and we want to switch virtual desktop on the remote side.

If we don't block it right now, then we'd have to annoy users by changing behaviour and blocking it once we have that feature. But if we block it now, then we are removing local functionality without giving the users something else in return.

CendioOssman:
> Unfortunately we need to have the user explicitly approve this (which really is a good thing, security wise), and Apple have chosen to mark this feature as only for accessibility.

The text message from the system for this is a bit poor and vague. Fortunately, the system popup doesn't do much and we can easily replace it with our own:

```objc
    // Translators: "Security & Privacy" and "Accessibility" are from
    //              System Settings in macOS and should match the
    //              translations there
    ret = fl_choice(_("TigerVNC requires accessibility permission to pass system keys to the server.\n"
                      "Enable TigerVNC in Security & Privacy > Accessiblity."),
                    nullptr, _("Open System Settings"), _("Not now"));
    if (ret == 1) {
      NSString* prefPage = @"x-apple.systempreferences:com.apple.preference.security?Privacy_Accessibility";
      [[NSWorkspace sharedWorkspace] openURL:[NSURL URLWithString:prefPage]];
    }
```

The pseudo-synchronous nature of `fl_choice()` is causing some issues, though. I think I'll keep the system popup for now and we can improve this later.

CendioOssman:
This is now working well and ready to be merged.

CendioOssman:
This fixes #931 as well.

Issue #1583, opened 2023-01-11
Interactive toggle of ViewOnly mode


freonheat:
It would be fantastic if TigerVNC could support an interactive option for toggling if the client is in ViewOnly mode or not.

We use TigerVNC for remote support and there are many times we don't want to interfere with the remote users' screen, unless needed.

I propose a "-ToggleViewOnly" flag that will enable a menu option in the local F8 menu to toggle into or out of ViewOnly mode.  This option can be combined with "-ViewOnly" on launch so the session starts ViewOnly and then can be toggled to be non-ViewOnly through the menu.  Or, started without "-ViewOnly", and then use F8 to change into ViewOnly later, if needed.

PR #1906, opened 2025-02-06
Force grab/ungrab keyboard+mouse with Right Ctrl key

phd:
Force grab/ungrab keyboard+mouse with Right Ctrl key,
Allows for control integration similar to what Virtual Box does.

Solves pointer movement glitches with applications that require grabbing the mouse internally (like modern web browsers using the Pointer Lock API).

Test:
1. Open any modern web browser inside TigerVNC.
2. Navigate to https://mdn.github.io/dom-examples/pointer-lock/
3. Move the mouse far (off-screen) in any direction.

Without this patch the above website works only in TigerVNC's fullscreen mode.


phd:
I have no experience with the VNC protocol nor its extensions.
Just wanted to fix the Pointer Lock API issue and this solution works well for me while being very intuitive. That's why I wanted to share it.
It does not affect user experience in any way other than adding this key binding, so it won't pester users not willing to use it.
Meanwhile it allows to comfortably use TigerVNC with clients that require Pointer Lock API or other forms of pointer positioning. And it does it right now. When a proper solution is implemented this could be removed (or reduced to forced grab/ungrab which could still be useful e.g. for handling system-wide shortcuts).

CendioOssman:
The test case you have seems to work fine with the existing support in TigerVNC for relative mouse movements. So I think I misunderstood what this PR actually changes.

The focus is rather to allow grabs in windowed mode. The big problem with that is adding more key bindings means more risk of conflict with what applications need. Have you looked at the existing discussion in #102? And the work in progress in #1375?

The proposed plan there tries to minimize the conflict, and provides a way out in case a conflict still arises.

phd:
> The test case you have seems to work fine with the existing support in TigerVNC for relative mouse movements. So I think I misunderstood what this PR actually changes.

It only works in fullscreen mode, where mouse grab is automatic. In windowed mode it fails when the mouse goes too far in any direction.

>The focus is rather to allow grabs in windowed mode.

Yes, indeed. I wanted to have control over when TigerVNC accepts my input. Similar to what VirtualBox and (as noted in #102) Qemu do using the Right Ctrl key (seems to be quite a standard key binding).

I have added more commits to support when keyboard and/or mouse grabs are *required* for input to be accepted. This way I cannot mess up the remote desktop without doing the grab first.
Hence the GrabOnlyKeyboard and GrabOnlyMouse options (because sometimes keyboard input without the grab is ok for me).

Actually this PR provides a nice weak read-only mode, which is easy to toggle with Right Ctrl. This way TigerVNC is perfect for my needs.

>The big problem with that is adding more key bindings means more risk of conflict with what applications need

It's true that creating key binding conflicts is bad. I like how #1375 allows to rebind everything.
It also adds a hot key binding for grabbing and ungrabbing the keyboard, but it lacks the features this PR provides (like GrabOnlyKeyboard and GrabOnlyMouse).

phd:
Added:
Option to disable grabbing mouse+keyboard with Right Ctrl
New option: GrabToggleWithRightCtrl (on by default)

CendioOssman:
I'm not quite sure what the extra feature with the "GrabOnly" settings are. Is it largely a convenient way to toggle "view only" on and off?

phd:
> I'm not quite sure what the extra feature with the "GrabOnly" settings are. Is it largely a convenient way to toggle "view only" on and off?

Somewhat. However it allows to apply the "view only" to keyboard and mouse input separately: GrabOnlyKeyboard, GrabOnlyMouse.
It also disables this "view only"-like mode in more ways than just with a hotkey (GrabToggleWithRightCtrl) - particularly and most useful GrabWithMouseClick, and also GrabToggleWithMiddleButton.

It is also different to the original "view only" mode in that it acts selectively on user input while allowing every other form of interaction like clipboard being set both ways.

CendioOssman:
Could you describe your workflow a bit more that needs this kind of fine-grained control? We want to keep things understandable. Which means we would like to keep the number of settings to a minimum.

phd:
> Could you describe your workflow a bit more that needs this kind of fine-grained control?

I'm running several web-based apps remotely and view them via VNC on multiple local virtual desktops. Some of the apps require Pointer Lock API, hence the need for mouse grab.

I want to run VNC clients windowed, not fullscreen.

VNC clients must not accept any mouse move input just because the mouse pointer happened to be over them (e.g. after switching virtual desktop) - that will mess the web apps for everyone connected.
So I need a view-only mode. One that is easy to disable, preferably with a simple mouse click (but that mouse click event must be eaten and not sent forward to the client) and automatically switch into the input-grab mode.

I'm currently running this patched version of TightVNC.
I've added the GrabToggleWithRightCtrl option first as a common way of going in and out of such virtual environments (see VirtualBox, QEMU). But I do not use it anymore, because I find GrabWithMouseClick and GrabToggleWithMiddleButton much more convenient.
I've personally found it hard to decide between GrabWithMouseClick=1 and 2 (eating the click event or not) for some of the web-apps, because each of them has a different use case. And I'm sure users may want to be able to configure this to fit their own workflow.

> We want to keep things understandable. Which means we would like to keep the number of settings to a minimum.

I don't think lots of useful options are a bad thing. They allow users to configure and fine-tune the program to meet their needs.
What is important are sane defaults (i.e. turning most of these off by default - currently only GrabToggleWithRightCtrl is enabled, with the ability to turn it off in case of conflicts).

phd:
This seems to solve several issues: #2, #102, #619 (?), #1583

emma592:
Tested this on Ubuntu 24.04.1 and I might have misunderstood the purpose of this. 

Ran with `--GrabWithMouseClick=1` and managed to get the window title [GRAB], however the mouse does not seem grabbed and behaves the same as non-modified tigervnc allowing the cursor to exit the vnc window and interact with other windows.

emma592:
Thank you! Using X11 works!

phd:
> We want to keep things understandable. Which means we would like to keep the number of settings to a minimum.

I guess the multiple `GrabOnly*` options could be replaced with one enum (flags?).
That would be more understandable and reduce the number of settings.
That makes 4 new options:

- `GrabOnly` enum/flags: `Keyboard`, `Mouse`, `Clipboard` (and combinations)
- `GrabToggleWithRightCtrl`
- `GrabToggleWithMiddleButton`
- `GrabWithMouseClick`

`GrabToggleWithRightCtrl` and `GrabToggleWithMiddleButton` could become one option as well.
That leaves 3 new options:

- `GrabOnly` enum/flags: `Keyboard`, `Mouse`, `Clipboard` (and combinations)
- `GrabToggleWith`: `RightCtrl`, `MiddleButton`, `Both`
- `GrabWithMouseClick`

And `GrabWithMouseClick` could probably be merged as well, leaving only 2 new options:

- `GrabOnly` enum/flags: `Keyboard`, `Mouse`, `Clipboard` (and combinations)
- `GrabWith`: `RightCtrl`, `MouseMiddleButton`, `MouseClick`, `MouseClickSupressed` (and combinations)

These options seem quite understandable and at least do not litter the `--help` message:
- one for defining a way to grab/ungrab the input,
- the other for limiting what kind of input is ignored without a grab (which is very different from a view-only mode).

phd:
Updated.

```
  GrabOnly       - Input that should be allowed only when mouse and keyboard
                   are grabbed (None, Keyboard, Mouse, Clipboard, All)
                   (default=None)
  GrabWith       - Choose ways to grab keyboard and mouse with (None,
                   RightCtrl, MouseMiddleButton, MouseClick,
                   MouseClickSupressed) (default=RightCtrl)
```

This looks much more sane :)

emma592:
Just my 2 cents as I monitor this thread. GrabOnly leads my thoughts to only grabbing specific inputs, not only considering specific inputs while grabbed.

As an example, WhileGrabbedOnly/OnlyWhileGrabbed would be clearer to me.

phd:
> OnlyWhileGrabbed
+1

PR #1908, opened 2025-02-07
Add SELinux policy rules allowing to create directories under /root

grulja:
We have policy that allows to create ~/.local or ~/.config, but we don't have rule that allows the same under /root directory, where we fail in case any of these directories doesn't exist.

### DIFF
--- a/unix/vncserver/selinux/vncsession.te
+++ b/unix/vncserver/selinux/vncsession.te
@@ -48,6 +48,14 @@ optional_policy(`
 	create_dirs_pattern(vnc_session_t, gconf_home_t, gconf_home_t)
 ')
 
+# Allowed to create /root/.local
+optional_policy(`
+	gen_require(`
+		type admin_home_t;
+	')
+	create_dirs_pattern(vnc_session_t, admin_home_t, admin_home_t)
+')
+
 # Manage TigerVNC files (mainly ~/.local/state/*.log)
 create_dirs_pattern(vnc_session_t, vnc_home_t, vnc_home_t)
 manage_files_pattern(vnc_session_t, vnc_home_t, vnc_home_t)
@@ -88,6 +96,7 @@ optional_policy(`
 	gen_require(`
 		attribute userdomain;
 		type gconf_home_t;
+		type admin_home_t;
 	')
 	userdom_admin_home_dir_filetrans(userdomain, vnc_home_t, dir, ".vnc")
 	userdom_user_home_dir_filetrans(userdomain, vnc_home_t, dir, ".vnc")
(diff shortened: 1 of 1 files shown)


unix/vncserver/selinux/vncsession.fc
```diff
@@ -18,12 +18,19 @@
 #
 
 HOME_DIR/\.vnc(/.*)?      gen_context(system_u:object_r:vnc_home_t,s0)
-/root/\.vnc(/.*)?      gen_context(system_u:object_r:vnc_home_t,s0)
+HOME_DIR/\.config      gen_context(system_u:object_r:user_home_t,s0)
```
zpytela, line 21: You should not use types from other modules. Applies to multiple places in this PR.

unix/vncserver/selinux/vncsession.te
```diff
@@ -31,6 +31,7 @@ type vnc_home_t;
 userdom_user_home_content(vnc_home_t)
 
 allow vnc_session_t self:capability { chown dac_override dac_read_search fowner kill setgid setuid sys_resource };
+allow vnc_session_t self:dir { create write setattr };
```
zpytela, line 34: This means the service handles files in /proc manually?

grulja: Is this about `/proc`? I assumed it just allows to create directories I specified in the other file.

unix/vncserver/selinux/vncsession.fc
grulja: I see. I thought I have to specify what types should be used for parent directories and that using `vnc_home_t` will create them with a wrong context.

grulja:
So I think this is close to be working, but it's still missing one thing. The code I added now allows to create `/root/.local/state/tigervnc`, but `tigervnc` folder is labeled as `admin_home_t` and I don't know how to change that. 
Obviously running `restorecon -Rv /root/.local/state/tigervnc` fixes it and it starts to work.

I tried adding:
```
diff --git a/unix/vncserver/selinux/vncsession.te b/unix/vncserver/selinux/vncsession.te
index e1ed6c11..ce751df9 100644
--- a/unix/vncserver/selinux/vncsession.te
+++ b/unix/vncserver/selinux/vncsession.te
@@ -96,7 +96,9 @@ optional_policy(`
        gen_require(`
                attribute userdomain;
                type gconf_home_t;
+               type admin_home_t;
        ')
+       userdom_admin_home_dir_filetrans(userdomain, vnc_home_t, dir, ".local")
        userdom_admin_home_dir_filetrans(userdomain, vnc_home_t, dir, ".vnc")
        userdom_user_home_dir_filetrans(userdomain, vnc_home_t, dir, ".vnc")
 ```
 But that's not it.
 
 @zpytela would you know please? Thank you.

grulja:
Alright, after reading some documentation and testing it now seems to work as expected.

CendioOssman (approved):
Thanks! Can confirm it fixes the issue here as well. Tested on Fedora 41.

PR #1916, opened 2025-02-21
Clear clipboard state when invalid UTF-8 string

CendioHalim:
This commit adds another check to ensure that we don't announce clipboard data when the data is invalid.

Running `xclip -sel clip -t STRING image.png` for example would trigger a clipboard announce with invalid content.

This is the same fix as commit 6f6d9406035d59b6a9f59ff0f38206c0c98a0266.

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -411,6 +411,14 @@ int Viewport::handle(int event)
   case FL_PASTE:
     if (!isValidUTF8(Fl::event_text(), Fl::event_length())) {
       vlog.error("Invalid UTF-8 sequence in system clipboard");
+      // Reset the state as if we don't have any clipboard data at all
+      this->pendingClientClipboard = false;
+      try {
+        this->cc->announceClipboard(false);
+      } catch (std::exception& e) {
+        vlog.error("%s", e.what());
+        abort_connection_with_unexpected_error(e);
+      }
       return 1;
     }
 


vncviewer/Viewport.cxx
```diff
@@ -411,6 +411,14 @@ int Viewport::handle(int event)
   case FL_PASTE:
     if (!isValidUTF8(Fl::event_text(), Fl::event_length())) {
       vlog.error("Invalid UTF-8 sequence in system clipboard");
+      // Reset the state as if we don't have any clipboard data at all
+      this->pendingClientClipboard = false;
+      try {
+        this->cc->announceClipboard(false);
```
CendioOssman, line 417: I suppose this is better than nothing, but it's not ideal as we'll end up announcing that we have a clipboard initially, only to back down from that.

Should we check things earlier? And/Or send an empty string so we at least send something (like we promised)?

CendioOssman (changes requested):

vncviewer/Viewport.cxx
CendioHalim: Looks like we already have code in `vncselection.c` :`vncHandleClipboardAnnounce()` that handles aborting a clipboard announce, which is why this fix works.

CendioHalim:
> I wonder if this also means a bug in FLTK. We should only be getting UTF-8 from it. Did you find why it is giving us bad data?

It looks like it. I put a Latin-1 encoded string ("åäö") into my clipboard with a small X11 application. When I paste inside vncviewer, I get garbage '???' characters instead of "åäö".

I got the same behaviour with GTK applications, so this is not something unique to FLTK.

X11 says to only store Latin-1 encoding strings in the `STRING` target [1], so I guess it's up to the applications/toolkits to do the necessary conversion? Nonetheless, this is probably not a common issue in practice since most toolkits & apps use/prefer `UTF8_STRING` anyway.

[1] https://x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html

CendioOssman (approved):

PR #1917, opened 2025-02-27
MSVC: fix compile error in Socket.cxx

KangLin:


CendioOssman:
Is this a new issue? I can't see that these includes changed recently.

KangLin:
msvc is not follow head files:

#include <unistd.h>
#include <signal.h>
#include <fcntl.h>
#include <errno.h>

so include !win32

CendioOssman:
Yes, I figured as much. But those includes have been around for years. How come it is an issue now and not earlier?

PR #1918, opened 2025-02-27
Add SELinux policy rules allowing to access /proc/sys/fs/nr_open

grulja:
This is needed when the nofile limit is set to unlimited, otherwise we will fail to start a VNC session.

### DIFF
--- a/unix/vncserver/selinux/vncsession.te
+++ b/unix/vncserver/selinux/vncsession.te
@@ -34,6 +34,14 @@ allow vnc_session_t self:capability { chown dac_override dac_read_search fowner 
 allow vnc_session_t self:process { getcap setexec setrlimit setsched };
 allow vnc_session_t self:fifo_file rw_fifo_file_perms;
 
+optional_policy(`
+        gen_require(`
+                type sysctl_fs_t;
+        ')
+        allow vnc_session_t sysctl_fs_t:dir search;
+        allow vnc_session_t sysctl_fs_t:file { getattr open read };
+')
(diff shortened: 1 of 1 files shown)


grulja:
This is supposed to fix https://issues.redhat.com/browse/RHEL-77973. I tested this with both a regular and root users.

You can reproduce it by setting `nofile` limit to `unlimited` in `/etc/security/limits.conf`.

CendioOssman:
Thanks for the fix!

It is a bit non-SPOT that every service used by PAM has to duplicate permissions like this. Is there any work being done to provide an interface macro that coordinates permissions that PAM-modules might need? Probably a question for @zpytela rather than @grulja.

zpytela:
@grulja The result is fine, but you should rather use `kernel_read_fs_sysctls(vnc_session_t)` - that's what interfaces are for. You can get a hint with `audit2allow -R`, not always the best option, but good for the suggestion where to look for the right interface.

@CendioOssman You are generally right, it just needs to be assessed if the added permissions for possibly dozens of services, which do not typically need it, are the good way of addressing the known problem and preventing future ones. I am not telling either way right now.

grulja:
> @grulja The result is fine, but you should rather use `kernel_read_fs_sysctls(vnc_session_t)` - that's what interfaces are for. You can get a hint with `audit2allow -R`, not always the best option, but good for the suggestion where to look for the right interface.
> 

Thanks. I will try to come up with a follow-up next week.

PR #1923, opened 2025-03-03
SELinux: Use kernel_read_fs_sysctls interface instead custom policy

grulja:
The kernel_read_fs_sysctls interface does exactly the same thing and we don't reinvent the wheel. It's also easier to maintain.

Supersedes #1918


unix/vncserver/selinux/vncsession.te
```diff
@@ -34,12 +34,10 @@ allow vnc_session_t self:capability { chown dac_override dac_read_search fowner
...
 
+# Allow access to /proc/sys/fs/nr_open
+# Needed when the nofile limit is set to unlimited.
 optional_policy(`
-        gen_require(`
-                type sysctl_fs_t;
-        ')
-        allow vnc_session_t sysctl_fs_t:dir search;
-        allow vnc_session_t sysctl_fs_t:file { getattr open read };
+    kernel_read_fs_sysctls(vnc_session_t)
```
zpytela, line 40: kernel_* should not appear in an optional block, it should be moved too (current) line 46

zpytela:
Correct, also improves legibility and is just the correct way: For using types from other modules, including base, interfaces should be used. 

Now similar changes should also apply to other parts, it may not be that easy this time.

grulja:
> Now similar changes should also apply to other parts, it may not be that easy this time.

What do you mean by that?

zpytela:
:+1:

zpytela:
> > Now similar changes should also apply to other parts, it may not be that easy this time.
> 
> What do you mean by that?

For using types from other modules interfaces should be used. No direct reference to gconf_home_t and admin_home_t.

grulja:
> > > Now similar changes should also apply to other parts, it may not be that easy this time.
> > 
> > 
> > What do you mean by that?
> 
> For using types from other modules interfaces should be used. No direct reference to gconf_home_t and admin_home_t.

I see. The use of `gconf_home_t` was introduced by https://github.com/TigerVNC/tigervnc/commit/337c136ce018fc8d9d12b9e188634925238d11cc, but `admin_home_t` is what I added recently. 

What complications this can cause? Or is it just the right thing to do?

zpytela:
> What complications this can cause? Or is it just the right thing to do?

To ensure types from other modules are handled properly, e. g. when a particular module is disabled.

Certainly it is the right thing, improves legibility and maintenance etc.

Issue #1927, opened 2025-03-11
macOS and Windows app/bundle/shortcut not translated


CendioOssman:
**Describe the bug**
The launcher for TigerVNC is always in English on macOS and Windows.
(as is other metadata, like the copyright notice)

**To Reproduce**
1. Install TigerVNC
2. Open the start menu/Applications

**Expected behavior**
The application name is in the system language.

**Client (please complete the following information):**
 - OS: Windows and macOS
 - VNC client: TigerVNC
 - VNC client version: 1.15.0
 - Client downloaded from: sourceforge

**Additional context**
Broken out of #1317.

PR #1928, opened 2025-03-11
Rename viewer to just "TigerVNC", and clean up packaging builds

CendioOssman:
Most users will not care that there is also server components in our project, so the "viewer" or "client" qualifier is just unnecessary noise.

Also a bit of a various cleanups in how release packages are generated.


CendioOssman:
Fixes #1927 by side stepping the issue

KangLin:
It's better to keep the original name. The server is one of the features of this project that I pay attention to.

CendioOssman:
You're a developer, though, so I think you'll manage. :)

The target audience for this is casual users. Many of which will only use the client.

bphinz:
I don't see any show stoppers

PR #1929, opened 2025-03-12
Refresh framebuffer on options change

phd:
Refresh framebuffer on options change.

ISSUE: After quality or color level options were modified only screen regions that changed since were repainted leaving unchanged regions painted using old settings.

Picture attached below.
Remote server is playing a YouTube movie in a browser.
Color level option is changed in TigerVNC viewer's OptionsDialog from Full to Very Low.
Only regions of the screen where the movie picture changed on the remote server were sent for updating and thus were repainted using the new Very Low color setting.
This resulted in TigerVNC window with both previously used Full color and the new Very Low color regions:

quality-color-level-options-modified-only-screen-regions-that-changed-since-repainted

### DIFF
--- a/vncviewer/CConn.cxx
+++ b/vncviewer/CConn.cxx
@@ -584,6 +584,8 @@ void CConn::handleOptions(void *data)
     self->setQualityLevel(-1);
 
   self->updatePixelFormat();
+
+  self->refreshFramebuffer();
 }
 
 void CConn::handleUpdateTimeout(void *data)


ossman:
We've explicitly avoided doing a refresh when the colour mode changes. The problem was when bandwidth improved slightly we turned on full colour again. But the full refresh consumed so much bandwidth that we triggered a switch back to low colour mode. And it ended up toggling back and forth, becoming very unresponsive.

But that was connected to the automatic switching. Since this is for the manual switching, it should be more safe.

Did you consider doing this in the server instead? We could use the "lossy refresh" mechanism to fix up the screen. It has the benefit of doing things partially, and won't block any "real" updates whilst doing so.

It should just be a matter of getting `EncodeManager` to add everything to the `lossyRegion` tracking.

CendioOssman:
This piqued my interest, so I tried coming up with a patch:

```diff
diff --git a/common/rfb/EncodeManager.cxx b/common/rfb/EncodeManager.cxx
index 7ec70e69c..6a63fa6f6 100644
--- a/common/rfb/EncodeManager.cxx
+++ b/common/rfb/EncodeManager.cxx
@@ -278,6 +278,13 @@ void EncodeManager::pruneLosslessRefresh(const core::Region& limits)
   pendingRefreshRegion.assign_intersect(limits);
 }
 
+void EncodeManager::forceRefresh(const core::Region& req)
+{
+  lossyRegion.assign_union(req);
+  if (!recentChangeTimer.isStarted())
+    pendingRefreshRegion.assign_union(req);
+}
+
 void EncodeManager::writeUpdate(const UpdateInfo& ui, const PixelBuffer* pb,
                                 const RenderedCursor* renderedCursor)
 {
diff --git a/common/rfb/EncodeManager.h b/common/rfb/EncodeManager.h
index 959c13d63..4ce6d0ce7 100644
--- a/common/rfb/EncodeManager.h
+++ b/common/rfb/EncodeManager.h
@@ -54,6 +54,8 @@ namespace rfb {
 
     void pruneLosslessRefresh(const core::Region& limits);
 
+    void forceRefresh(const core::Region& req);
+
     void writeUpdate(const UpdateInfo& ui, const PixelBuffer* pb,
                      const RenderedCursor* renderedCursor);
 
diff --git a/common/rfb/VNCSConnectionST.cxx b/common/rfb/VNCSConnectionST.cxx
index 0b3537ad5..0038daff0 100644
--- a/common/rfb/VNCSConnectionST.cxx
+++ b/common/rfb/VNCSConnectionST.cxx
@@ -467,6 +467,7 @@ void VNCSConnectionST::setPixelFormat(const PixelFormat& pf)
   pf.print(buffer, 256);
   vlog.info("Client pixel format %s", buffer);
   setCursor();
+  encodeManager.forceRefresh(server->getPixelBuffer()->getRect());
 }
 
 void VNCSConnectionST::pointerEvent(const core::Point& pos,
```

phd:
Seems like fixing it on the server side is a better idea.

Will it also work with shared connections when only one client decreases the color level?

Side note: it may take a lot of time for servers out there to upgrade their TigerVNC (most likely together with the entire stable distro release). Many will stay on the old version for years. Meanwhile users may want to have this fixed with their latest version of the TigerVNC client.

CendioOssman:
> Will it also work with shared connections when only one client decreases the color level?
> 

Yes, this is in the per-client code.

> Side note: it may take a lot of time for servers out there to upgrade their TigerVNC (most likely together with the entire stable distro release). Many will stay on the old version for years. Meanwhile users may want to have this fixed with their latest version of the TigerVNC client.

What better way to incentivize distributions to upgrade than to add fun new features. :)

CendioOssman:
Committed as 6e0da494c9c862567dba9c4ff8aec93695f11cfb.

As a workaround until you are upgraded, the context menu has the "Refresh screen" entry.

PR #1931, opened 2025-03-20
Fix mouse click event leak on context menu close

phd:
When closing the F8 popup menu by clicking somewhere on the TigerVNC viewer's window a mouse click event is sent to the server.
No other toolkit nor application behaves like that when closing a popup menu - the closing click is ignored as an input.
And so it should be ignored here, especially that this input is being sent to a remote machine.

This commit fixes it by relying on FLTK's event loop which lets us know the lifetime of our popup menu and therefore easily filter out the closing mouse click event.

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -98,7 +98,7 @@ Viewport::Viewport(int w, int h, CConn* cc_)
     lastPointerPos(0, 0), lastButtonMask(0),
     keyboard(nullptr), shortcutBypass(false), shortcutActive(false),
     firstLEDState(true), pendingClientClipboard(false),
-    menuCtrlKey(false), menuAltKey(false), cursor(nullptr),
+    menuOpened(false), menuCtrlKey(false), menuAltKey(false), cursor(nullptr),
     cursorIsBlank(false)
 {
 #if defined(WIN32)
@@ -481,6 +481,20 @@ int Viewport::handle(int event)
     if (Fl::event_button3())
       buttonMask |= 1 << 2;
 
+    // Suppress handling of mouse events until the context popup menu is closed.
+    // This prevents a situation when closing the context menu by clicking
+    // somewhere on the TigerVNC viewer's window causes a mouse click event
+    // to be sent to the server.
+    if (menuOpened) {
+      if (buttonMask != 0) {
+        vlog.debug("Ignored mouse event %d,(%d,%d),%d while closing the menu.",
+                   event, Fl::event_x(), Fl::event_y(), buttonMask);
+        return 1;
+      }
+      vlog.debug("Allowed mouse event %d,(%d,%d),%d while closing the menu.",
+                 event, Fl::event_x(), Fl::event_y(), buttonMask);
+    }
+
   // The back/forward buttons are not supported by FTLK 1.3 and require
   // a patch which adds these buttons to the FLTK API. These buttons
   // will be part of the upcoming 1.4 API:
@@ -646,6 +660,14 @@ void Viewport::flushPendingClipboard()
   }
 
   pendingClientClipboard = false;
+}
+
+
+void Viewport::handleMenuClosed(void *data)
+{
+  Viewport *self = (Viewport *)data;
+  assert(self);
+  self->menuOpened = false;
 }
 
 
@@ -981,7 +1003,9 @@ void Viewport::popupContextMenu()
   // FLTK also doesn't switch focus properly for menus
   Fl::handle(FL_UNFOCUS, window());
 
+  menuOpened = true;
   m = contextMenu->popup();
+  Fl::add_timeout(0.0, handleMenuClosed, this);
 
   Fl::handle(FL_FOCUS, window());
 
--- a/vncviewer/Viewport.h
+++ b/vncviewer/Viewport.h
@@ -103,6 +103,8 @@ private:
   void initContextMenu();
   void popupContextMenu();
 
+  static void handleMenuClosed(void *data);
+
   static void handleOptions(void *data);
 
 private:
@@ -126,6 +128,7 @@ private:
   int clipboardSource;
 
   Fl_Menu_Button *contextMenu;
+  bool menuOpened;
 
   bool menuCtrlKey;
   bool menuAltKey;


vncviewer/Viewport.cxx
```diff
@@ -632,9 +632,19 @@ void Viewport::flushPendingClipboard()
...
+  assert(self);
+  self->menuOpened = false;
+}
+
+
 void Viewport::handlePointerEvent(const core::Point& pos,
                                   uint16_t buttonMask)
 {
+  if (menuOpened)
+    return;
```
CendioOssman, line 647: This is very non-obvious, so a comment would be nice, explaining that this is a workaround and for what

vncviewer/Viewport.cxx
```diff
@@ -805,7 +811,9 @@ void Viewport::popupContextMenu()
   // FLTK also doesn't switch focus properly for menus
   handle(FL_UNFOCUS);
 
+  menuOpened = true;
   m = contextMenu->popup();
+  Fl::add_timeout(0.0, handleMenuClosed, this);
```
CendioOssman, line 816: Why the timer? Isn't the mouse press handled by the time `popup()` returns?

CendioOssman (changes requested):
This sounds like a FLTK issue, so it should be ideally be fixed there. Have you reported it to them?

That said, a minimal workaround is acceptable if it causes practical practical problems for you.

phd:
> This sounds like a FLTK issue, so it should be ideally be fixed there. Have you reported it to them?

It seems like it's NOT a FLTK bug per se.
FLTK's [menubar.cxx](https://github.com/fltk/fltk/blob/master/test/menubar.cxx) test binary works fine. Its `Fl_Menu_Button` is showing a popup menu and clicking it away does not trigger any events in the underlying window.
Which is strange that TigerVNC's `Viewport::handle(int event)` is called. Investigating...

Tested with FLTK 1.3.11, which is the old legacy version, and it works fine even there.
My system's FLTK version is at 1.3.8, not sure if it makes a difference.

> That said, a minimal workaround is acceptable if it causes practical practical problems for you.

Not that I use this popup menu a lot. But when I do and decide to click it away it causes unwanted input to be sent to the server potentially messing with the running application.

phd:
> Which is strange that TigerVNC's `Viewport::handle(int event)` is called. Investigating...

I have modified the above `menubar.cxx` test with my own `Fl_Box`:

```c++
class My_Box : public Fl_Box {
public:
  using Fl_Box::Fl_Box;
  int handle(int event) {
    printf("My_Box::event(%d)\n", event);
    return Fl_Box::handle(event);
  }
};
```
```c++
  My_Box b(200,200,200,100,"Press right button\nfor a pop-up menu");
```

Fl_Widget::handle(int) is indeed called a lot of times when the widget happens to be under the mouse when clicking away a popup menu:

```
My_Box::event(3)    // FL_ENTER
My_Box::event(11)   // FL_MOVE
My_Box::event(11)   // FL_MOVE
My_Box::event(11)   // FL_MOVE
My_Box::event(0)    // FL_NO_EVENT
My_Box::event(2)    // FL_RELEASE
My_Box::event(11)   // FL_MOVE
```

And in TigerVNC's viewer it's the FL_RELEASE event that is being sent to the client.

FLTK must handle this situation internally for its widgets not to act on these events.

So we would either need to filter it out somehow which might be tricky (FL_RELEASE with no earlier FL_PUSH after FL_ENTER?), or simply use this workaround.

vncviewer/Viewport.cxx
phd: No. But when the `contextMenu->popup();` returns, the mouse events that caused the popup to close are already queued (and they fire `Viewport::handle(int)`). So we queue this instant timer event after them and ignore the mouse events until the timeout event happens.
All subsequent mouse events will be queued after our timer, so they will be handled normally.

CendioOssman:
Ah, I see what's going on. Unfortunately, it's pretty fundamental to how VNC and vncviewer work. So it will be difficult to work around in a nice way.

The problem is that VNC isn't concerned with FL_MOVE vs FL_PUSH vs FL_RELEASE. It only cares about the current state of the mouse. So we never look at what type of event we got, only what buttons are pressed.

The menu closes when a mouse button is pressed. This is consistent with other toolkits. But after that press, we'll start getting move events. Again, I'd say this is consistent with other toolkits. The issue is that for those move events, we'll correctly deduce that a mouse button is pressed. So we'll send that to the server. Which will result in the applications seeing a mouse press.

phd:
Subsequent move events (and ButtonPress events they may cause) are not an issue here.
The orphaned Release Event is. It happens even when the user does not move the mouse, just clicks away the menu with a quick ButtonPress and ButtonRelease.

Clicking away any popup menu (Qt, GTK, FLTK) with mouse pointer over a **locally** running `xev` window shows this:

```
EnterNotify event, serial 40, synthetic NO, window 0x5200001,
    root 0x637, subw 0x0, time 691586, (141,73), root:(141,630),
    mode NotifyUngrab, detail NotifyNonlinear, same_screen YES,
    focus NO, state 256

KeymapNotify event, serial 40, synthetic NO, window 0x0,
    keys:  1   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   
           0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   0   

ButtonRelease event, serial 40, synthetic NO, window 0x5200001,
    root 0x637, subw 0x0, time 691597, (141,73), root:(141,630),
    state 0x100, button 1, same_screen YES
```

This never generates click events in the underlying windows, so they do not perform any unwanted actions.

But when you click away the TigerVNC's context menu over a `xev` window running on the **remote** server, it outputs:

```
MotionNotify event, serial 38, synthetic NO, window 0x1800001,
    root 0x47c, subw 0x0, time 893923, (36,125), root:(870,155),
    state 0x0, is_hint 0, same_screen YES

ButtonPress event, serial 38, synthetic NO, window 0x1800001,
    root 0x47c, subw 0x0, time 893923, (36,125), root:(870,155),
    state 0x0, button 1, same_screen YES

ButtonRelease event, serial 38, synthetic NO, window 0x1800001,
    root 0x47c, subw 0x0, time 893976, (36,125), root:(870,155),
    state 0x100, button 1, same_screen YES
```

Indeed we have a problem with ButtonPress Event generated by the VNC.

Clicking away TigerVNC's context menu causes remote applications to perform actions related to this unwanted click event.

This patch suppresses the orphaned Release Event, which solves this problem.
Now remotely running `xev` shows no events at all for a simple mouse click, which is what we want.

And if you click away the context menu by pressing-and-holding the left mouse button over the remote `xev` and then drag the mouse, it will not suppress the subsequent move events and the remotely running `xev` will (correctly) show:

```
[...]

ButtonPress event, serial 38, synthetic NO, window 0x1800001,
    root 0x47c, subw 0x0, time 1288942, (73,109), root:(907,139),
    state 0x0, button 1, same_screen YES

MotionNotify event, serial 38, synthetic NO, window 0x1800001,
    root 0x47c, subw 0x0, time 1288963, (74,106), root:(908,136),
    state 0x100, is_hint 0, same_screen YES

MotionNotify event, serial 38, synthetic NO, window 0x1800001,
    root 0x47c, subw 0x0, time 1288981, (78,103), root:(912,133),
    state 0x100, is_hint 0, same_screen YES

MotionNotify event, serial 38, synthetic NO, window 0x1800001,
    root 0x47c, subw 0x0, time 1288998, (84,100), root:(918,130),
    state 0x100, is_hint 0, same_screen YES

ButtonRelease event, serial 38, synthetic NO, window 0x1800001,
    root 0x47c, subw 0x0, time 1289859, (116,81), root:(950,111),
    state 0x100, button 1, same_screen YES
```

phd:
Updated:
- moved the check to `Viewport::handle(int event)`
- added vlog.debug() for the suppressed events

You are right, they are all indeed move events (reported from FLTK), 11 == FL_MOVE:

```
Fri Mar 21 23:00:00 2025
 Viewport:    Ignored pointer event 11 because of opened menu.
 Viewport:    Ignored pointer event 11 because of opened menu.
 Viewport:    Ignored pointer event 11 because of opened menu.
 Viewport:    Ignored pointer event 11 because of opened menu.
```

But the mouse pointer didn't move during the click!

Are these move events which occurred when the mouse pointer left the context menu and was moving over the main window, before the click?

vncviewer/Viewport.cxx
```diff
@@ -464,6 +464,18 @@ int Viewport::handle(int event)
...
+    // This prevents a situation when closing the context menu by clicking
+    // somewhere on the TigerVNC viewer's window causes a mouse click event
+    // to be sent to the server.
+    if (menuOpened) {
+      if (buttonMask != 0) {
+        vlog.debug("Ignored mouse event %d,%d while closing menu.", event, buttonMask);
+        return 1;
+      }
+      vlog.debug("Allowed mouse event %d,%d while closing menu.", event, buttonMask);
+    }
```
phd, line 477: Updated to allow (move) events that report no buttons pressed in `buttonMask`.
This properly updates the mouse position when closing the context menu with an ESC key.

It seems there is some kind of race condition in FLTK when a popup menu is being closed, because these move events have buttons pressed randomly reported for them.
Or maybe reading Fl::event_button1/2/3() for such events queued while closing a popup menu does not work reliably?

```
Fri Mar 21 23:17:39 2025
 Viewport:    Ignored mouse event 11,1 while closing menu.
 Viewport:    Ignored mouse event 11,1 while closing menu.
 Viewport:    Ignored mouse event 11,1 while closing menu.
 Viewport:    Ignored mouse event 11,1 while closing menu.

Fri Mar 21 23:17:40 2025
 Viewport:    Allowed mouse event 11,0 while closing menu.
 Viewport:    Allowed mouse event 11,0 while closing menu.
 Viewport:    Allowed mouse event 11,0 while closing menu.
 Viewport:    Allowed mouse event 11,0 while closing menu.
```
11,1 = `event`,`buttonMask`

phd, line 477: Updated to allow (move) events that report no buttons pressed in `buttonMask`.
This properly updates the mouse position when closing the context menu with an ESC key.

It seems there is some kind of race condition in FLTK when a popup menu is being closed, because these move events have buttons pressed randomly reported for them.
Or maybe reading Fl::event_button1/2/3() for such events queued while closing a popup menu does not work reliably?

```
Fri Mar 21 23:17:39 2025
 Viewport:    Ignored mouse event 11,1 while closing menu.
 Viewport:    Ignored mouse event 11,1 while closing menu.
 Viewport:    Ignored mouse event 11,1 while closing menu.
 Viewport:    Ignored mouse event 11,1 while closing menu.

Fri Mar 21 23:17:40 2025
 Viewport:    Allowed mouse event 11,0 while closing menu.
 Viewport:    Allowed mouse event 11,0 while closing menu.
 Viewport:    Allowed mouse event 11,0 while closing menu.
 Viewport:    Allowed mouse event 11,0 while closing menu.
```
11,1 = `event`,`buttonMask`

phd: I have added one more temporary `debug()` to log every mouse event occurring in those situations:
```c++
if (menuOpened) {
  if (buttonMask != 0) {
    vlog.debug("Ignored mouse event %d,(%d,%d),%d while closing the menu.", event, Fl::event_x(), Fl::event_y(), buttonMask);
    return 1;
  }
  vlog.debug("Allowed mouse event %d,(%d,%d),%d while closing the menu.", event, Fl::event_x(), Fl::event_y(), buttonMask);
} else {
  vlog.debug("Allowed mouse event %d,(%d,%d),%d.", event, Fl::event_x(), Fl::event_y(), buttonMask);
}
```

Clicking away the menu, **button pressed randomly reported**:

```
Fri Mar 21 23:57:33 2025
 Viewport:    Ignored mouse event 11,(980,525),1 while closing the menu.
 Viewport:    Ignored mouse event 11,(980,525),1 while closing the menu.
 Viewport:    Ignored mouse event 11,(980,525),1 while closing the menu.
 Viewport:    Ignored mouse event 11,(980,525),1 while closing the menu.
 Viewport:    Allowed mouse event 2,(980,525),0.
 Viewport:    Allowed mouse event 11,(980,525),0.
```

Clicking away the menu, **no buttons randomly reported**:

```
Fri Mar 21 23:57:34 2025
 Viewport:    Allowed mouse event 11,(941,556),0 while closing the menu.
 Viewport:    Allowed mouse event 11,(941,556),0 while closing the menu.
 Viewport:    Allowed mouse event 11,(941,556),0 while closing the menu.
 Viewport:    Allowed mouse event 11,(941,556),0 while closing the menu.
```

Clicking away the menu with **press & hold & drag**:

```
Fri Mar 21 23:58:11 2025
 Viewport:    Ignored mouse event 11,(952,473),1 while closing the menu.
 Viewport:    Ignored mouse event 11,(952,473),1 while closing the menu.
 Viewport:    Ignored mouse event 11,(952,473),1 while closing the menu.
 Viewport:    Ignored mouse event 11,(952,473),1 while closing the menu.
 Viewport:    Allowed mouse event 11,(953,473),1.
 Viewport:    Allowed mouse event 11,(954,473),1.
 Viewport:    Allowed mouse event 11,(955,473),1.
[...]
 Viewport:    Allowed mouse event 11,(980,486),1.
 Viewport:    Allowed mouse event 2,(980,486),0.
 Viewport:    Allowed mouse event 11,(980,486),0.
```

Closing the menu with an **ESC key**:

```
Fri Mar 21 23:59:38 2025
 Viewport:    Allowed mouse event 11,(1213,468),0 while closing the menu.
 Viewport:    Allowed mouse event 11,(1213,468),0 while closing the menu.
 Viewport:    Allowed mouse event 11,(1213,468),0 while closing the menu.
 Viewport:    Allowed mouse event 11,(1213,468),0 while closing the menu.
```

These 4 phantom move events at constant pointer position are consistent.

And it seems the patch properly filters out the random reports for the button still being pressed when the menu is being closed.

phd:
@CendioOssman 
What the `Viewport::handle(int event)` function sees are actually Move Events with (or without) mouse button pressed.
So my comment you are quoting to was a bit wrong about the Release Events.

Please check my later comments:
https://github.com/TigerVNC/tigervnc/pull/1931#issuecomment-2744567954
https://github.com/TigerVNC/tigervnc/pull/1931#pullrequestreview-2707464633
https://github.com/TigerVNC/tigervnc/pull/1931#discussion_r2008467165

The problem is with 4 phantom Move Events (FL_MOVE == 11) which should probably never occur in FLTK in the first place.
They are consistent and (selectively) skipping them until the next event loop tick seems like a very reliable way of fixing this.
Every real mouse button press or release by the user will be delivered in one of subsequent event loop passes.

CendioOssman:
I think this will give a rather unpredictable user experience. If the mouse is sensitive enough, or if the user is not still enough, they'll still get clicks, but now more randomly.

This needs to be fixed more fully for it to be merged.

PR #1932, opened 2025-03-20
Fixed build warnings: ignoring return value

phd:
Fixed build warnings: ignoring return value

No more warnings during builds (unless introduced by new changes, which is useful to notice).

### DIFF
--- a/unix/vncserver/vncsession.c
+++ b/unix/vncserver/vncsession.c
@@ -134,7 +134,7 @@ begin_daemon(void)
 static void
 finish_daemon(void)
 {
-    write(daemon_pipe_fd, "+", 1);
+    if (write(daemon_pipe_fd, "+", 1) == -1) {}
     close(daemon_pipe_fd);
     daemon_pipe_fd = -1;
 }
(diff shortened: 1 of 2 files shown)


unix/vncserver/vncsession.c
```diff
@@ -134,7 +134,8 @@ begin_daemon(void)
 static void
 finish_daemon(void)
 {
-    write(daemon_pipe_fd, "+", 1);
+    if (write(daemon_pipe_fd, "+", 1) == -1)
+        fprintf(stderr, "Failed to write() to daemon_pipe_fd.\n");
```
CendioOssman, line 138: This is not going to work. stderr is set to /dev/null at this point.

Handling this error is probably not crucial. So might just want to silence the compiler warning.

unix/vncserver/vncsession.c
```diff
@@ -546,7 +547,8 @@ run_script(const char *username, const char *display, char **envp)
     switch_user(pwent->pw_name, pwent->pw_uid, pwent->pw_gid);
 
     if (chdir(pwent->pw_dir) == -1)
-        chdir("/");
+        if (chdir("/") == -1)
+            syslog(LOG_CRIT, "chdir: %s", strerror(errno));
```
CendioOssman, line 551: The second `chdir()` is an emergency fallback. So we should probably log the issue with the first one in that case.

And a more descriptive error message so the user/sysadmin knows what failed.

CendioOssman (changes requested):
Thanks!

I'm not getting those warnings here, though. Have you forced some different compiler flags?

phd:
> Have you forced some different compiler flags?

No, I haven't changed compiler flags. GCC 13.3.0 from Ubuntu 24.04.2 LTS.

unix/vncserver/vncsession.c
phd: Done.
Seems like `if (...) {}` is the best way to silence this, as casting the result to `(void)...` no longer works.

CendioOssman:
I found the reason we see different things. Ubuntu apparently sets `_FORTIFY_SOURCE` by default. I'll have a look at getting that in to our builds more consistently so we all get the same warnings.

CendioOssman (approved):

PR #1950, opened 2025-05-16
Use notepad to open text files

adetabs:
Change ReadMe and License files to open with Notepad instead of WordPad, since [WordPad (`write.exe`) has been removed in new versions of Windows.](https://learn.microsoft.com/en-gb/windows/whats-new/deprecated-features-resources#wordpad)

Currently the Start Menu icons Read Me and License lead to this on Windows 11 24H2 build 26100.4061:

image


CendioOssman:
Thanks!

Do we need to specify a program at all? What happens if we just specify the documents themselves in `FileName`?

jobatabs:
I don't know enough about InnoSetup to say. I would wager that a link to a file would not be a valid Start Menu shortcut, but could be wrong though.

CendioOssman:
I guess you have a working setup where you can build the installer? Could you give it a try and see what happens?

adetabs:
I actually don’t, I made this patch on the web editor…

PR #1957, opened 2025-06-01
Make Xvnc pass the display name to PAM modules.

casantos:
Fixes: https://issues.redhat.com/browse/RHEL-34880
Supersedes: https://github.com/TigerVNC/tigervnc/pull/1792


CendioOssman (changes requested):
Giving PAM more details looks fine. But I'm not a fan of some secret global variable. We should have something that fits in better with the object structure.

Unfortunately, the actual authentication is pretty abstracted away at the moment. So it's not obvious what structure we need.

common/rfb/pam.c
```diff
@@ -24,10 +24,13 @@
 
 #include <stdlib.h>
 #include <string.h>
+#include <syslog.h>
 #include <security/pam_appl.h>
 
 #include <rfb/pam.h>
 
+char displayStr[32]; /* set by Xvnc and x0vncserver */
```
CendioOssman: Given how we handle similar things, I'd say the most appropriate place is for this to be a static member of a class. `UnixPasswordValidator` is the wrapper for this functionality, so let's put it there.

common/rfb/pam.c
```diff
@@ -84,8 +87,22 @@ int do_pam_auth(const char *service, const char *username, const char *password)
...
   ret = pam_start(service, username, &conv, &h);
-  if (ret == PAM_SUCCESS)
+  if (ret == PAM_SUCCESS) {
+
+#ifdef PAM_XDISPLAY
+    /* Pass the display name to PAM modules. */
+    ret = pam_set_item(h, PAM_XDISPLAY, displayStr);
+    /* Note: PAM_XDISPLAY is only supported by modern versions of PAM */
+    if (ret== PAM_SUCCESS || ret == PAM_BAD_ITEM) {
+        ret = pam_authenticate(h, 0);
```
CendioOssman: This `if`-structure will let users pass on without authentication if something unexpected happens with `pam_set_item()`.

Let's restructure this so that we don't duplicate the call to `pam_authenticate()`.

common/rfb/pam.c
```diff
@@ -84,8 +87,22 @@ int do_pam_auth(const char *service, const char *username, const char *password)
   };
   pam_handle_t *h = 0;
   ret = pam_start(service, username, &conv, &h);
-  if (ret == PAM_SUCCESS)
+  if (ret == PAM_SUCCESS) {
+
+#ifdef PAM_XDISPLAY
+    /* Pass the display name to PAM modules. */
+    ret = pam_set_item(h, PAM_XDISPLAY, displayStr);
+    /* Note: PAM_XDISPLAY is only supported by modern versions of PAM */
```
CendioOssman: It's almost 20 years old. Do such systems still exist, and will run binaries built on a more modern system?

common/rfb/pam.c
```diff
@@ -84,8 +87,22 @@ int do_pam_auth(const char *service, const char *username, const char *password)
...
+    if (ret== PAM_SUCCESS || ret == PAM_BAD_ITEM) {
+        ret = pam_authenticate(h, 0);
+    } else {
+        syslog(LOG_CRIT, "pam_set_item(PAM_XDISPLAY) failed: %d (%s)",
+               ret, pam_strerror(h, ret));
```
CendioOssman: We don't generally use syslog, so let's avoid making this some odd exception. Please use the normal logging.

Which might mean we need to merge `pam.c` in to `UnixPasswordValidator`. Not sure if there was something preventing that, and that's why we have separate files?

unix/x0vncserver/x0vncserver.cxx
```diff
@@ -334,12 +336,15 @@ int main(int argc, char** argv)
...
+  const char *dname = XDisplayName(displayname);
   if (!(dpy = XOpenDisplay(displayname))) {
     // FIXME: Why not vlog.error(...)?
     fprintf(stderr,"%s: Unable to open display \"%s\"\r\n",
-            programName, XDisplayName(displayname));
+            programName, dname);
     exit(1);
   }
+  strncpy(displayStr, dname, 31);
+  displayStr[31] = '\0';
```
CendioOssman: We've generally switched to using `std::string` to avoid risky code like this. Hopefully we can do the same here.

unix/xserver/hw/vnc/xvnc.c
```diff
@@ -111,7 +111,8 @@ static Bool vncPixmapDepths[33];
 static Bool Render = TRUE;
 
 static Bool displaySpecified = FALSE;
-static char displayNumStr[16];
+
+extern char displayStr[32];  /* This is where we pass the display to PAM modules */
```
CendioOssman: I think it's more modular to let them keep their own two copies of things.

unix/xserver/hw/vnc/xvnc.c
```diff
@@ -277,8 +278,11 @@ ddxProcessArgument(int argc, char *argv[], int i)
         vncInitRFB();
     }
 
-    if (argv[i][0] == ':')
+    if (argv[i][0] == ':') {
         displaySpecified = TRUE;
+        strncpy(displayStr, argv[i], 31);
+        displayStr[31] = '\0';
+    }
```
CendioOssman: Looks like `display` is the canonical truth here. Probably best to look at that once everything is set up.

CendioOssman (changes requested):

common/rfb/pam.c
casantos: Not sure if we can pass a static member of a C++ class to the C code in libpam without resorting to come trick, because we need to pass `pam_callback()` to `pam_start()` in a `pam_conv`struct, but I will investigate this approach.

common/rfb/pam.c
casantos: ACK

common/rfb/pam.c
casantos: It's tested in [vncsession.c](https://github.com/TigerVNC/tigervnc/blob/5bfcb7a0c7559c3f029da68d89619e8944021fa8/unix/vncserver/vncsession.c#L215). I just followed the pattern. Anyway, the test is harmless.

common/rfb/pam.c
casantos: Is probably because `libpam`, which is pure C code needs to have access to the functions. I will investigate this possibility.

unix/x0vncserver/x0vncserver.cxx
casantos: You can't pass an `std::string` to a PAM module unless it's written in C++, which breaks the ABI. PAM modules must be loadable by C-only programs.

unix/xserver/hw/vnc/xvnc.c
casantos: ACK

unix/xserver/hw/vnc/xvnc.c
casantos: We can't reference `display` elsewhere (e.g. in `pam.c` or `x0vncserver`). It's a pointer defined in the X server code, which linked only to Xvnc.

common/rfb/UnixPasswordValidator.cxx
```diff
@@ -21,25 +21,101 @@
...
+
 static core::StringParameter pamService
   ("PAMService", "Service name for PAM password validation", "vnc");
 core::AliasParameter pam_service("pam_service", "Alias for PAMService",
                                  &pamService);
 
-int do_pam_auth(const char *service, const char *username,
-	        const char *password);
+extern "C" {
+
```
CendioOssman: I don't think this is needed?

unix/x0vncserver/x0vncserver.cxx
CendioOssman: `pam_set_item()` copies the given data, so `std::string::c_str()` will be able to do what you want.

unix/xserver/hw/vnc/xvnc.c
CendioOssman: Indeed. But we should be able to copy its value.

My point was that there might be more things influencing how `display` is set. But it should hopefully be stable once the X server is started.

So it should be more robust looking at the final value of `display`, rather than trying to grab it mid flight.

If nothing else, the fact that we need to intercept things in two different places suggests that there's a better way to do things.

common/rfb/UnixPasswordValidator.h
```diff
@@ -21,8 +21,13 @@
 #ifndef __PASSWORD_VALIDATOR_H__
 #define __PASSWORD_VALIDATOR_H__
 
+#ifdef __cplusplus
+
 #include <rfb/SSecurityPlain.h>
 
+extern "C" void setDisplayByName(const char *display);
+extern "C" void setDisplayByNumber(int displayNum);
+
```
CendioOssman: No, this isn't even namespaced. We don't want a mix of C code in here with the C++ interfaces.

I understand you want this so you can call it from `xvnc.c`. But that creates a bit of a mess. Please add a wrapper in `RFBGlue.cc` like we have for everything else where C code needs to reach the C++ stuff.

common/rfb/UnixPasswordValidator.cxx
casantos: pam_callback must have C linkage to be called by the C code in libpam.
By the way, is that `#if defined(__sun)` test still required? Is anybody still building TigerVNC on Solaris?

CendioOssman: > pam_callback must have C linkage to be called by the C code in libpam.

You're passing a pointer to the function, so the name mangling/resolution is not an issue. The compiler would complain if the pointers were incompatible in some way.

> By the way, is that `#if defined(__sun)` test still required? Is anybody still building TigerVNC on Solaris?

I think so. @alanc usually mentions it whenever I happen to break something. :)

No idea if the `#if` is still relevant, though.

alanc: Yes, we build and package TigerVNC as the primary VNC implementation in Solaris 11.4: https://github.com/oracle/solaris-userland/tree/master/components/x11/xserver/xvnc

Unfortunately, yes, the #ifdef is still required because the Solaris <security/pam_appl.h> header does not include `const` on that argument in the conversation prototype.

common/rfb/pam.c
casantos: I confirmed that PAM_XDISPLAY does not exist in Solaris 11.4, so the test is still necessary.

Issue #1958, opened 2025-06-02
x0vncserver aborts on startup


casantos:
**Describe the bug**
x0vncserver crashes on startup.

**To Reproduce**
Run x0vncserver.

**Expected behavior**
Should start normally.

**Client (please complete the following information):**
 - Not relevant

**Server (please complete the following information):**
 - OS: RHEL 9.6
 - VNC server: TigerVNC
 - VNC server version:  developer head (commit 049e15d3)
 - Server downloaded from: built from source
 - Server was started using: `x0vncserver`

**Additional context**

Crash transcription:

$ x0vncserver

Sun Jun  1 15:41:42 2025
 Geometry:    Desktop geometry is set to 1024x768+0+0
 XDesktop:    Using evdev codemap
 XDesktop:    
 XDesktop:    XTest extension present - version 2.2
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = std::__cxx11::basic_string<char>; _Alloc = std::allocator<std::__cxx11::basic_string<char> >; std::vector<_Tp, _Alloc>::reference = std::__cxx11::basic_string<char>&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Aborted (core dumped)

Debug transcription:

Program terminated with signal SIGABRT, Aborted.
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
44	      return INTERNAL_SYSCALL_ERROR_P (ret) ? INTERNAL_SYSCALL_ERRNO (ret) : 0;
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007f742ce8be93 in __pthread_kill_internal (signo=6, threadid=<optimized out>) at pthread_kill.c:78
#2  0x00007f742ce3eb46 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007f742ce28833 in __GI_abort () at abort.c:79
[...]

PR #1959, opened 2025-06-02
Fix crash on x0vncserver startup

casantos:
Test if the "parts" vector is not empty in TcpFilter::parsePattern() before attempting to access its first element.

Fixes: https://github.com/TigerVNC/tigervnc/issues/1958


CendioOssman:
Thanks! This is fallout from fb55b64a29cc81475c13f9dca484638dcf92ebb8.

PR #1961, opened 2025-06-08
Silence several GCC 14 warnings

casantos:
xvnc.c: remove unnecessary test of function return value
core::Configuration: add a size() method
RFBGlue.cc: use size() obtain the number of configuration parameters

### DIFF
--- a/common/core/Configuration.h
+++ b/common/core/Configuration.h
@@ -86,6 +86,8 @@ namespace core {
     std::list<VoidParameter*>::iterator begin() { return params.begin(); }
     std::list<VoidParameter*>::iterator end() { return params.end(); }
 
+    // - Returns the number of parameters
+    int size() { return params.size(); }
 
     // - Get the Global Configuration object
     //   NB: This call does NOT lock the Configuration system.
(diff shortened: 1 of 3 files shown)


unix/xserver/hw/vnc/vncSelection.c
```diff
@@ -69,7 +69,7 @@ static int vncOwnSelection(Atom selection);
 static int vncConvertSelection(ClientPtr client, Atom selection,
                                Atom target, Atom property,
                                Window requestor, CARD32 time,
-                               const char* data);
+                               char* data);
```
CendioOssman: Are you using an old xorg-server? `dixChangeWindowProperty()` takes a `const void *` argument since xorg-server 21:

https://gitlab.freedesktop.org/xorg/xserver/-/commit/d83c84bd9dc7fbd3588ce70b0c8bf0f63ed29ff3

CendioOssman (changes requested):
Changes look except for the selection code one. Please have a look.

unix/xserver/hw/vnc/vncSelection.c
casantos: RHEL 9 still has xorg-server 1.20.11. Let's revert this change and coexist with the warning.

PR #1969, opened 2025-06-20
Add BUILD_SERVER similar to BUILD_VIEWER

ceamac:
Add a BUILD_SERVER option similar to BUILD_VIEWER.  When BUILD_SERVER is not active, we do not require pam, so also disable everything that needs pam.  This allows building the viewer on a machine where pam is not installed.


CendioOssman (changes requested):
PAM is a very common library, so I'm doubtful this is worthwhile.

What is your system setup where installing PAM is an issue?

ceamac:
The main part was not building `vncserver` and related tools when only the client is needed.  PAM is just bonus, because the client does not use it, so it gets pulled as a dependency for no reason if not also building the server.

We carry this patch in Gentoo for 3 years now, so I figured I should try to send it upstream.  But I can drop this PR, or I can simplify it by removing the PAM related changes.

CendioOssman:
I see. I don't think that's worth the complexity either. I understand not building more than what you need. But the build time for those things is very short and isn't worth the added risk and maintenance of more options.

PR #1973, opened 2025-06-29
Add translations to everything

CendioOssman:
Currently, only the native viewer is translated. However, a lot of text the user sees also comes from the common library code.

There are also other components in the project that should be translated for users.

This PR moves the translation infrastructure to the common libraries, and tags everything for translation.


CendioOssman:
One issue here is that the translation files will be shared between the viewer and the servers. That's a problem because both the RPMs and debs have things split up in separate packages.

@JoachimFalk and @grulja, do you have any suggestions or preferences here? Any other packages with the same structure that we can mimic?

grulja:
Hi, I don't think that would be a problem from the packaging point of view. There are two options for Fedora/RHEL:
1) Currently there is `tigervnc` package with `vncviewer` only. We can make it a separate package, like `tigervnc-viewer` or similar and combine it with currently separate `tigervnc-icons` and `tigervnc-license`. This package would be then required by both `tigervnc-server` and `tigervnc-viewer`.
2) Introduce a new package, something like `tigervnc-common` with the translations, icons and the license and also do the same requires for `tigervnc` (where we have the viewer) and `tigervnc-server`.

I don't know any other package that would have to deal with similar problem, but I can imagine it's done in similar fashion as I wrote above.

JoachimFalk:
At the moment, translation are only contained in tigervnc-viewer. But, in Debian, we have tigervnc-common, which tigervnc-tools (vncpasswd, etc.), tigervnc-scraping-server, tigervnc-standalone-server, tigervnc-xorg-extension, and tigervnc-viewer all depend upon.

The common translation would probably go into tigervnc-common if you do not intend to also provide a common library package  for which these translations are and which you also considered to be user consumable, i.e., has public headers and you would consider to be usefull to others. At the moment, the common library code is not made available as a package or even installed as a static lib in the client or server packages.

PR #1974, opened 2025-07-12
Clean up CMakeList.txt platform checks, make PAM dependency configurable

ccxcz:
Moved dependency on PAM library behind a proper build configuration option instead of the confusing `if(UNIX AND NOT APPLE)` being a proxy for what might be a PAM-enabled platform.
As added benefit it makes it more clear what the child directory `CMakeList.txt` conditionals and the `#ifdef` statements in the code actually guard for.

Additionally changed two more instances of `if(UNIX AND NOT APPLE)` that make no sense since given libraries are very clearly Linux-only.


ccxcz:
I've made these edits to compile `Xvnc` on a rather minimal musl-based system (used via unix domain sockets, no extra authentication needed).

I figured I'll publish the patch if anyone else has similar needs even if it doesn't make it into mainline. Take what you find useful out of it if anything.

That said I think this actually improves the situation with cmake conditionals slightly because:
1) The checks for non-PAM platforms are there anyway (unless macOS and Windows are suddenly unsupported).
2) Conditionals on features generally map on build units that are there already, so the only thing they do is to make you write out build/install targets for each feature as a contiguous block. This is unlike conditionals on platforms which indeed tend to devolve in hard to maintain mess especially on C/C++ source (ifdef) level.

If there are any changes to the patchset that'd make you happier with it let me know.

CendioOssman:
Sorry for letting this linger. I've reworked your commit a bit and merged it as part of #2114.

PR #1984, opened 2025-08-31
Searchable server history suggestions

samhed:
This adds a new widget for displaying suggestions in a **_searchable_** way, built into an input widget. It is used in ServerDialog to display the history of VNC servers instead of the Fl_Input_Choice widget.

The server name input field will show a drop-down list below if there are any historical VNC servers. At most 4 suggestions are shown at the same time. The following three actions trigger the suggestions to show:

- Typing in the input field
- Clicking the input field
- Pressing the down arrow key while focused

The suggestion drop-down can be closed by a click outside or by pressing ESC (utilizes event_dispatch_handler).

The suggestions match the user input and can be selected by a click or by pressing the enter key. Suggestions can be removed by clicking the x button. Suggestions can be cycled by pressing the up and down arrow keys. Suggestion matches are emphasized with a bold font, and separate words can be used to search for partial matches. Matching is case-insensitive because DNS names are case-insensitive.

The list can display both matches and historical suggestions at the same time. The list is sorted and shows the most relevant matches first, with historical suggestions below.


samhed:
Possibilities for future improvements on this includes:
- fuzzy searching
- scrolling in the list
- keyboard focus handling for the list items

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -0,0 +1,711 @@
...
+#include "Fl_Suggestion_Input.h"
+#include "event_dispatch_handler.h"
+#include "layout.h"
+
+/* The item needs to be offset inside the dropdown window
+   to not clip the border */
+#define SUGGESTION_OFFSET  2
+#define SUGGESTION_ITEM_HEIGHT  25
+#define REMOVE_ITEM_SIZE 17
+#define MAX_SHOWN_ITEMS  4
```
CendioOssman: This seems a bit low?

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -0,0 +1,711 @@
...
+    input->when(FL_WHEN_CHANGED);
+
+    max_width = w - 2 * SUGGESTION_OFFSET;
+    y += h;
+
+    end(); // Close Fl_Group
+
+    // The dropdown window is not a child of Fl_Suggestion_Input, but will
+    // instead be a sibling. Before being shown it is added to the parent.
+    dropdown_window = new Fl_Window(x, y, w, 0);
```
CendioOssman: This does not seem to work correctly. I'm not getting a new window, but rather just a widget container inside the existing window.

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Increased to 6 in b58233b02182d8cc90519140478afb936ec9eb42

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Switched to a separate window rather than a subwindow in bd4f33a945e4aa20fa0662d1d7db66ea2ce4cf71

samhed:
I'm satisfied with this now. I used separate commits in case someone wants to review; should squash most of them before merge.

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -0,0 +1,826 @@
...
+
+/* The item needs to be offset inside the dropdown window
+   to not clip the border. On macOS the border is outside. */
+#if defined(__APPLE__)
+#define SUGGESTION_OFFSET  0
+#else
+#define SUGGESTION_OFFSET  1
+#endif
+#define SUGGESTION_ITEM_HEIGHT  25
+#define REMOVE_ITEM_SIZE  17
```
CendioOssman: Should these be relative the font size? Perhaps margins are the more relevant metric?

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -0,0 +1,826 @@
...
+#define MAX_SHOWN_ITEMS  6
+
+/* The input needs to be offset inside the group
+   to not clip the rounded corners */
+#define INPUT_OFFSET  2
```
CendioOssman: Shouldn't it get this from the box model in that case?

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -0,0 +1,826 @@
...
+
+/*
+ * ============== Utility functions ================
+ */
+
+std::list<std::string> split_on_space(const std::string s)
```
CendioOssman: Local functions need to be `static`.

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -0,0 +1,826 @@
...
+
+Fl_Suggestion_Item::Fl_Suggestion_Item(
+    int x, int y, int w, int h, const std::string label,
+    Fl_Normalize_Callback *normalize_cb, Fl_Suggestion_Input *_parent
+)
+: Fl_Group(x, y, w, h), on_normalize_cb(normalize_cb), parent(_parent),
+    full_label(label), selected(false)
+{
+    base_color = FL_BACKGROUND2_COLOR;
+    base_hover_color = fl_color_average(FL_BLACK, base_color, 0.05);
```
CendioOssman: Does this match hover colors in other places of FLTK?

vncviewer/fltk/Fl_Suggestion_Input.h
```diff
@@ -0,0 +1,163 @@
...
+#include <FL/Fl_Menu_Window.H>
+
+class Fl_Box;
+class Fl_Input;
+class Fl_Suggestion_Input;
+
+typedef std::string (Fl_Normalize_Callback)(const std::string s);
+typedef void (Fl_Suggestion_Callback)(Fl_Widget*, std::string s, void*);
+
+class Fl_Suggestion_Item: public Fl_Group {
```
CendioOssman: This and the window are internal helpers. Have you considered moving them to separate file(s) to make the main interface more clear?

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -0,0 +1,826 @@
...
+}
+
+Fl_Suggestion_Input::~Fl_Suggestion_Input()
+{
+    Fl::remove_idle(show_suggestions_soon, this);
+    Fl::remove_idle(hide_suggestions_if_unfocused, this);
+    Fl::remove_system_handler(handle_system_event);
+    fl_remove_event_dispatch(dispatch_helper, this);
+}
+void Fl_Suggestion_Input::draw()
```
CendioOssman, line 106: Missing blank line

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -324,6 +325,21 @@ Fl_Suggestion_Input::~Fl_Suggestion_Input()
     Fl::remove_idle(show_suggestions_soon, this);
     fl_remove_event_dispatch(dispatch_helper, this);
 }
+void Fl_Suggestion_Input::draw()
+{
+    // Draw the box taller to give the appearance of the rounded corners
+    // disappearing when the dropdown window is shown. Without the rounded
+    // bottom corners the dropdown visually sticks to the input in a nice way.
+    if (suggestions_visible())
+        draw_box(box(), x(), y(), w(), h() + Fl::box_dh(box()), color());
```
CendioOssman, line 114: This means you draw outside of the widget? Which means you may need to do extra damage when cleaning things up?

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -324,6 +325,21 @@ Fl_Suggestion_Input::~Fl_Suggestion_Input()
...
+    else
+        Fl_Group::draw_box();
+
+    input->damage(FL_DAMAGE_ALL); // Ensure the input is fully redrawn
+    Fl_Group::draw_backdrop();
+    Fl_Group::draw_children();
+    Fl_Group::draw_label();
```
CendioOssman: Not just `Fl_Group::draw()`?

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -536,6 +552,7 @@ void Fl_Suggestion_Input::hide_suggestions()
     shown_suggestions.clear();
     dropdown_window->hide();
     highlight = "";
+    window()->damage(FL_DAMAGE_OVERLAY, x(), y(), w(), h() + Fl::box_dh(box()));
```
CendioOssman, line 346: Is this perhaps the cleanup redraw? Might need a comment.

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -570,6 +587,7 @@ void Fl_Suggestion_Input::dropdown_window_show()
         parent->add(dropdown_window);
 
     dropdown_window->show();
+    damage(FL_DAMAGE_OVERLAY);
```
CendioOssman, line 376: Same thing here. What does this do?

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -324,6 +325,21 @@ Fl_Suggestion_Input::~Fl_Suggestion_Input()
...
+{
+    // Draw the box taller to give the appearance of the rounded corners
+    // disappearing when the dropdown window is shown. Without the rounded
+    // bottom corners the dropdown visually sticks to the input in a nice way.
+    if (suggestions_visible())
+        draw_box(box(), x(), y(), w(), h() + Fl::box_dh(box()), color());
+    else
+        Fl_Group::draw_box();
+
+    input->damage(FL_DAMAGE_ALL); // Ensure the input is fully redrawn
```
CendioOssman: This looks a bit loop-prone. You report damage in a draw routine, which might be the result of damage?

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -633,7 +635,30 @@ void Fl_Suggestion_Input::dropdown_window_show()
...
+int Fl_Suggestion_Input::handle_system_event(void *, void *data)
+{
+    Fl_Suggestion_Input *self = (Fl_Suggestion_Input*) data;
+
+    // Since this dropdown window is separate it doesn't follow the parent
+    // window when the user moves it. FLTK doesn't have any event for a window
+    // being moved. We need to listen to system events - hide when moved.
+    if (self->window()->x() != self->last_x ||
+        self->window()->y() != self->last_y)
+        self->hide_suggestions();
```
CendioOssman: Probably safer to do this in an idle handler. The event you are responding to might be for the suggestion window. Which you are deleting if you hide it here.

vncviewer/fltk/Fl_Suggestion_Input.cxx
```diff
@@ -633,7 +635,30 @@ void Fl_Suggestion_Input::dropdown_window_show()
...
+
+    // FIXME: We're not getting any WM_MOVE events on Windows, this means the
+    //        dropdown stays open until the move is complete..
```
CendioOssman: Is this still true?

CendioOssman (changes requested):
Seems to work a lot better. Nice work! :)

Please squash things so we can see the final commit layout.

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Good idea, fixed now.

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Ah, yeah thats better! Fixed

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: No longer a local function after split, added to header file since it is needed in both `Fl_Suggestion_Input` and `Fl_Suggestion_Item_`.

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: No. I wasn't aware FLTK had hover colors anywhere?

vncviewer/fltk/Fl_Suggestion_Input.h
samhed: Fixed now. I added `_` as a suffix to class and file names to indicate that these are internal.

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Fixed

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Yep. As you noticed, we do extra damage in hide_suggestions().

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Fixed now

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Agreed, added a comment :+1:

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Added a comment

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: True, fixed by explicitly calling `Fl_Group::draw_child(*input)` instead

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Ah, good point. Fixed now

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Unfortunately, yes.

samhed: Fixed now.

vncviewer/fltk/event_dispatch_handler.h
```diff
@@ -0,0 +1,30 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#include <FL/Fl_Window.H>
+
+// Signature of fl_add_event_dispatch functions passed as parameters
+typedef int (*Fl_Event_Dispatch_Handler)(int event, Fl_Window *w, void *data);
+
+void fl_add_event_dispatch(Fl_Event_Dispatch_Handler ha, void *data);
```
CendioOssman, line 29: `DesktopWindow` needs to be moved over to this new interface.

vncviewer/fltk/Fl_Suggestion_Input.cxx
CendioOssman: If it is global, it probably needs to be namespaced better.

Can you put it under  `Fl_Suggestion_Input`?

vncviewer/fltk/Fl_Suggestion_Input.cxx
CendioOssman: No idea. I was just hoping we're sticking to a common FLTK pattern.

I assume this will break if you change the colors to a "dark mode"?

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: Sure, fixed now.

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: I don't think this is a common pattern. Yeah, it will probably break

vncviewer/fltk/event_dispatch_handler.h
samhed: Fixed in cb8238e

vncviewer/fltk/Fl_Suggestion_Input.cxx
samhed: I have no ambitions to work on dark mode support here, marking as resolved.

vncviewer/DesktopWindow.cxx
```diff
@@ -989,6 +990,12 @@ int DesktopWindow::handle(int event)
 }
 
 
+int DesktopWindow::dispatch_helper(int event, Fl_Window *w, void *data)
+{
+    DesktopWindow *self = (DesktopWindow*) data;
+    return self->fltkDispatch(event, w);
+}
+
```
CendioOssman: `fltkDispatch()` is a static method, so you don't need to figure out `this`.

It also means you don't need this wrapper. Modify the signature of `fltkDispatch()` instead.

vncviewer/DesktopWindow.cxx
```diff
@@ -989,6 +990,12 @@ int DesktopWindow::handle(int event)
...
 int DesktopWindow::fltkDispatch(int event, Fl_Window *win)
```
CendioOssman: You need to make sure this no longer calls `Fl::hande_()` as well.

vncviewer/DesktopWindow.cxx
samhed: Ah, nice.

Fixed in bb751c9

vncviewer/DesktopWindow.cxx
samhed: Fixed in bb751c9

samhed:
> It seems like the change to `DesktopWindow` breaks the event handling unfortunately. Something needs a bit of a redesign there.

Good catch, fixed in 95b6a0e. Tested the regrabOnFocus logic on Fedora 42 with GNOME on Wayland, macOS 26, and Windows 11.

PR #1986, opened 2025-09-09
Add wlroots support to w0vncserver

CendioHalim:
This PR adds wlroots support to w0vncserver with the use of the following Wayland protocols:
* `wlr-screencopy-unstable-v1`
* `wlr-virtual-pointer-unstable-v1`
*  `virtual-keyboard-unstable-v1`.

Just as the current xdg-desktop-portal implementation, this one has a few limitations compared to x0vncserver:
* No clipboard support
* No multi-head support
* Screen resizing only works when resizing the remote display (client can't initiate a resize).
* No local cursor rendering (this requires the newer `ext-image-copy-capture-v1` protocol AFAIK).
* Limited keyboard functionality
  * Currently, the servers' keyboard layout is used, and LED states are not synced.
  
The Wayland object class hierarchy (`WObject` & subclasses) is not the cleanest and could be improved.


CMakeLists.txt
```diff
@@ -409,9 +413,12 @@ if(UNIX AND NOT APPLE)
...
     find_package(Uuid REQUIRED)
+
+    # wlroots specific
+    find_package(Wayland REQUIRED)
   endif()
   if(NOT GLIB_FOUND OR NOT GIO_FOUND OR NOT GOBJECT_FOUND OR NOT PIPEWIRE_FOUND
-     OR NOT UUID_FOUND)
+     OR NOT UUID_FOUND OR NOT WAYLAND_FOUND)
     set(ENABLE_WAYLAND 0)
     message(WARNING "GLib, Gio, Gobject, PipeWire or Uuid NOT found. w0vncserver disabled.")
```
CendioOssman: "or Wayland"

cmake/Modules/FindWayland.cmake
```diff
@@ -0,0 +1,50 @@
...
+mark_as_advanced(Wayland_LIBRARIES)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Wayland
+  REQUIRED_VARS
+    Wayland_client_INCLUDE_DIR Wayland_LIBRARIES Wayland_scanner_EXECUTABLE
+)
+
+if(Wayland_FOUND)
+  set(WAYLAND_INCLUDE_DIRS ${Wayland_client_INCLUDE_DIR} ${Wayland_protocols_INCLUDE_DIR} ${Wayland_server_INCLUDE_DIR})
```
CendioOssman: Where did protocols and server come from?

cmake/Modules/FindWayland.cmake
```diff
@@ -0,0 +1,50 @@
...
+  the libraries to link against to use libwayland.
+``WAYLAND_FOUND``
+  TRUE if found
+#]=======================================================================]
+
+find_package(PkgConfig QUIET)
+
+if (PKG_CONFIG_FOUND)
+  pkg_check_modules(PC_Wayland_client QUIET wayland-client)
+endif()
```
CendioOssman, line 20: Given that this is just the client library, should the module also be "FindWaylandClient"?

cmake/Modules/FindWayland.cmake
```diff
@@ -0,0 +1,50 @@
...
+find_library(Wayland_LIBRARIES NAMES wayland-client
+  HINTS
+    ${PC_Wayland_client_LIBRARY_DIRS}
+)
+mark_as_advanced(Wayland_LIBRARIES)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(Wayland
+  REQUIRED_VARS
+    Wayland_client_INCLUDE_DIR Wayland_LIBRARIES Wayland_scanner_EXECUTABLE
```
CendioOssman: Library variable first, as that is what it prints.

contrib/packages/rpm/el8/SPECS/tigervnc.spec
```diff
@@ -24,6 +24,7 @@ BuildRequires:  freetype-devel, libjpeg-turbo-devel, pam-devel
 BuildRequires:  gnutls-devel, nettle-devel, gmp-devel
 BuildRequires:  zlib-devel
 BuildRequires:  libuuid-devel, glib2-devel, pipewire-devel
+BuildRequires:  wayland-devel, libwayland-client
```
CendioOssman: "libwayland-client" should be implied from "wayland-devel".

unix/w0vncserver/CMakeLists.txt
```diff
@@ -15,12 +24,41 @@ add_executable(w0vncserver
...
+)
+
+foreach(P ${WLR_PROTOCOLS})
+  set(XML ${CMAKE_CURRENT_SOURCE_DIR}/protocols/${P}.xml)
+  set(H ${CMAKE_CURRENT_BINARY_DIR}/${P}.h)
+  set(C ${CMAKE_CURRENT_BINARY_DIR}/${P}.c)
+
+  add_custom_command(
+    OUTPUT ${C}
+    BYPRODUCTS ${C}
```
CendioOssman: `${H}`?

unix/w0vncserver/CMakeLists.txt
```diff
@@ -15,12 +24,41 @@ add_executable(w0vncserver
...
+    COMMAND ${WAYLAND_SCANNER_EXECUTABLE} private-code ${XML} ${C}
+    COMMAND ${WAYLAND_SCANNER_EXECUTABLE} client-header ${XML} ${H}
+    DEPENDS ${XML}
+    COMMENT "Generating ${P}.c and ${P}.h"
```
CendioOssman: Might want to use `${C}` and `${H}` here as well to avoid surprises.

unix/w0vncserver/CMakeLists.txt
```diff
@@ -15,12 +24,41 @@ add_executable(w0vncserver
...
+target_sources(w0vncserver PRIVATE ${GEN_WAYLAND_SRCS})
+
 target_include_directories(w0vncserver SYSTEM PUBLIC ${GLIB_INCLUDE_DIRS})
 target_include_directories(w0vncserver SYSTEM PUBLIC ${GIO_INCLUDE_DIRS})
 target_include_directories(w0vncserver SYSTEM PUBLIC ${GOBJECT_INCLUDE_DIRS})
 target_include_directories(w0vncserver SYSTEM PUBLIC ${PIXMAN_INCLUDE_DIRS})
 target_include_directories(w0vncserver SYSTEM PUBLIC ${PIPEWIRE_INCLUDE_DIRS})
 target_include_directories(w0vncserver SYSTEM PUBLIC ${UUID_INCLUDE_DIRS})
+target_include_directories(w0vncserver SYSTEM PUBLIC ${WAYLAND_INCLUDE_DIRS})
+target_include_directories(w0vncserver PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
```
CendioOssman: This seems inappropriate.

unix/w0vncserver/w0vncserver.cxx
```diff
@@ -35,6 +35,7 @@
 #include <core/Logger_stdio.h>
 #include <core/LogWriter.h>
 
+#include "wlr/WlrDesktop.h"
```
CendioOssman: Why is this in a subdirectory, but PortalDesktop is not? Makes it hard to follow what is shared and what isn't.

unix/w0vncserver/wayland/GWaylandSource.h
```diff
@@ -0,0 +1,45 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this software; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ * USA.
+ */
+
+#ifndef __G_WLR_SOURCE_H__
+#define __G_WLR_SOURCE_H__
```
CendioOssman: I guess you changed the directory for this at some point?

unix/w0vncserver/wayland/Desktop.cxx
```diff
@@ -0,0 +1,115 @@
...
+void WlrDesktop::start()
+{
+  std::function<void()> desktopReadyCb = [this]() {
+    server->setPixelBuffer(pb);
+  };
+
+  pb = new WlrPixelBuffer(display, output, server, desktopReadyCb);
+
+  wlrSource = new GWaylandSource(display);
+  wlrSource->attach(g_main_loop_get_context(loop));
```
CendioOssman: Should perhaps this be implicit when you create a `WDisplay`?

unix/w0vncserver/wlr/WlrPixelBuffer.cxx
```diff
@@ -0,0 +1,113 @@
...
+    // FIXME: We're likely to make two calls to startFrameCapture()
+    // for our first frame. Let's filter out the message the first time
+    // as it is annoying.
+    if (!firstFrame)
+      vlog.debug("Frame capture already in progress, skipping frame");
+    return;
+  }
+
+  if (!firstFrame)
+    server->blockUpdates();
```
CendioOssman: This is pretty heavy handed. It would be preferred it this is blocked very briefly. Ideally without waiting for something external.

unix/w0vncserver/wlr/WlrDesktop.cxx
```diff
@@ -0,0 +1,115 @@
...
+  wlrSource = nullptr;
+
+  delete pb;
+  pb = nullptr;
+}
+
+void WlrDesktop::frameTick(uint64_t /* msc */)
+{
+  // FIXME: Should we use the monitor refresh rate instead?
+  pb->captureFrame();
```
CendioOssman: This is a bit backwards for how the system is designed to work. The desktop should be pushing changes to the server, not the other way around.

Can we get a notification when something changes instead? And if reading is optional, it would be nice if we can delay reading until needed. Not every screen updated will be sent to client.

CMakeLists.txt
```diff
@@ -416,9 +417,10 @@ if(UNIX AND NOT APPLE)
 
     # wlroots specific
     find_package(Wayland REQUIRED)
+    find_package(Xkbcommon REQUIRED)
   endif()
   if(NOT GLIB_FOUND OR NOT GIO_FOUND OR NOT GOBJECT_FOUND OR NOT PIPEWIRE_FOUND
-     OR NOT UUID_FOUND OR NOT WAYLAND_FOUND)
+     OR NOT UUID_FOUND OR NOT WAYLAND_FOUND OR NOT XKBCOMMON_FOUND)
     set(ENABLE_WAYLAND 0)
     message(WARNING "GLib, Gio, Gobject, PipeWire or Uuid NOT found. w0vncserver disabled.")
```
CendioOssman: "or xkbcommon"

unix/w0vncserver/wlr/WlrVirtualKeyboard.cxx
```diff
@@ -0,0 +1,66 @@
...
+{
+  if (manager)
+    zwp_virtual_keyboard_manager_v1_destroy(manager);
+  if (keyboard)
+    zwp_virtual_keyboard_v1_destroy(keyboard);
+}
+
+void WlrVirtualKeyboard::keyEvent(uint32_t keysym, uint32_t keycode, bool down)
+{
+  zwp_virtual_keyboard_v1_key(keyboard, keysym, keycode, down);
```
CendioOssman: This seems wrong. Where are things translated from RFB keycode to Wayland keycode?

unix/w0vncserver/wayland/WKeyboard.cxx
```diff
@@ -20,16 +20,23 @@
...
+#include "../w0vncserver.h"
 #include "WDisplay.h"
 #include "WSeat.h"
 #include "WKeyboard.h"
 
 static core::LogWriter vlog("WKeyboard");
 
+extern const unsigned short code_map_qnum_to_xorgevdev[];
+extern const unsigned int code_map_qnum_to_xorgevdev_len;
+
```
CendioOssman: These aren't used here?

unix/w0vncserver/wayland/WaylandDesktop.cxx
```diff
@@ -141,5 +141,6 @@ bool WlrDesktop::available()
   WDisplay display;
 
   return display.interfaceAvailable("zwlr_screencopy_manager_v1") &&
-         display.interfaceAvailable("zwlr_virtual_pointer_manager_v1");
+         display.interfaceAvailable("zwlr_virtual_pointer_manager_v1") &&
+         display.interfaceAvailable("zwp_virtual_keyboard_manager_v1");
```
CendioOssman, line 154: This should have been in the previous commit?

unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
```diff
@@ -20,6 +20,11 @@
 #include <config.h>
 #endif
 
+#include <time.h>
+
+#include <core/LogWriter.h>
+#include <xkbcommon/xkbcommon-keysyms.h>
```
CendioOssman, line 26: We try to do external includes before our own.

unix/w0vncserver/wlr/WlrVirtualKeyboard.cxx
```diff
@@ -58,9 +68,34 @@ WlrVirtualKeyboard::~WlrVirtualKeyboard()
...
-void WlrVirtualKeyboard::keyEvent(uint32_t keysym, uint32_t keycode, bool down)
+void WlrVirtualKeyboard::keyEvent(uint32_t /* keysym */, uint32_t keycode,
+                                  bool down)
 {
-  zwp_virtual_keyboard_v1_key(keyboard, keysym, keycode, down);
+  timespec ts;
+  uint32_t time;
+  bool updated;
+  if (keycode < code_map_qnum_to_xorgevdev_len)
+    keycode = code_map_qnum_to_xorgevdev[keycode];
```
CendioOssman: Is Wayland well-defined to always use evdev codes? Or should we map things via the keyboard layout like we do in vncviewer?

unix/w0vncserver/wlr/WlrVirtualKeyboard.cxx
```diff
@@ -58,9 +68,34 @@ WlrVirtualKeyboard::~WlrVirtualKeyboard()
...
+
+  // FIXME: look at keysym
```
CendioOssman: I guess this is a major difference between portals and wlroots? Should probably mention that somewhere.

unix/w0vncserver/wayland/WOutput.cxx
```diff
@@ -90,17 +97,27 @@ void WOutput::handleGeometry(int32_t /* x */, int32_t /* y */,
...
     .width = width,
     .height = height,
     .refresh = refresh,
   };
+
+  if (mode.width) {
+    resized = true;
+    vlog.debug("Resized from %dx%d to %dx%d", mode.width, mode.height,
+               mode_.width, mode_.height);
+  }
```
CendioOssman: Or put this earlier and you don't have to have temporary variables.

unix/w0vncserver/wayland/WOutput.h
```diff
@@ -39,6 +39,9 @@ class WOutput : public WObject {
   WOutput(WDisplay* display);
   ~WOutput();
 
+  void resizeComplete();
+  bool hasResized() const { return resized; }
```
CendioOssman: This seems like state that WlrPixelBuffer should keep, not WOutput.

unix/w0vncserver/wlr/WlrPixelBuffer.cxx
```diff
@@ -111,3 +116,19 @@ void WlrPixelBuffer::captureFrameDone()
...
+  setBuffer(output->getWidth(), output->getHeight(), getBufferData(),
+            output->getWidth());
+
+  server->setPixelBuffer(this);
+
+  if (captureInProgress)
+    server->unblockUpdates();
+  captureInProgress = false;
+
+  captureFrame();
```
CendioOssman: This looks suspicious. If a capture is in progress, then I assume it needs to be aborted? And the buffer is not valid at this point?

unix/w0vncserver/wlr/WlrPixelBuffer.h
```diff
@@ -27,12 +27,8 @@
 
 namespace rfb { class VNCServer; }
 
-struct wl_buffer;
-struct BufferInfo;
 class WOutput;
 class WDisplay;
-class WShm;
-class WShmPool;
```
CendioOssman: This cleanup looks like it should have been elsewhere.

CendioOssman (changes requested):
Looks promising. Nice abstraction of all the Wayland objects!

There are a few questions from my end. Please see the comments.

cmake/Modules/FindWayland.cmake
CendioHalim: Oh right, these were from an earlier iteration. I'll remove them.

cmake/Modules/FindWayland.cmake
CendioHalim: Good point

unix/w0vncserver/wayland/GWaylandSource.h
CendioHalim: Directory and class name even :smile:

unix/w0vncserver/CMakeLists.txt
CendioHalim: Moved it & added a comment to make it more clear

unix/w0vncserver/wlr/WlrDesktop.cxx
CendioHalim: When we call `zwlr_screencopy_frame_v1_copy_with_damage`, it will wait until damage occurs and notify us (and do the screen reading at the same time). From what I can tell, the mechanisms for getting damage information and getting the actual screen contents are tightly coupled :/

Not sure how to solve this nicely without some kind of rate limiting. I'll have another go at this and see if there is something I'm missing.

CendioOssman: I'm afraid we cannot have things blocked that long. I think you'll need to have a shadow buffer in that case, and copy things between them.

unix/w0vncserver/wayland/Desktop.cxx
CendioHalim: I wanted to avoid coupling the main loop with anything as much as possible, especially core Wayland objects.

unix/w0vncserver/wlr/WlrVirtualKeyboard.cxx
CendioHalim: Yeah this is totally wrong, fixed now.

unix/w0vncserver/w0vncserver.cxx
CendioHalim: I've added a commit that moves Portals related classes to the `portals/` directory

unix/w0vncserver/wlr/WlrDesktop.cxx
CendioHalim: I've added a shadow framebuffer which should simplify the code a bit.

unix/w0vncserver/wlr/WlrPixelBuffer.cxx
CendioHalim: I've now re-written this so that only one capture at a time is possible.

unix/w0vncserver/wlr/WlrVirtualKeyboard.cxx
CendioHalim: I've changed the keyboard code to only use keysyms. As of now, it is limited to using keysyms available in the remote's current keymap (same as in portals).

unix/w0vncserver/wlr/WlrPixelBuffer.cxx
```diff
@@ -0,0 +1,142 @@
...
+      return;
+    }
+
+    // Utilize a shadow framebuffer, as we have no way of accessing
+    // the screen contents directly.
+    shadowFramebuffer = new uint8_t[output->getWidth() *
+                                    output->getHeight() * (format.bpp / 8)];
+
+    setBuffer(output->getWidth(), output->getHeight(),
+              shadowFramebuffer, output->getWidth());
```
CendioOssman: Perhaps easier to use `ManagedPixelBuffer` at this point?

unix/w0vncserver/wlr/WlrPixelBuffer.cxx
```diff
@@ -0,0 +1,142 @@
...
+  srcStride = width();
+  dstBuffer = getBufferRW(getRect(), &dstStride);
+
+  ret = pixman_blt((uint32_t*)srcBuffer, (uint32_t*)dstBuffer,
+                    srcStride, dstStride, format.bpp / 8,
+                    format.bpp / 8, 0, 0, 0,0, width(), height());
+  commitBufferRW(getRect());
+
+  if (!ret)
+    imageRect(format, {0, 0, width(), height()}, srcBuffer, srcStride);
```
CendioOssman: Should probably use the damage information as this is likely a bottle neck.

CendioOssman (changes requested):

unix/w0vncserver/wayland/WKeyboard.cxx
```diff
@@ -0,0 +1,241 @@
...
+                                                  XKB_STATE_MODS_LOCKED);
+  modifiers.group = xkb_state_serialize_mods(context->state,
+                                             XKB_STATE_LAYOUT_EFFECTIVE);
+
+  previousLedState = modifiers.ledState;
+  updateLEDState();
+
+  changedLeds = previousLedState != modifiers.ledState;
+
+  return changedMods | changedLeds;
```
CendioOssman: This should be a `||`.

But more importantly, this shouldn't look at `changedLeds` at all? The return value here is used to see if the virtual keyboard state needs to be updated. And led state is not a separate xkb state, so it should not be relevant there?

unix/w0vncserver/wayland/WKeyboard.cxx
```diff
@@ -0,0 +1,241 @@
...
+}
+
+void WKeyboard::handleModifiers(uint32_t /* serial */, uint32_t modsDepressed,
+                                uint32_t modsLatched,
+                                uint32_t modsLocked, uint32_t group)
+{
+  modifiers.modsDepressed = modsDepressed;
+  modifiers.modsLatched = modsLatched;
+  modifiers.modsLocked = modsLocked;
+  modifiers.group = group;
```
CendioOssman: Can probably drop these and avoid duplicating state?

unix/w0vncserver/wayland/WKeyboard.cxx
```diff
@@ -0,0 +1,241 @@
...
+    ledState |= rfb::ledScrollLock;
+  if (xkb_state_led_name_is_active(context->state, XKB_LED_NAME_NUM))
+    ledState |= rfb::ledNumLock;
+  if (xkb_state_led_name_is_active(context->state, XKB_LED_NAME_CAPS))
+    ledState |= rfb::ledCapsLock;
+
+  modifiers.ledState = ledState;
+
+  setLEDstate(modifiers.ledState);
+}
```
CendioOssman: Missing end of line.

unix/w0vncserver/wlr/WlrDesktop.cxx
```diff
@@ -0,0 +1,159 @@
...
+  delete output;
+  delete display;
+}
+
+void WlrDesktop::init(rfb::VNCServer* vs)
+{
+  server = vs;
+
+  seat = new WSeat(display, std::bind(&rfb::VNCServer::setLEDState,
+                                      server, std::placeholders::_1));
```
CendioOssman: You could also keep this in the constructor, and create a `WlrDesktop::setLEDState()` that has a guard if it is called before `server` is set.

unix/w0vncserver/wlr/WlrPixelBuffer.cxx
CendioOssman: The switch should mean that this call is removed and replaced with `setSize()`.

unix/w0vncserver/wlr/WlrPixelBuffer.cxx
```diff
@@ -0,0 +1,153 @@
...
+void WlrPixelBuffer::syncBuffers()
+{
+  int srcStride;
+  int dstStride;
+  uint8_t* srcBuffer;
+  uint8_t* dstBuffer;
+  pixman_bool_t ret;
+  core::Rect rect;
+
+  rect = getDamage();
```
CendioOssman: Ideally, you keep this as a region and iterate over all rects in it.

unix/w0vncserver/wlr/WlrPixelBuffer.cxx
CendioHalim: Good point, this has been fixed now.

unix/w0vncserver/wayland/WKeyboard.cxx
CendioHalim: Re-did the LED state tracking, should be better now

unix/w0vncserver/wlr/WlrDesktop.cxx
CendioHalim: Good point!

CendioHalim:
Comments have been addressed.

Also namespaced the Wayland objects and move around some files to get a better overall structure.

unix/w0vncserver/wayland/Desktop.h
```diff
@@ -0,0 +1,77 @@
...
+#define __WAYLAND_DESKTOP_H__
+
+#include <glib.h>
+#include <stdint.h>
+
+#include <rfb/SDesktop.h>
+
+namespace rfb { class VNCServer; }
+
+namespace wayland {
```
CendioOssman: You sure you want to put these classes under the Wayland namespace as well, and not just the object wrappers? It's a bit asymmetrical compared to the portals code.

CendioOssman: E.g. `wayland::GWaylandSource` might be a bit redundant.

unix/w0vncserver/wayland/Desktop.cxx
```diff
@@ -0,0 +1,164 @@
...
+
+  delete virtualPointer;
+  virtualPointer = nullptr;
+
+  delete pb;
+  pb = nullptr;
+}
+
+void Desktop::pointerEvent(const core::Point& pos,
+                                  uint16_t buttonMask)
```
CendioOssman: Something wonky with the indentation here. Haven't checked if other places also got damaged in the rename.

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
```diff
@@ -0,0 +1,159 @@
...
+  srcStride = width();
+  dstBuffer = getBufferRW(getRect(), &dstStride);
+
+  region.get_rects(&rects);
+  for (core::Rect &rect : rects) {
+    ret = pixman_blt((uint32_t*)srcBuffer, (uint32_t*)dstBuffer,
+                     srcStride, dstStride, format.bpp, format.bpp,
+                     rect.tl.x, rect.tl.y, rect.tl.x, rect.tl.y,
+                     rect.width(), rect.height());
+    commitBufferRW(rect);
```
CendioOssman, line 148: There should be a symmetry between `getBufferRW()` and `commitBufferRW()`. So should probably move something around here.

unix/w0vncserver/wayland/Desktop.h
CendioHalim: To clarify, do you mean only the Desktop & GWaylandSource classes?

CendioOssman: No, I'd say everything that isn't a wayland object. So probably everything outside the `objects/` folder?

CendioHalim: Even the PixelBuffer, which inherits `wayland::ScreencopyManager`?

CendioOssman: Yes, since it is primarily glue between different parts of the code, and not a representation of a specific Wayland object/protocol. Just like the Wayland Desktop class is also primarily glue between components.

That means they likely need to get back prefixed names, just like e.g. PortalDesktop.

CendioHalim: Got it, all comments should have been addressed now

PR #1998, opened 2025-10-06
PortalDesktop: Enable using restore_token

CendioHalim:
Restore tokens can be used to save/restore a DBus session. This makes it possible to re-use a previous session, bypassing the remote desktop dialog that pops up.

In this commit, a new parameter "portalAskDisplayChoice" is added where users can choose between three options:

  1. "Always" - The remote desktop dialog will always pop up and restore tokens are not used. This is identical to the behaviour prior to this commit.

  2. "Once" - Only display the remote desktop dialogue the first time a user connects. Subsequent connections will re-use the same session that was created when the first user connected. The session is valid as long as the server is not closed. Users have to restart the server to get the dialogue again.

  3. "Never" - Display the remote desktop dialogue once, and remember the choice even after restarts. The restore token is stored to disk and can be loaded when the RemoteDesktop session starts. Users have to clear the file, or use another option to get the dialogue again.

The default option right now is "Once", which should be helpful for most users. Having "Never" as the default can be problematic as it might not be obvious to users how to "reset" their display choice.

### DIFF
--- a/unix/w0vncserver/portals/RemoteDesktop.cxx
+++ b/unix/w0vncserver/portals/RemoteDesktop.cxx
@@ -21,15 +21,21 @@
 #endif
 
 #include <stdint.h>
+#include <stdio.h>
 #include <assert.h>
 #include <sys/select.h>
+#include <uuid/uuid.h>
 #include <linux/input-event-codes.h>
+
+#include <string>
 
 #include <glib.h>
 #include <gio/gunixfdlist.h>
 
+#include <core/Configuration.h>
 #include <core/LogWriter.h>
 #include <core/string.h>
+#include <core/xdgdirs.h>
 
 #include "../w0vncserver.h"
 #include "portalConstants.h"
@@ -51,6 +57,12 @@ core::BoolParameter localCursor("experimentalPortalLocalCursor",
                                 "[EXPERIMENTAL] Render cursor locally",
                                 false);
 
+core::EnumParameter askDisplayChoice("portalAskDisplayChoice",
+                                      "Ask about display choice on connect",
+                                      {"Always", "Once", "Never"}, "Once");
+
+static const char* RESTORE_TOKEN_FILENAME =  "tigervnc.restoretoken";
+
 static core::LogWriter vlog("RemoteDesktop");
 
 static int getInputCode(uint32_t button)
@@ -71,13 +83,15 @@ static int getInputCode(uint32_t button)
   }
 }
 
-RemoteDesktop::RemoteDesktop(std::function<void(int fd, uint32_t nodeId)>
+RemoteDesktop::RemoteDesktop(std::string restoreToken_,
+                             std::function<void(int fd, uint32_t nodeId)>
                                startPipewireCb_,
                              std::function<void(const char*)>
                                cancelStartCb_)
   : sessionStarted(false), oldButtonMask(0), selectedDevices(0),
     sessionHandle(""), remoteDesktop(nullptr), screenCast(nullptr),
-    session(nullptr), startPipewireCb(startPipewireCb_),
+    session(nullptr), restoreToken(restoreToken_),
+    startPipewireCb(startPipewireCb_),
     cancelStartCb(cancelStartCb_)
 {
   remoteDesktop = new PortalProxy("org.freedesktop.portal.Desktop",
@@ -239,15 +253,19 @@ void RemoteDesktop::selectDevices()
   requestHandleToken = remoteDesktop->newHandle();
 
   g_variant_builder_init(&optionsBuilder, G_VARIANT_TYPE_VARDICT);
+
   g_variant_builder_add(&optionsBuilder, "{sv}", "types",
                         g_variant_new_uint32(DEV_KEYBOARD | DEV_POINTER));
   g_variant_builder_add(&optionsBuilder, "{sv}", "persist_mode",
-                        g_variant_new_uint32(0));
-
-  // FIXME: Do we want restore_token?
+                        g_variant_new_uint32(PERSIST_UNTIL_REVOKED));
 
   g_variant_builder_add(&optionsBuilder, "{sv}", "handle_token",
                         g_variant_new_string(requestHandleToken.c_str()));
+
+  if (loadRestoreToken()) {
+    g_variant_builder_add(&optionsBuilder, "{sv}", "restore_token",
+                          g_variant_new_string(restoreToken.c_str()));
+  }
 
   params = g_variant_new("(oa{sv})", sessionHandle.c_str(), &optionsBuilder);
 
@@ -375,6 +393,7 @@ void RemoteDesktop::handleStart(GVariant* parameters)
   GVariant* result;
   GVariant* streams_;
   GVariant* devices;
+  GVariant* restoreToken_;
 
   assert(!sessionStarted);
 
@@ -410,6 +429,14 @@ void RemoteDesktop::handleStart(GVariant* parameters)
     g_variant_unref(streams_);
     fatal_error("Failed to parse streams");
     return;
+  }
+
+  restoreToken_ = g_variant_lookup_value(result, "restore_token",
+                                         G_VARIANT_TYPE_STRING);
+
+  if (restoreToken_ && askDisplayChoice == "Never") {
+    storeRestoreToken(g_variant_get_string(restoreToken_, nullptr));
+    g_variant_unref(restoreToken_);
   }
 
   openPipewireRemote();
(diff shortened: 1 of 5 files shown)


CendioHalim:
Note that restore_token was added in version 2 of the RemoteDesktop interface. We should probably add a version check to this.

CendioHalim:
Also, the man page should be updated with the new parameter

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -51,6 +57,12 @@ core::BoolParameter localCursor("experimentalPortalLocalCursor",
                                 "[EXPERIMENTAL] Render cursor locally",
                                 false);
 
+core::EnumParameter askDisplayChoice("portalAskDisplayChoice",
```
CendioOssman: Why the "portal" prefix? Seems like technical detail that end users shouldn't need to know.

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -51,6 +57,12 @@ core::BoolParameter localCursor("experimentalPortalLocalCursor",
...
+                                      "Ask about display choice on connect",
```
CendioOssman: Don't need an essay, but probably a few more words to set a context

CendioOssman: Also note that we don't automatically print the choices, so they need to be part of the description string.

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -51,6 +57,12 @@ core::BoolParameter localCursor("experimentalPortalLocalCursor",
...
+                                      {"Always", "Once", "Never"}, "Once");
+
+static const char* RESTORE_TOKEN_FILENAME =  "tigervnc.restoretoken";
```
CendioOssman: We're already in a TigerVNC specific directory, so "restoretoken" should be enough.

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -239,16 +253,20 @@ void RemoteDesktop::selectDevices()
...
 
   g_variant_builder_init(&optionsBuilder, G_VARIANT_TYPE_VARDICT);
+
   g_variant_builder_add(&optionsBuilder, "{sv}", "types",
                         g_variant_new_uint32(DEV_KEYBOARD | DEV_POINTER));
   g_variant_builder_add(&optionsBuilder, "{sv}", "persist_mode",
-                        g_variant_new_uint32(0));
-
-  // FIXME: Do we want restore_token?
+                        g_variant_new_uint32(PERSIST_UNTIL_REVOKED));
```
CendioOssman: Any other values that are useful depending on our setting?

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -412,6 +431,14 @@ void RemoteDesktop::handleStart(GVariant* parameters)
     return;
   }
 
+  restoreToken_ = g_variant_lookup_value(result, "restore_token",
+                                         G_VARIANT_TYPE_STRING);
+
+  if (restoreToken_ && askDisplayChoice == "Never") {
```
CendioOssman: Should you be checking the choice here? Isn't that done by `storeRestoreToken()`?

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -412,6 +431,14 @@ void RemoteDesktop::handleStart(GVariant* parameters)
...
+    storeRestoreToken(g_variant_get_string(restoreToken_, nullptr));
+    g_variant_unref(restoreToken_);
```
CendioOssman: I guess this leaks memory depending on the choice setting?

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -499,3 +526,134 @@ bool RemoteDesktop::parseStreams(GVariant* streams)
...
 }
+
+bool RemoteDesktop::loadRestoreToken()
+{
+  char filepath[PATH_MAX];
+  FILE* f;
+  const char* stateDir;
+  char restoreToken_[37];
+
+  // Allows users to reset the restore token by not choosing "Never"
```
CendioOssman: That's a bit convoluted. A small script helper to reset the token would be preferred.

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -499,3 +526,134 @@ bool RemoteDesktop::parseStreams(GVariant* streams)
...
+    restoreToken = restoreToken_;
+    return true;
+  }
+  fclose(f);
+
+  vlog.error("Could not read restore token from \"%s\"", filepath);
+  return false;
+}
+
+bool RemoteDesktop::storeRestoreToken(const char* restoreToken_)
```
CendioOssman: You could also use a name like "newToken" to avoid weird underscore workarounds.

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -499,3 +526,134 @@ bool RemoteDesktop::parseStreams(GVariant* streams)
...
+  if (!stateDir) {
+    vlog.error("Could not get state directory");
+    return false;
+  }
+
+  snprintf(filepath, sizeof(filepath), "%s/%s", stateDir, RESTORE_TOKEN_FILENAME);
+
+  f = fopen(filepath, "w");
+  if (!f) {
+    vlog.error("Could not restore token from \"%s\": %s", filepath, strerror(errno));
```
CendioOssman: Wrong message.

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -499,3 +526,134 @@ bool RemoteDesktop::parseStreams(GVariant* streams)
...
+
+  stateDir = core::getvncstatedir();
+  if (!stateDir) {
+    vlog.error("Could not get state directory");
+    return false;
+  }
+
+  snprintf(filepath, sizeof(filepath), "%s/%s", stateDir, RESTORE_TOKEN_FILENAME);
+
+  f = fopen(filepath, "w");
```
CendioOssman: Not just remove it?

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -499,3 +526,134 @@ bool RemoteDesktop::parseStreams(GVariant* streams)
...
+  stateDir = core::getvncstatedir();
+  if (!stateDir) {
+    vlog.error("Could not get state directory");
+    return false;
+  }
+
+  snprintf(filepath, sizeof(filepath), "%s/%s", stateDir, RESTORE_TOKEN_FILENAME);
+  f = fopen(filepath, "r");
+  if (!f) {
+    vlog.error("Could not open \"%s\": %s", filepath, strerror(errno));
```
CendioOssman: It not existing will be a common case. Do we really want to complain about that?

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -499,3 +526,134 @@ bool RemoteDesktop::parseStreams(GVariant* streams)
...
+      return false;
+    }
+  }
+
+  snprintf(filepath, sizeof(filepath), "%s/%s", stateDir,
+           RESTORE_TOKEN_FILENAME);
+
+  f = fopen(filepath, "w");
+  if (!f) {
+    vlog.error("Could not restore token from \"%s\": %s", filepath, strerror(errno));
```
CendioOssman: Wrong message.

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: Unclear, the specification says:

> persist_mode (u)
>  
> How this session should persist. Default is 0. Accepted values are:
>  
>     0: Do not persist (default)
> 
>     1: Permissions persist as long as the application is running
> 
>     2: Permissions persist until explicitly revoked

I'm not sure how to interpret `1`, but it won't even let me get a restore token anyway.

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: Yeah, that would simplify the code a bit

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: Removed the prefix. Should it be documented that this feature only applies to the Portals implementation?

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: Good point

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: Added a utility `w0vncserver-forget` that does this instead.

PR #1999, opened 2025-10-08
w0vncserver: Add raw keyboard support

CendioHalim:
This add support for using keycodes in the WaylandDesktop and PortalDesktop.

### DIFF
--- a/unix/w0vncserver/portals/RemoteDesktop.cxx
+++ b/unix/w0vncserver/portals/RemoteDesktop.cxx
@@ -96,8 +96,9 @@ RemoteDesktop::~RemoteDesktop()
   delete remoteDesktop;
 }
 
-void RemoteDesktop::keyEvent(uint32_t keysym, uint32_t /* keycode */,
-                             bool down)
+void RemoteDesktop::notifyKeyboardKeysym(uint32_t keysym,
+                                         uint32_t /* keycode */,
+                                         bool down)
 {
   GVariantBuilder optionsBuilder;
   GVariant* params;
@@ -116,7 +117,8 @@ void RemoteDesktop::keyEvent(uint32_t keysym, uint32_t /* keycode */,
   remoteDesktop->call("NotifyKeyboardKeysym", params);
 }
 
-void RemoteDesktop::pointerEvent(int x, int y, uint16_t buttonMask)
+void RemoteDesktop::notifyPointerMotionAbsolute(int x, int y,
+                                                uint16_t buttonMask)
 {
   GVariantBuilder optionsBuilder;
   GVariant* params;
@@ -137,15 +139,15 @@ void RemoteDesktop::pointerEvent(int x, int y, uint16_t buttonMask)
   for (int32_t i = 0; i < BUTTONS; i++) {
     if ((buttonMask ^ oldButtonMask) & (1 << i)) {
       if (i > 2 && i < 7)
-        handleScrollWheel(i);
+        notifyPointerAxisDiscrete(i);
       else
-        handleButton(i, buttonMask & (1 << i));
+        notifyPointerButton(i, buttonMask & (1 << i));
     }
   }
   oldButtonMask = buttonMask;
 }
 
-void RemoteDesktop::handleButton(int32_t button, bool down)
+void RemoteDesktop::notifyPointerButton(int32_t button, bool down)
 {
   GVariantBuilder optionsBuilder;
   GVariant* params;
@@ -162,7 +164,7 @@ void RemoteDesktop::handleButton(int32_t button, bool down)
   remoteDesktop->call("NotifyPointerButton", params);
 }
 
-void RemoteDesktop::handleScrollWheel(int32_t button)
+void RemoteDesktop::notifyPointerAxisDiscrete(int32_t button)
 {
   GVariantBuilder optionsBuilder;
   GVariant* params;
--- a/unix/w0vncserver/portals/RemoteDesktop.cxx
+++ b/unix/w0vncserver/portals/RemoteDesktop.cxx
@@ -96,9 +96,7 @@ RemoteDesktop::~RemoteDesktop()
   delete remoteDesktop;
 }
 
-void RemoteDesktop::notifyKeyboardKeysym(uint32_t keysym,
-                                         uint32_t /* keycode */,
-                                         bool down)
+void RemoteDesktop::notifyKeyboardKeysym(uint32_t keysym, bool down)
 {
   GVariantBuilder optionsBuilder;
   GVariant* params;
@@ -107,14 +105,29 @@ void RemoteDesktop::notifyKeyboardKeysym(uint32_t keysym,
   if (!(selectedDevices & DEV_KEYBOARD))
     return;
 
-  // FIXME: Make use of keycodes?
-
   state = down ? 1 : 0;
 
   g_variant_builder_init(&optionsBuilder, G_VARIANT_TYPE("a{sv}"));
   params = g_variant_new("(oa{sv}iu)", sessionHandle.c_str(),
                          &optionsBuilder, keysym, state);
   remoteDesktop->call("NotifyKeyboardKeysym", params);
+}
+
+void RemoteDesktop::notifyKeyboardKeycode(uint32_t xtcode, bool down)
+{
+  GVariantBuilder optionsBuilder;
+  GVariant* params;
+  uint32_t state;
+
+  if (!(selectedDevices & DEV_KEYBOARD))
+    return;
+
+  state = down ? 1 : 0;
+
+  g_variant_builder_init(&optionsBuilder, G_VARIANT_TYPE("a{sv}"));
+  params = g_variant_new("(oa{sv}iu)", sessionHandle.c_str(),
+                         &optionsBuilder, xtcode, state);
+  remoteDesktop->call("NotifyKeyboardKeycode", params);
 }
 
 void RemoteDesktop::notifyPointerMotionAbsolute(int x, int y,
--- a/unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
+++ b/unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
@@ -29,6 +29,7 @@
 #include <rfb/KeysymStr.h>
 
 #include "../../w0vncserver.h"
+#include "../../parameters.h"
 #include "Display.h"
 #include "Seat.h"
 #include "Keyboard.h"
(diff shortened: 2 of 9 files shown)


unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -117,6 +113,23 @@ void RemoteDesktop::notifyKeyboardKeysym(uint32_t keysym,
...
+  uint32_t state;
+
+  if (!(selectedDevices & DEV_KEYBOARD))
+    return;
+
+  state = down ? 1 : 0;
+
+  g_variant_builder_init(&optionsBuilder, G_VARIANT_TYPE("a{sv}"));
+  params = g_variant_new("(oa{sv}iu)", sessionHandle.c_str(),
+                         &optionsBuilder, xtcode, state);
```
CendioOssman: This needs to be mapped between RFB and what the portal expects.

unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
```diff
@@ -91,11 +91,15 @@ void VirtualKeyboard::key(uint32_t keysym,
...
-    return;
+  if (!rawKeyboard) {
+    keyCode = wKeyboard->keysymToKeycode(keysym);
+    if ((unsigned int)keyCode == XKB_KEYCODE_INVALID) {
+      vlog.debug("Unable to map keysym XK_%s (0x%04x), ignoring key press",
+                 KeySymName(keysym), keysym);
+      return;
+    }
+  } else {
+    keyCode = xtcode;
```
CendioOssman: Same thing here, also needs to be mapped.

CendioOssman (changes requested):

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: The compositors expect evdev codes. I just copied `qnum_to_xorgevdev.c` and used that for mapping. Maybe it should be renamed to `qnum_to_evdev.c` or something?

CendioOssman: It might be "xorg" specific in that it has that odd 8 offset.

CendioOssman: And nothing in the spec says that is is supposed to be evdev codes as far as I can tell.

If that happens to be the de facto-standard, then we can play along. But we should have a comment that this is not properly standardised, and a reference to a bug report to the specification that this is lacking.

(similarily, I don't see anything properly defining keysyms either)

CendioHalim: https://github.com/flatpak/xdg-desktop-portal/issues/1834

CendioHalim: I've added a FIXME explaining that the portals expect evdev keycodes and xkbcommon keysyms

unix/w0vncserver/wayland/WaylandDesktop.cxx
```diff
@@ -129,6 +133,14 @@ void WaylandDesktop::pointerEvent(const core::Point& pos, uint16_t buttonMask)
 
 void WaylandDesktop::keyEvent(uint32_t keysym, uint32_t keycode, bool down)
 {
+  if (rawKeyboard) {
+    if (keycode >= code_map_qnum_to_xorgevdev_len) {
+      vlog.error("Could not map key event to evdev key code");
+      return;
+    }
+    keycode = code_map_qnum_to_xorgevdev[keycode];
```
CendioOssman: I don't think this is correct. You have a layout here, so it's not a given that evdev key codes are used. You should probably do what vncviewer does to map things.

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -117,6 +113,30 @@ void RemoteDesktop::notifyKeyboardKeysym(uint32_t keysym,
   remoteDesktop->call("NotifyKeyboardKeysym", params);
 }
 
+void RemoteDesktop::notifyKeyboardKeycode(uint32_t xtcode, bool down)
```
CendioOssman: These do not use xt key codes, so I think this argument name can be confusing.

unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
```diff
@@ -76,8 +77,7 @@ unsigned int VirtualKeyboard::getLEDState()
   return seat->getKeyboard()->getLEDState();
 }
 
-void VirtualKeyboard::key(uint32_t keysym,
-                          uint32_t /* xtcode */, bool down)
+void VirtualKeyboard::key(uint32_t keysym, uint32_t xtcode, bool down)
```
CendioOssman: Same thing here, these are not xt key codes.

unix/w0vncserver/wayland/WaylandDesktop.cxx
CendioHalim: Right, I've changed it so that a new keycodemap is generated whenever the keymap is changed.

Also updated all other keycodemaps we have, might as well if we're adding a new map.

unix/w0vncserver/wayland/objects/Keyboard.cxx
```diff
@@ -233,3 +250,43 @@ unsigned int Keyboard::getLEDState()
...
+    xkb_keycode_t keycode;
+    bool match;
+    const char* qnumKeyName;
+
+    match = false;
+    qnumKeyName = code_map_qnum_to_xkb[i];
+    if (!qnumKeyName)
+      continue;
+
+    keycode = 0;
```
CendioOssman: This line does nothing.

unix/w0vncserver/wayland/objects/Keyboard.cxx
```diff
@@ -233,3 +250,43 @@ unsigned int Keyboard::getLEDState()
...
+
+  for (unsigned int i = 0; i < code_map_qnum_to_xkb_len; i++) {
+    xkb_keycode_t keycode;
+    bool match;
+    const char* qnumKeyName;
+
+    match = false;
+    qnumKeyName = code_map_qnum_to_xkb[i];
+    if (!qnumKeyName)
+      continue;
```
CendioOssman: Can this really be a case? Isn't an `assert()` more appropriate?

unix/w0vncserver/wayland/objects/Keyboard.h
```diff
@@ -65,6 +68,7 @@ namespace wayland {
     XkbContext* context;
     // Called when the LED state has changed
     std::function<void(unsigned int)> setLEDstate;
+    int code_map_qnum_to_keycode[256];
```
CendioOssman: Not `uint32_t`?

unix/w0vncserver/wayland/objects/Keyboard.cxx
```diff
@@ -233,3 +251,50 @@ unsigned int Keyboard::getLEDState()
...
+        continue;
+
+      xkbKeyCode = xkb_keymap_key_by_name(context->keymap, keyName);
+      if (xkbKeyCode != XKB_KEYCODE_INVALID)
+        break;
+    }
+
+    if (xkbKeyCode != 0) {
+      vlog.debug("Found mapping for key %.4s: 0x%04x -> 0x%04x",
+                 keyName, rfbcode, xkbKeyCode);
```
CendioOssman: We probably don't want this massive dump on each debug startup. The failures should be sufficient.

unix/w0vncserver/wayland/objects/Keyboard.cxx
```diff
@@ -233,3 +251,50 @@ unsigned int Keyboard::getLEDState()
...
+      codeMapQnumToKeyCode[rfbcode] = xkbKeyCode;
+    } else {
+      vlog.debug("No mapping found for key 0x%04x", rfbcode);
```
CendioOssman, line 302: Do we get any false complaints from this? I suspect some QEMU codes won't be present at all in `code_map_xkb_to_qnum`? We probably only want to complain when there is some issue with the local layout. I.e. when none of the names are found.

unix/w0vncserver/wayland/objects/Keyboard.cxx
```diff
@@ -233,3 +251,50 @@ unsigned int Keyboard::getLEDState()
...
+
+    xkbKeyCode = 0;
+    keyName = "";
+    for (unsigned int j = 0; j < code_map_xkb_to_qnum_len; j++) {
+      if (code_map_xkb_to_qnum[j].to != rfbcode)
+        continue;
+
+      keyName = code_map_xkb_to_qnum[j].from;
+      if (!keyName)
+        continue;
```
CendioOssman: Shouldn't this be an `assert()`?

unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
```diff
@@ -91,11 +91,15 @@ void VirtualKeyboard::key(uint32_t keysym,
...
-    return;
+  if (!rawKeyboard) {
+    key = wKeyboard->keysymToKeycode(keysym);
+    if ((unsigned int)key == XKB_KEYCODE_INVALID) {
+      vlog.debug("Unable to map keysym XK_%s (0x%04x), ignoring key press",
+                 KeySymName(keysym), keysym);
+      return;
+    }
+  } else {
+    key = wKeyboard->rfbcodeToKeycode(keycode);
```
CendioOssman, line 102: Need a check here as well if there was no mapping.

unix/w0vncserver/wayland/objects/Keyboard.cxx
CendioHalim: It avoids this warning:

`/tigervnc/unix/w0vncserver/wayland/objects/Keyboard.cxx:291:5: error: ‘xkbKeyCode’ may be used uninitialized [-Werror=maybe-uninitialized]
`

unix/w0vncserver/wayland/objects/Keyboard.cxx
CendioHalim: Sure, I'll add an assert.

unix/w0vncserver/wayland/objects/Keyboard.cxx
CendioHalim: Okay.

unix/w0vncserver/wayland/objects/Keyboard.cxx
CendioHalim: What you're saying is that we only want to complain if we find nothing at all?

unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
CendioHalim: Fixed now.

unix/w0vncserver/wayland/objects/Keyboard.cxx
CendioOssman: No, it could be useful to complain about each problem. The issue is that this code will likely complain about two things:

 * Missing key names from the compositor's layout
 * Missing QEMU entries in `code_map_xkb_to_qnum`

The latter is our fault, and static, so it is just noise to complain about that.

PR #2001, opened 2025-10-10
Detect dropped frames in PipeWirePixelBuffer

CendioHalim:
If we don't process the events coming from PipeWire quick enough, we
might skip processing some frames. The effect we saw from this was that
the damage tracking would be incorrect, and the client would get sent
invalid frames.

PipeWire sends us a sequence with each buffer which we can use to detect
skipped/dropped events. When that happens, we can no longer rely on our
damage tracking and have to mark the entire display as dirty.

### DIFF
--- a/unix/w0vncserver/pipewire/PipeWirePixelBuffer.cxx
+++ b/unix/w0vncserver/pipewire/PipeWirePixelBuffer.cxx
@@ -189,7 +189,6 @@ void PipeWirePixelBuffer::processCursor(spa_buffer* buffer)
 void PipeWirePixelBuffer::processDamage(spa_buffer* buffer)
 {
   spa_meta* damage;
-  core::Region damagedRegion;
   spa_meta_region* metaRegion;
 
   damage = spa_buffer_find_meta(buffer, SPA_META_VideoDamage);
@@ -198,14 +197,15 @@ void PipeWirePixelBuffer::processDamage(spa_buffer* buffer)
     return;
 
   spa_meta_for_each(metaRegion, damage) {
+    if (!spa_meta_region_is_valid(metaRegion))
+      continue;
+
     core::Point tl{metaRegion->region.position.x,
                    metaRegion->region.position.y};
     core::Point br{static_cast<int>(tl.x + metaRegion->region.size.width),
                    static_cast<int>(tl.y + metaRegion->region.size.height)};
-    damagedRegion.assign_union({{tl, br}});
+    accumulatedDamage.assign_union({{tl, br}});
   }
-
-  accumulatedDamage.assign_union(damagedRegion);
 }
 
 bool PipeWirePixelBuffer::hasCursorData(spa_buffer* buffer)


unix/w0vncserver/pipewire/PipeWirePixelBuffer.cxx
```diff
@@ -188,7 +188,6 @@ void PipeWirePixelBuffer::processCursor(spa_buffer* buffer)
 void PipeWirePixelBuffer::processDamage(spa_buffer* buffer)
 {
   spa_meta* damage;
-  core::Region damagedRegion;
```
CendioOssman, line 179: This might be an optimization depending how well behaved the compositor is. But we can revisit that when someone has done some profiling.

unix/w0vncserver/pipewire/PipeWirePixelBuffer.cxx
```diff
@@ -106,7 +109,16 @@ void PipeWirePixelBuffer::processFrame(spa_buffer* buffer)
     return;
   }
 
-  rect = accumulatedDamage.get_bounding_rect();
+  header = (spa_meta_header*)spa_buffer_find_meta_data(buffer,
+                                                       SPA_META_Header,
+                                                       sizeof(*header));
+
+  // Detect dropped frames. We can't rely on the damage events we've
+  // gotten and have to mark the entire frame as damaged
```
CendioOssman, line 117: Sounds like a bug in the protocol or the implementations. A bug reference would be nice here so we can keep an eye on things. Since it has a big performance hit, it's something we want resolved.

CendioHalim: https://bugs.kde.org/show_bug.cgi?id=510561

unix/w0vncserver/pipewire/PipeWirePixelBuffer.cxx
CendioHalim: Not sure what you mean?

unix/w0vncserver/pipewire/PipeWirePixelBuffer.cxx
CendioOssman: Could you put the link in the comment? It'll get lost if it's just here in the PR.

unix/w0vncserver/pipewire/PipeWirePixelBuffer.cxx
CendioOssman: Adding rects to a region causes it to go through all existing rects and see if it can merge things. The cost of this depends on much is already in the region.

Which means there could be a performance difference between putting stuff in to the main region directly, compared to first merging all the rects in this specific frame first.

But either way is fine at this point as we don't have any numbers on which is better. Or if it even matters.

unix/w0vncserver/pipewire/PipeWirePixelBuffer.cxx
CendioHalim: Sure, added the link now

PR #2010, opened 2025-11-04
Fix WaylandClient find module

CendioHalim:
CMake will complain if the argument passed to
find_package_handle_standard_args doesn't match the argument passed to the find module. In this case, "Wayland" didn't match "WaylandClient".

### DIFF
--- a/cmake/Modules/FindWaylandClient.cmake
+++ b/cmake/Modules/FindWaylandClient.cmake
@@ -1,13 +1,13 @@
 #[=======================================================================[.rst:
-FindWayland
-----------
+FindWaylandClient
+-----------------
 Find the Wayland client library
 Result variables
 ^^^^^^^^^^^^^^^^
 This module will set the following variables if found:
-``WAYLAND_CLIENT_INCLUDE_DIRS``
+``WAYLANDCLIENT_INCLUDE_DIRS``
   where to find wayland-client.h.h, etc.
-``WAYLAND_CLIENT_LIBRARIES``
+``WAYLANDCLIENT_LIBRARIES``
   the libraries to link against to use wayland client.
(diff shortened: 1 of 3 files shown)


cmake/Modules/FindWaylandClient.cmake
```diff
@@ -1,5 +1,5 @@
 #[=======================================================================[.rst:
-FindWayland
+FindWaylandClient
 ----------
```
CendioOssman: This line should be extended to match

cmake/Modules/FindWaylandClient.cmake
```diff
@@ -31,20 +31,20 @@ find_path(Wayland_client_INCLUDE_DIR NAMES wayland-client.h
 )
 mark_as_advanced(Wayland_client_INCLUDE_DIR)
 
-find_library(Wayland_LIBRARIES NAMES wayland-client
+find_library(Wayland_client_LIBRARIES NAMES wayland-client
```
CendioOssman: Why not `WaylandClient_LIBRARIES` to keep the naming consistent?

CendioOssman (changes requested):

CendioOssman:
> I'm not sure how to best handle the inconsistency with `WaylandClient_FOUND` / `WAYLANDCLIENT_FOUND` and `WAYLAND_CLIENT_(INCLUDE_DIRS|LIBRARIES)`. It would be nice if the variables were both prefixed `WAYLAND_CLIENT`

Or just accept how CMake likes to name things and call them `WAYLANDCLIENT_(INCLUDE_DIRS|LIBRARIES)`.

CendioOssman (approved):

PR #2012, opened 2025-11-12
PortalDesktop: Add clipboard support

CendioHalim:
This PR adds clipboard support to PortalDesktop.

Tested with GNOME, seems to work well.

I could not get this working at all on KDE (`clipboard_enabled` always returns 0). Will leave this as a draft until we get further insights.

### DIFF
--- a/unix/w0vncserver/portals/Clipboard.cxx
+++ b/unix/w0vncserver/portals/Clipboard.cxx
@@ -1,0 +1,552 @@
+/* Copyright 2025 Adam Halim for Cendio AB
+ *
+ * This is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this software; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ * USA.
+ */
+
+#include <assert.h>
+#include <sys/stat.h>
+
+#include <gio/gio.h>
+#include <gio/gunixfdlist.h>
+#include <gio-unix-2.0/gio/gunixinputstream.h>
+#include <glib-object.h>
+#include <glib.h>
+
+#include <core/LogWriter.h>
+#include <core/string.h>
+
+#include "../w0vncserver.h"
+#include "PortalProxy.h"
+#include "Clipboard.h"
+
+static core::LogWriter vlog("Clipboard");
+
+const char* MIME_TEXT_PLAIN = "text/plain";
+const char* MIME_TEXT_PLAIN_UTF8 = "text/plain;charset=utf-8";
+
+const char* MIME_TYPES[] {
+  MIME_TEXT_PLAIN_UTF8,
+  MIME_TEXT_PLAIN,
+};
+
+struct PendingData {
+  uint32_t serial;
+  std::string mimeType;
+};
+
+Clipboard::Clipboard(std::string sessionHandle_,
+                     std::function<void(const char* data)>
+                       sendClipboardDataCb_,
+                     std::function<void(bool available)>
+                        clipboardAnnounceCb_,
+                     std::function<void()> clipboardRequestCb_)
+  : clipboard(nullptr), sessionHandle(sessionHandle_),
+    selectionOwner(true), readInProgress(false),
+    pendingReadMimeType(nullptr), readStream(nullptr),
+    readCancellable(nullptr),
+    sendClipboardDataCb(sendClipboardDataCb_),
+    clipboardAnnounceCb(clipboardAnnounceCb_),
+    clipboardRequestCb(clipboardRequestCb_)
+{
+  assert(available());
+
+  clipboard = new PortalProxy("org.freedesktop.portal.Desktop",
+                              "/org/freedesktop/portal/desktop",
+                              "org.freedesktop.portal.Clipboard");
+}
+
+Clipboard::~Clipboard()
+{
+  if (clipboard)
+    clearPendingSerials();
+
+  delete clipboard;
+
+  if (readStream)
+    g_object_unref(readStream);
+  // FIXME: Do we cancel a pending read here?
+  if (readCancellable)
+    g_object_unref(readCancellable);
+}
+
+bool Clipboard::available()
+{
+  return PortalProxy::interfacesAvailable({"org.freedesktop.portal.Clipboard"});
+}
+
+void Clipboard::subscribe()
+{
+  clipboard->subscribe("SelectionTransfer",
+                       std::bind(&Clipboard::handleSelectionTransfer,
+                                 this, std::placeholders::_1));
+  clipboard->subscribe("SelectionOwnerChanged",
+                       std::bind(&Clipboard::handleSelectionOwnerChanged,
+                                 this, std::placeholders::_1));
+}
+
+void Clipboard::selectionWrite(const char* data)
+{
+  while (!pendingSerials.empty()) {
+    PendingData pending;
+
+    pending = pendingSerials.front();
+    pendingSerials.pop();
+    selectionWrite(pending, data);
+  }
+}
+
+void Clipboard::requestClipboard()
+{
+  GVariant* params;
+  GVariantBuilder optionsBuilder;
+
+  g_variant_builder_init(&optionsBuilder, G_VARIANT_TYPE_VARDICT);
+  params = g_variant_new("(oa{sv})", sessionHandle.c_str(), &optionsBuilder);
+
+  clipboard->call("RequestClipboard", params);
+}
+
+void Clipboard::setSelection()
+{
+  GVariantBuilder optionsBuilder;
+  GVariant* params;
+  GVariant* mimeTypes;
+
+  mimeTypes = g_variant_new_strv(MIME_TYPES, sizeof(MIME_TYPES) /
+                                             sizeof(MIME_TYPES[0]));
+
+  g_variant_builder_init(&optionsBuilder, G_VARIANT_TYPE_VARDICT);
+  g_variant_builder_add(&optionsBuilder, "{sv}", "mime_types", mimeTypes);
+  params = g_variant_new("(oa{sv})", sessionHandle.c_str(),
+                         &optionsBuilder);
+
+  try  {
+    clipboard->call("SetSelection", params);
+  } catch (std::exception& e) {
+    vlog.error("Could not set selection: %s", e.what());
+  }
+}
+
+void Clipboard::selectionWrite(PendingData pendingData, const char* data)
+{
+  GError* error = nullptr;
+  GUnixFDList* fdList = nullptr;
+  GVariant* params;
+  GVariant* response;
+  int32_t fdIndex;
+  int fd;
+  size_t remaining;
+  ssize_t written;
+  const char* buf;
+  std::string ascii;
+  uint32_t serial;
+  const char* mimeType;
+
+  serial = pendingData.serial;
+  mimeType = pendingData.mimeType.c_str();
+
+  params = g_variant_new("(ou)", sessionHandle.c_str(), serial);
+
+  response = g_dbus_proxy_call_with_unix_fd_list_sync(
+               clipboard->getProxy(), "SelectionWrite", params,
+               G_DBUS_CALL_FLAGS_NONE, 3000, nullptr, &fdList, nullptr,
+               &error);
+
+  if (error) {
+    vlog.error("Error writing to clipboard: %s", error->message);
+    g_variant_unref(response);
+    g_error_free(error);
+    selectionWriteDone(serial, false);
+    return;
+  }
+
+  g_variant_get(response, "(h)", &fdIndex);
+  g_variant_unref(response);
+
+  // FIXME: Can we always assume index 0?
+  fd = g_unix_fd_list_get(fdList, 0, &error);
+  if (error) {
+    g_object_unref(fdList);
+    vlog.error("Error writing to clipboard: %s", error->message);
+    g_error_free(error);
+    selectionWriteDone(serial, false);
+    return;
+  }
+
+  if (strcmp(mimeType, MIME_TEXT_PLAIN) == 0) {
+    ascii = core::utf8ToAscii(data, strlen(data));
+    buf = ascii.c_str();
+    remaining = ascii.size();
+  } else if (strcmp(mimeType, MIME_TEXT_PLAIN_UTF8) == 0) {
+    buf = data;
+    remaining = strlen(data);
+  } else {
+    vlog.error("Error writing to clipboard: unsupported mime type: '%s'", mimeType);
+    selectionWriteDone(serial, false);
+    return;
+  }
+
+  while (remaining > 0) {
+    written = write(fd, buf, remaining);
+
+    if (written < 0) {
+      vlog.error("Error writing to clipboard: %s", strerror(errno));
+      selectionWriteDone(serial, false);
+      g_object_unref(fdList);
+      close(fd);
+      return;
+    }
+
+    remaining -= written;
+    buf += written;
+  }
+
+  if (close(fd) != 0) {
+    vlog.error("Error writing to clipboard: %s", strerror(errno));
+    selectionWriteDone(serial, false);
+  } else {
+    selectionWriteDone(serial, true);
+  }
+
+  g_object_unref(fdList);
+}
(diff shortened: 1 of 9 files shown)


common/core/string.cxx
```diff
@@ -606,6 +606,20 @@ namespace core {
     return true;
   }
 
+  bool isValidAscii(const char* str, size_t bytes)
+  {
+    if (str == nullptr || bytes == 0)
+      return false;
```
CendioOssman: Shouldn't an empty string be valid?

common/core/string.cxx
```diff
@@ -606,6 +606,20 @@ namespace core {
...
+
+    while(*str != '\0' && bytes-- > 0) {
```
CendioOssman: I would suggest following the style of the other helpers, to make things more readable.

unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -0,0 +1,486 @@
...
+void Clipboard::selectionWrite(uint32_t serial)
+{
+  GError* error = nullptr;
+  GVariant* params;
+
+  params = g_variant_new("(ou)", sessionHandle.c_str(), serial);
+
+  pendingSerials.push(serial);
+
+  g_dbus_proxy_call_with_unix_fd_list(
```
CendioOssman: Can these be called "sync"? Otherwise we have a crashy race here if the clipboard object is deleted before the callback is called.

unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -0,0 +1,486 @@
...
+  uint32_t serial;
+  char* sessionHandle_; // ignored
+  char* mimeType;
+
+  g_variant_get(parameters, "(osu)", &sessionHandle_, &mimeType, &serial);
+
+  // Always write UTF-8
+  if (strcmp(mimeType, MIME_TEXT_PLAIN_UTF8) != 0) {
+    vlog.error("Unsupported mime type %s", mimeType);
+    return;
```
CendioOssman: No error handling here to inform the other side they're doing invalid things?

unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -0,0 +1,486 @@
...
+  data = (const char*)g_bytes_get_data(bytes, &dataSize);
+
+  if (strcmp(pendingReadMimeType, MIME_TEXT_PLAIN) == 0) {
+    if (!core::isValidAscii(data, dataSize)) {
+      vlog.error("Invalid ASCII sequence in clipboard - ignoring");
+      g_cancellable_cancel(readCancellable);
+      return;
+    }
+  } else if (strcmp(pendingReadMimeType, MIME_TEXT_PLAIN_UTF8) == 0) {
+    if (!core::isValidUTF8(data, dataSize)) {
```
CendioOssman: You can't call this on partial data. You'll get false negatives.

unix/w0vncserver/portals/PortalDesktop.cxx
```diff
@@ -140,6 +150,22 @@ void PortalDesktop::pointerEvent(const core::Point& pos,
...
+void PortalDesktop::handleClipboardAnnounce(bool available)
+{
+  if (available)
+    server->requestClipboard();
+}
+
+void PortalDesktop::handleClipboardData(const char* data)
+{
+  if (data) remoteDesktop->setSelection(data);
+}
```
CendioOssman: This should ideally be more async and not request data until the compositor does.

CendioOssman (changes requested):

CendioOssman:
> I could not get this working at all on KDE (`clipboard_enabled` always returns 0). Will leave this as a draft until we get further insights.

Do you have a bug report with KDE for this?

CendioHalim:
https://bugs.kde.org/show_bug.cgi?id=512067

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: There is, we can call `SelectionWriteDone(serial, false)`

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: Oh right, I'll move this check to when the transfer is complete

CendioHalim:
* https://bugs.kde.org/show_bug.cgi?id=512075
* https://bugs.kde.org/show_bug.cgi?id=512076

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: Sure thing

unix/w0vncserver/portals/PortalDesktop.cxx
CendioHalim: I've re-written this to be async now

common/core/string.cxx
```diff
@@ -572,6 +572,53 @@ namespace core {
     return out;
   }
 
+  std::string utf8ToAscii(const char* src, size_t bytes) {
```
CendioOssman: These functions have unit tests. Please add a few for these new functions as well.

common/core/string.cxx
```diff
@@ -572,6 +572,53 @@ namespace core {
...
+        in_len -= 3;
+      } else if (ascii >= 0xf0 && ascii <= 0xf4) {
+        if (in_len < 4)
+          break;
+        in += 4;
+        in_len -= 4;
+      } else {
+        // Malformed utf-8
+        assert(false);
+      }
```
CendioOssman: We already have handling for this. Reuse that.

And related, make sure your new functions follow the style of the existing ones.

unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -0,0 +1,526 @@
...
+{
+  clipboard->subscribe("SelectionTransfer",
+                       std::bind(&Clipboard::handleSelectionTransfer,
+                                 this, std::placeholders::_1));
+  clipboard->subscribe("SelectionOwnerChanged",
+                       std::bind(&Clipboard::handleSelectionOwnerChanged,
+                                 this, std::placeholders::_1));
+}
+
+void Clipboard::handleClipboardData(const char* data)
```
CendioOssman: Don't you want to match the name with the portal name?

unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -0,0 +1,526 @@
...
+
+  params = g_variant_new("(ou)", sessionHandle.c_str(), serial);
+
+  response = g_dbus_proxy_call_with_unix_fd_list_sync(
+               clipboard->getProxy(), "SelectionWrite", params,
+               G_DBUS_CALL_FLAGS_NONE, 3000, nullptr, &fdList, nullptr,
+               &error);
+
+  if (error) {
+    vlog.error("SelectionWrite() failed: %s", error->message);
```
CendioOssman: Non-debug log messages should ideally be written with more casual users in mind.

unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -0,0 +1,526 @@
...
+  mimeType = "";
+  for (uint i = 0; i < sizeof(MIME_TYPES) / sizeof(MIME_TYPES[0]); i++) {
+    if (availableMimeTypes.find(MIME_TYPES[i]) != availableMimeTypes.end()) {
+      mimeType = MIME_TYPES[i];
+      break;
+    }
+  }
+
+  if (!mimeType) {
+    vlog.error("Could not find a supported available mime type, skipping selectionRead()");
```
CendioOssman: Won't this be somewhat common? E.g. copying images? Should not log this as an error in that case.

unix/w0vncserver/portals/PortalDesktop.cxx
```diff
@@ -140,6 +175,23 @@ void PortalDesktop::pointerEvent(const core::Point& pos,
...
+void PortalDesktop::handleClipboardRequest()
+{
+  if (clipboard)
+    clipboard->selectionRead();
+}
+
+void PortalDesktop::handleClipboardAnnounce(bool available)
+{
+  if (available)
+    clipboard->setSelection();
```
CendioOssman, line 202: It should handle if `available` is `false` as well. Otherwise the other end will incorrectly think we still have clipboard data to share.

unix/w0vncserver/portals/PortalDesktop.cxx
```diff
@@ -140,6 +175,23 @@ void PortalDesktop::pointerEvent(const core::Point& pos,
...
 
+void PortalDesktop::handleClipboardRequest()
+{
+  if (clipboard)
+    clipboard->selectionRead();
+}
+
+void PortalDesktop::handleClipboardAnnounce(bool available)
+{
+  if (available)
```
CendioOssman, line 201: Don't you need `if (clipboard)` here as well?

unix/w0vncserver/w0vncserver-forget.man
```diff
@@ -6,12 +6,15 @@ w0vncserver-forget \- forget w0vncserver display choice
 .SH DESCRIPTION
 .B w0vncserver-forget
 is a utility for \fBw0vncserver\fP(1) which will forget the stored
-display choice configuration.
+display choice and clipboard configuration.
```
CendioOssman: Perhaps it's best to avoid details and simply state that it forgets all stored choices? It looks like the list could grow in the future.

unix/w0vncserver/w0vncserver-forget.man
```diff
@@ -6,12 +6,15 @@ w0vncserver-forget \- forget w0vncserver display choice
...
 
 .SH FILES
 .TP
 \fI$XDG_STATE_HOME/tigervnc/restoretoken\fP
 Default location of the restoretoken file.
+.TP
+\fI$XDG_STATE_HOME/tigervnc/clipboard\fP
```
CendioOssman: This is a very generic name. I wonder if we should start grouping things in some way. Either with a common naming scheme, or a directory.

unix/w0vncserver/portals/PortalDesktop.cxx
```diff
@@ -140,6 +175,34 @@ void PortalDesktop::pointerEvent(const core::Point& pos,
   remoteDesktop->notifyPointerMotionAbsolute(pos.x, pos.y, buttonMask);
 }
 
+void PortalDesktop::handleClipboardRequest()
```
CendioOssman, line 186: You need to update the handling and documentation of the clipboard settings as well. E.g. `AcceptCutText`.

CendioOssman (changes requested):

common/core/string.cxx
CendioHalim: You mean something like calling `utf8ToLatin1` and then masking all chars > 127 to `?` ?

unix/w0vncserver/portals/PortalDesktop.cxx
CendioHalim: Oops, yeah

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: Sure

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: Yeah good point, changing it to `vlog.info()`

CendioOssman: info is also logged by default, so that will still be a lot of noise.

common/core/string.cxx
CendioOssman: No, I mean following the same structure as `utf8ToLatin1()`. I.e. build it around calls to `utf8ToUCS4()`.

common/core/string.cxx
CendioHalim: Added unit tests now

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: Alright, nothing is logged now

unix/w0vncserver/w0vncserver-forget.man
CendioHalim: Stopped storing the clipboard preference and instead just request clipboard access each time.

unix/w0vncserver/portals/PortalDesktop.cxx
CendioHalim: Right, I've added a method `Clipboard::clearSelection()` which should make us lose clipboard ownership.

common/core/string.cxx
```diff
@@ -572,6 +572,51 @@ namespace core {
     return out;
   }
 
+  std::string utf8ToAscii(const char* src, size_t bytes) {
```
CendioOssman: `{` on a new line for functions.

common/core/string.cxx
```diff
@@ -606,6 +651,17 @@ namespace core {
     return true;
   }
 
+  bool isValidAscii(const char* str, size_t bytes)
+  {
+    while ((bytes > 0) && (*str != '\0')) {
+      if (!isascii(*str))
+        return false;
+     str++;
```
CendioOssman: Indentation looks off here.

unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -0,0 +1,560 @@
...
+      close(fd);
+      return;
+    }
+
+    remaining -= written;
+    buf += written;
+  }
+
+  if (close(fd) != 0) {
+    vlog.error("Error closing clipboard fd: %s", strerror(errno));
```
CendioOssman: Errors from `close()` can be delayed errors from previous writes. So I would suggest using the same generic error message you have elsewhere.

unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -0,0 +1,560 @@
...
+
+  g_variant_builder_init(&optionsBuilder, G_VARIANT_TYPE_VARDICT);
+  g_variant_builder_add(&optionsBuilder, "{sv}", "mime_types", mimeTypes);
+  params = g_variant_new("(oa{sv})", sessionHandle.c_str(),
+                         &optionsBuilder);
+
+  error = clipboard->call("SetSelection", params);
+
+  if (error) {
+    vlog.error("%s", error->message);
```
CendioOssman: Don't we want some context in this error message?

unix/w0vncserver/portals/PortalDesktop.cxx
```diff
@@ -76,7 +78,47 @@ void PortalDesktop::start()
...
+
+    assert(remoteDesktop);
+
+    clipboard = new Clipboard(remoteDesktop->getSessionHandle(),
+                              sendClipboardData, clipboardAnnounceCb,
+                              clipboardRequestCb);
+    clipboardAccess = true;
+
+    error = clipboard->requestClipboard();
+    if (error) {
```
CendioOssman: Isn't a `bool` sufficient here if you're not interested in the error message?

unix/w0vncserver/portals/PortalProxy.h
```diff
@@ -38,11 +40,19 @@ class PortalProxy {
   // signalCallback() when a response signal is received. A
   // requestHandleToken must be set to link the response signal.
   // signalCallback can be set to nullptr is no response is expected.
-  void call(const char* method, GVariant* parameters,
-            const char* requestHandleToken = nullptr,
-            std::function<void(GVariant* parameters)>
-              signalCallback = nullptr,
-            GDBusCallFlags flags = G_DBUS_CALL_FLAGS_NONE);
+  // Caller is responsible for freeing GError
```
CendioOssman: If you want to be fancy, then you can set an attribute on the method to make the compiler throw an error if a caller ignores the return value.

CendioOssman (changes requested):

unix/w0vncserver/w0vncserver.man
```diff
@@ -196,6 +206,19 @@ layout configured on the server instead of the layout configured on the
 client. Default is off.
 .
 .TP
+.B \-AcceptCutText
+Accept clipboard updates from clients. Default is on.
+.
+.TP
+.B \-MaxCutText \fIbytes\fP
```
CendioOssman: These should be alphabetically ordered

CendioOssman (changes requested):

unix/w0vncserver/portals/PortalProxy.h
CendioHalim: Changed `call()` to throw an exception instead

unix/w0vncserver/w0vncserver.man
CendioHalim: Fixed

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: Fixed

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: Changed most messages to be more generic/similar where possible.

PR #2013, opened 2025-11-13
Fix configuration when ENABLE_WAYLAND=0

ceamac:
CMakeLists.txt assumes ENABLE_WAYLAND is either AUTO or ON.
If ENABLE_WAYLAND=OFF it still requires all wayland dependencies and fails if not found.

Fix by switching `else` to `elseif`.  I also moved the big if inside the `AUTO` branch.


ceamac:
Sure!  I'll update the PR later today

PR #2021, opened 2025-11-24
Support VK_PACKET in vncviewer on Windows

svenssonaxel:


svenssonaxel:
Squashed.
@CendioOssman ping

vncviewer/KeyboardWin32.cxx
```diff
@@ -241,7 +242,7 @@ bool KeyboardWin32::handleEvent(const void* event)
 
     // Windows sets the scan code to 0x00 for multimedia keys, so we
     // have to do a reverse lookup based on the vKey.
-    if (systemKeyCode == 0x00) {
+    if (systemKeyCode == 0x00 && vKey != VK_PACKET) {
```
CendioOssman: A system key code is crucial, as that is now how we track things. Do we not get anything useful when there is a `VK_PACKET`?

If so, then we'll have to fake something. But I guess we won't handle overlapping `VK_PACKET`? So a single static fake scan code should be sufficient?

vncviewer/KeyboardWin32.cxx
```diff
@@ -273,6 +273,45 @@ bool KeyboardWin32::handleEvent(const void* event)
...
+          vlog.error(_("Unmatched UTF-16 surrogate pair through VK_PACKET, "
+                       "code: 0x%04x"),
+                     ucsCode);
+          return true;
+        }
+        uint32_t codePoint = (((vkPacketHighSurrogate & 0x03ff) << 10) |
+                              (ucsCode & 0x03ff)) + 0x010000;
+        vkPacketHighSurrogate = 0;
+        keySym = ucs2keysym(codePoint);
+        keyCode = 0x01000000 | codePoint; // Fake key code
```
CendioOssman: This should be `0` rather than faked under the new model.

vncviewer/KeyboardWin32.cxx
```diff
@@ -273,6 +273,45 @@ bool KeyboardWin32::handleEvent(const void* event)
...
+        handler->handleKeyPress(systemKeyCode, keyCode, keySym);
```
CendioOssman: I don't see any handling of key release, and I'm not sure we can trust `VK_PACKET` with that anyway, so shouldn't there be a call to `handleKeyRelease()` here as well?

CendioOssman (changes requested):
Looks really good! There are some minor tweaks and we can get this merged.

vncviewer/KeyboardWin32.cxx
svenssonaxel: Faked as `0x1ff`. It's a little ugly and requires changing `if (systemKeyCode & ~0x7f) {` -> `if (systemKeyCode != SCAN_VK_PACKET && (systemKeyCode & ~0x7f)) {` but I don't know what fake value <128 would be safe.

vncviewer/KeyboardWin32.cxx
svenssonaxel: Fixed in `KeyboardWin32::translateSystemKeyCode`

vncviewer/KeyboardWin32.cxx
svenssonaxel: Release is handled via the normal path and required no special code when I let `systemKeyCode == 0`. Now that you want a fake value, I set it in `WM_KEYUP` branch as well. The key up events seem to arrive consistently right after the key down events.

vncviewer/KeyboardWin32.cxx
svenssonaxel: Overlapping VK_PACKET shouldn't exist, as the system issues key release immediately after key down. But, your idea of issuing a key down+release in response to a key down, rules out this possibility even in theory.

vncviewer/KeyboardWin32.cxx
svenssonaxel: Changed my mind and implemented your idea, it's a more defensive style.

svenssonaxel:
@CendioOssman ping

CendioOssman:
I'm unfortunately days away from my summer holiday. I'll see if I can find some time to test and merge this, but there is a risk it will have to wait until I'm back.

vncviewer/KeyboardWin32.cxx
```diff
@@ -239,6 +243,11 @@ bool KeyboardWin32::handleEvent(const void* event)
       return true;
     }
 
+    if (vKey == VK_PACKET) {
+      assert(systemKeyCode == 0);
```
CendioOssman: An assert might be excessive since this is out of our control. Let's just ignore it.

vncviewer/KeyboardWin32.cxx
```diff
@@ -586,6 +644,11 @@ uint32_t KeyboardWin32::translateSystemKeyCode(int systemKeyCode)
   if (systemKeyCode == 0xc5)
     return 0x45;
 
+  // VK_PACKET only has a fake key code, and should be sent to server as 0.
+  if (systemKeyCode == SCAN_VK_PACKET) {
```
CendioOssman: No `{}` for single statements.

vncviewer/KeyboardWin32.cxx
CendioOssman, line 268: Does this need to be excluded as well, for robustness?

vncviewer/KeyboardWin32.cxx
```diff
@@ -273,6 +282,45 @@ bool KeyboardWin32::handleEvent(const void* event)
...
+        // the low surrogate which should come immediately after.
+        if (vkPacketHighSurrogate) {
+          // High surrogate followed by another high surrogate
+          vlog.error(_("Unmatched UTF-16 surrogate pair through VK_PACKET, "
+                       "codes: 0x%04x, 0x%04x"),
+                     vkPacketHighSurrogate, ucsCode);
+        }
+        vkPacketHighSurrogate = ucsCode;
+        return true;
+      }
```
CendioOssman: Should not be a newline here

vncviewer/KeyboardWin32.cxx
```diff
@@ -273,6 +282,45 @@ bool KeyboardWin32::handleEvent(const void* event)
...
+      else {
+        assert((ucsCode & 0xfc00) == 0xdc00);
```
CendioOssman, line 302: I assume this comes from Windows? We should be prepared for bugs there and not just crash. A check and return is more appropriate.

vncviewer/KeyboardWin32.cxx
```diff
@@ -273,6 +282,45 @@ bool KeyboardWin32::handleEvent(const void* event)
...
+        // We have received a low surrogate code unit. We should have a high
+        // surrogate saved that we can use to calculate the code point.
+        if (!vkPacketHighSurrogate) {
+          // Low surrogate not directly preceded by a high surrogate
+          vlog.error(_("Unmatched UTF-16 surrogate pair through VK_PACKET, "
+                       "code: 0x%04x"),
```
CendioOssman: It's best if we can keep the strings identical to avoid burdening the translators.

vncviewer/KeyboardWin32.h
```diff
@@ -52,6 +52,7 @@ class KeyboardWin32 : public Keyboard
   HKL currentLayout;
 
   bool altGrArmed;
+  uint32_t vkPacketHighSurrogate;
```
CendioOssman: This is a confusing placement as it is not related to the AltGr variables above and below it.

I suggest giving it its own section below those two.

vncviewer/keysym2ucs.c
```diff
@@ -167,8 +167,18 @@ unsigned ucs2keysym(unsigned ucs)
...
-  if ((ucs & 0xff000000) == 0)
+  /* surrogates? */
+  if (0xd800 <= ucs && ucs <= 0xdfff)
+    return NoSymbol;
+
+  /* private use? */
+  if ((0xe000 <= ucs && ucs <= 0xf8ff) ||
+      (0x0f0000 <= ucs && ucs <= 0x0ffffd) ||
+      (0x100000 <= ucs && ucs <= 0x10fffd))
+    return NoSymbol;
```
CendioOssman: This is a nice addition, but not strictly related to VK_PACKET? Could you put it as a separate commit?

CendioOssman (changes requested):
Looks very nice. Some small stylistic fixes and then we're ready.

What good ways are there to test this?

svenssonaxel:
> What good ways are there to test this?

I don't know how "good" it is, but here's the way I test it manually:
- Connect from a Windows client, using vncviewer.exe to a VNC Server. I'm not sure how much the server matters here, but I run x11vnc + Xvfb on Debian.
- On the Client, install Keepass and add an entry with the following password:
  ```
  !"#$%&'()*+,-./09:;<=>?@AZ[\]^_`az{|}~A"B'C^D`E~F＋G𐀂H𐌼IαJωKᛟL
  ```
- On the server, open any text editor.
- Use Keypass's auto-type feature to type the password into the server.

Expected / with this fix:
- The password is reproduced exactly on the server.

Actual / without this fix:
- Some characters are missing or replaced with other characters.

vncviewer/KeyboardWin32.cxx
svenssonaxel: Fixed.

vncviewer/KeyboardWin32.cxx
svenssonaxel: Fixed

vncviewer/KeyboardWin32.cxx
svenssonaxel: I don't know enough to tell. Left as-is, which at least has been tested for over a year.

vncviewer/KeyboardWin32.cxx
svenssonaxel: Fixed

vncviewer/KeyboardWin32.cxx
svenssonaxel: No, this is a provable invariant. Added comments.

vncviewer/KeyboardWin32.cxx
svenssonaxel: Fixed

vncviewer/KeyboardWin32.h
svenssonaxel: Fixed

vncviewer/keysym2ucs.c
svenssonaxel: Moved to https://github.com/TigerVNC/tigervnc/pull/2142

svenssonaxel: Oh wait, separate commit, not PR. Fixed and force pushed.

CendioOssman (approved):
Looks great!

I managed to test this with Microsoft's Android RDP client. Without the patch, almost nothing on the keyboard works. With the patch, almost everything works. So a very nice improvement. :)

PR #2022, opened 2025-11-24
Fix handling of large solid tight rects

CendioOssman:
Fix for ancient bug in the server that violates the Tight protocol, and more leniency in the client to handle the already deployed servers that won't be upgraded for some time.


svenssonaxel:
> It's in `readRect()`:

Oh, now I get it. The check was only relaxed for fill type rectangles, for which this loop is never executed.

PR #2023, opened 2025-11-24
Fix too short java array

svenssonaxel:


CendioOssman:
Should this also be applied on the 1.16 branch, @bphinz?

bphinz:
Yes please

Issue #2031, opened 2025-12-15
'AskDisplayChoice Never' causing View only mode


axet:
**Describe the bug**

Using '-AskDisplayChoice Never' causing remote connection to become view only, even if you choice "Remote control" before.

**To Reproduce**

run 'w0vncserver -AskDisplayChoice Never' expect no questions asked and "Remote control" enabled.

**Client (please complete the following information):**
 - OS: Debian Trixie
 - VNC client: Remmina 1.4.39

**Server (please complete the following information):**
 - OS: Debian Trixie
 - VNC server: w0vncserver
 - VNC server version: 1.15.90+ (recent master)
 - Server was started using: 'w0vncserver -SecurityTypes None -localhost -AskDisplayChoice Never'

axet:
It starts to work. I guess it does remember last user choose. I tried to figure it out using 'strace' but found no configs.

axet:
No problem. I guess better call it "Previous" or "Last" instead of "Never" to prevent confusion. Linux always been about control and configs. And it feels like we loosing it.

axet:
"during the program's current session. Never would be prompted once between app restarts... My wording not great. Or rename it to "Persist". And instead Once would be better to have "Session"...

CendioHalim:
Forgive me, I'm not sure if I follow.

Just to clarify, do you experience the behaviour described in the PR where this option was added? 
* https://github.com/TigerVNC/tigervnc/pull/1998#issue-3486119819

So for example, if you chose `Never`, you're not supposed to see the dialogue again (after the first time) until you run `w0vncserver-forget`.

axet:
That is correct. As it described in the PR message. Not sure what is confusion about.

axet:
That is correct.

PR #2038, opened 2026-01-12
Rename display choice parameter in w0vncserver

CendioHalim:
We had a user report that the display choice parameter option "Never" was misleading : #2031.

This PR changes the parameter from `askDisplayChoice` to `saveDisplayChoice`, with the options `Yes`, `No`, and `Session`.

I am not fully content with the option name `Session`, but cannot think of something more fitting. Also, I am debating whether `Yes` should be the default instead of `Session`. What do you think?

### DIFF
--- a/unix/w0vncserver/portals/RemoteDesktop.cxx
+++ b/unix/w0vncserver/portals/RemoteDesktop.cxx
@@ -59,9 +59,9 @@ core::BoolParameter
               false);
 
 core::EnumParameter
-  askDisplayChoice("AskDisplayChoice",
-                   "Ask which displays to share when user connects (Always, Once, Never)",
-                   {"Always", "Once", "Never"}, "Once");
+  rememberDisplayChoice("RememberDisplayChoice",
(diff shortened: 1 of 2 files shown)


unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -59,9 +59,9 @@ core::BoolParameter
               false);
 
 core::EnumParameter
-  askDisplayChoice("AskDisplayChoice",
-                   "Ask which displays to share when user connects (Always, Once, Never)",
-                   {"Always", "Once", "Never"}, "Once");
+  rememberDisplayChoice("RememberDisplayChoice",
+                        "Remember display choice when user connects (Awlays, Never, Session)",
```
CendioOssman: Typo :)

CendioOssman (changes requested):

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: :)

PR #2041, opened 2026-01-15
w0vncserver: Improve error handling in Portal implementation

CendioHalim:
Better error handling and some changes to how errors are logged.

Removed all calls to `fatal_error()` in the Portal implementation except for one in `PipeWireStream`

### DIFF
--- a/unix/w0vncserver/portals/RemoteDesktop.cxx
+++ b/unix/w0vncserver/portals/RemoteDesktop.cxx
@@ -406,6 +406,12 @@ void RemoteDesktop::handleCreateSession(GVariant *parameters)
                                                 G_VARIANT_TYPE_STRING);
   g_variant_unref(result);
 
+  if (!sessionHandleVariant) {
+    vlog.error("Failed to create session: no session handle");
+    cancelStartCb("Failed to start remote desktop session");
+    return;
+  }
+
   sessionHandle = g_variant_get_string(sessionHandleVariant, nullptr);
   g_variant_unref(sessionHandleVariant);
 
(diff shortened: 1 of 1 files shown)


unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -451,15 +457,28 @@ void RemoteDesktop::handleStart(GVariant* parameters)
...
 
-  if (!parseStreams(streams_)) {
-    g_variant_unref(streams_);
+  if (!streams) {
+    vlog.error("Failed to start remote desktop session");
+    cancelStartCb("Failed to get streams");
+    return;
+  }
+
+  if (!parseStreams(streams)) {
```
CendioOssman, line 481: Don't you still need `g_variant_unref(streams);`?

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -546,7 +568,6 @@ bool RemoteDesktop::parseStreams(GVariant* streams)
 
   if (n_streams  < 1) {
     vlog.error("Could not find streams to parse");
-    cancelStartCb("Failed to start remote desktop session");
```
CendioOssman: This seems unrelated?

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: Yes, missed that one

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: Kind of, I figured it's a bit confusing to call `cancelStartCb()` once inside `parseStreams()`, and then once again directly after returning false. Makes more sense to just let the caller do it?

CendioOssman: Ah. Yes, I agree. It's better to have all error handling in the same layer.

Issue #2042, opened 2026-01-15
w0vncserver Segmentation Fault ARM64


DaanSelen:
**Describe the bug**
Using the w0vncserver on a raspberry pi

**To Reproduce**
Compile the w0vncserver on an arm64 device and then copy the binary to another arm64 device.

**Expected behavior**
For VNC to work normally on port 5900

**Screenshots**
n.a.

**Client (please complete the following information):**
 - OS: Linux (noVNC) JS client
 - VNC client: noVNC
 - VNC client version: unknown
 - Client downloaded from: Included in MeshCentral

**Server (please complete the following information):**
 - OS: Raspberry Pi OS
 - VNC server: TigerVNC
 - VNC server version: TigerVNC Server version 1.16.80
 - Server downloaded from: Self-compiled
 - Server was started using: `./w0vncserver -SecurityTypes=None`

**Additional context**
```
Thu Jan 15 14:44:58 2026
 main:        Listening for VNC connections on all interface(s), port 5900

Thu Jan 15 14:45:04 2026
 Connections: Accepted: 127.0.0.1::33777
 SConnection: Client needs protocol version 3.8
 SConnection: Client requests security type None(1)
 main:        Keyboard keymap is not set
 VNCSConnST:  Server default pixel format depth 24 (32bpp) little-endian bgr888
Segmentation fault
```
SEGFAULT happens when trying to connect with noVNC

DaanSelen:
For context this is a Raspberry Pi OS (Mostly default)

But here:

```
Id=1
User=1000
Name=<my-user>
Timestamp=Wed 2026-01-14 15:43:49 CET
TimestampMonotonic=8187038
VTNr=7
Seat=seat0
Remote=no
Service=lightdm-autologin
Desktop=rpd-labwc
Scope=session-1.scope
Leader=1443
Audit=1
Type=wayland
Class=user
Active=yes
State=active
IdleHint=no
IdleSinceHint=0
IdleSinceHintMonotonic=0
CanIdle=yes
CanLock=yes
LockedHint=no
```

```
# labwc --version
labwc 0.9.2 (+xwayland +nls +rsvg +libsfdo)
```

```
echo $XDG_SESSION_TYPE
wayland

echo $WAYLAND_DISPLAY


echo $XDG_RUNTIME_DIR
/run/user/1000
```

If there is any more information needed, let me know.

CendioHalim:
@DaanSelen 

[...]

PR #2043, opened 2026-01-19
w0vncserver: Detect updates to keymap

CendioHalim:
There is currently a race where the `VirtualKeyboard` can be created before `Keyboard::handleKeyMap()` has been triggered (#2042).

This PR makes the keyboard handling more robust by disabling the keyboard if no keymap is set, and by detecting updates to the keymap and updating the virtual keyboard.

Note that this is only for the `WaylandDesktop`.

### DIFF
--- a/unix/w0vncserver/wayland/objects/Keyboard.cxx
+++ b/unix/w0vncserver/wayland/objects/Keyboard.cxx
@@ -134,32 +134,36 @@ void Keyboard::handleKeyMap(uint32_t format, int32_t fd, uint32_t size)
   if (!context || !context->ctx)
     return;
 
-  memset(codeMapQnumToKeyCode, 0, sizeof(codeMapQnumToKeyCode));
-
-  if (keyMap)
-    munmap(keyMap, keyboardSize);
-  keyboardFormat = format;
-  keyboardFd = fd;
-  keyboardSize = size;
-
   // https://wayland.app/protocols/wayland#wl_keyboard:enum:keymap_format
   // Format is either:
   //  no_keymap 0, or
   //  xkb_v1    1
-  if (keyboardFormat == WL_KEYBOARD_KEYMAP_FORMAT_XKB_V1) {
+  if (format == WL_KEYBOARD_KEYMAP_FORMAT_XKB_V1) {
     xkb_keymap* map;
     xkb_state* state;
+    char* newKeyMap;
 
-    keyMap = (char*)mmap(nullptr, keyboardSize, PROT_READ, MAP_PRIVATE,
-                         keyboardFd, 0);
-    if (!keyMap) {
+    newKeyMap = (char*)mmap(nullptr, size, PROT_READ, MAP_PRIVATE, fd, 0);
+    if (!newKeyMap) {
       vlog.error("Failed to map keymap");
       clearKeyMap();
       return;
     }
 
+    if (keyMap && strcmp(keyMap, newKeyMap) == 0) {
+      // Keymap unchanged, no need to update anything.
+      // This is a workaround for where zwp_virtual_keyboard_v1_keymap()
+      // will trigger a keymap event with the same keymap as before
+      return;
+    }
+
+    if (keyMap)
+      munmap(keyMap, keyboardSize);
+
+    keyMap = newKeyMap;
+
     map = xkb_keymap_new_from_buffer(context->ctx, keyMap,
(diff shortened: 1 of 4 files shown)


unix/w0vncserver/wayland/objects/Keyboard.cxx
```diff
@@ -76,17 +76,19 @@ struct XkbContext {
...
     setLEDstate(setLEDstate_)
 {
   xkb_context* ctx;
 
   ctx = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
   if (!ctx)  {
     // FIXME: fallback?
-    fatal_error("Failed to create xkb context");
+    vlog.error("Failed to create xkb context - keyboard will not work");
     return;
```
CendioOssman: Seems dangerous to return from a constructor like this. Should we throw an exception instead?

unix/w0vncserver/wayland/objects/Keyboard.cxx
```diff
@@ -134,7 +139,8 @@ bool Keyboard::updateState(uint32_t keycode, bool down,
 
 void Keyboard::handleKeyMap(uint32_t format, int32_t fd, uint32_t size)
 {
-  assert(context->ctx);
+  if (!context)
```
CendioOssman: Should we not keep checking `context->ctx`?

unix/w0vncserver/wayland/objects/Keyboard.cxx
CendioOssman: Or at least continue but skipping creating a `XkbContext`?

unix/w0vncserver/wayland/objects/Keyboard.h
```diff
@@ -36,8 +36,8 @@ namespace wayland {
     ~Keyboard();
 
     uint32_t getFormat() const { return keyboardFormat; }
-    int getFd() const { return keyboardFd; }
-    int getSize() const { return keyboardSize; }
+    int32_t getFd() const { return keyboardFd; }
```
CendioOssman, line 39: Vadan denna ändring? Fildeskriptorer är `int` som regel.

unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
```diff
@@ -46,26 +46,21 @@ VirtualKeyboard::VirtualKeyboard(Display* display, Seat* seat_)
...
+    vlog.error("Failed to create virtual keyboard - keyboard will not work");
     return;
   }
 
-  zwp_virtual_keyboard_v1_keymap(keyboard,
-                                 seat->getKeyboard()->getFormat(),
-                                 seat->getKeyboard()->getFd(),
-                                 seat->getKeyboard()->getSize());
+  if (!seat->getKeyboard()->hasKeymap()) {
+    vlog.error("Keyboard keymap is not set - keyboard will not work until it is set");
```
CendioOssman: Kommer de här visas varje uppstart för vissa compositors? Är inte det lite förvirrande och skrämmande?

unix/w0vncserver/wayland/objects/Keyboard.h
CendioHalim:
```c
void (*keymap)(void *data,
		       struct wl_keyboard *wl_keyboard,
		       uint32_t format,
		       int32_t fd,
		       uint32_t size);
```

This is to to match the types used in the protocol.

unix/w0vncserver/wayland/objects/Keyboard.cxx
CendioHalim: We could throw an exception. It's a bit annoying,though, since we create the `Keyboard` in an event handler from `Seat::seatCapabilities()`. Not sure how to handle the exception from `Seat`  in a good way.

Not creating `XkbContext` is a bit simpler.

unix/w0vncserver/wayland/objects/Keyboard.cxx
CendioHalim: Good point

unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
CendioHalim: Good point, changed to debug log, as well as another debug log the first time the keymap is set

PR #2044, opened 2026-01-21
Don't use fatal_error() in constructors

CendioHalim:
Better to use exceptions properly to get better control flow.

### DIFF
--- a/unix/w0vncserver/wayland/objects/Display.cxx
+++ b/unix/w0vncserver/wayland/objects/Display.cxx
@@ -22,6 +22,8 @@
 
 #include <assert.h>
 
+#include <stdexcept>
+
 #include <wayland-client-core.h>
 #include <wayland-client-protocol.h>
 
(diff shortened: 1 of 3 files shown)


unix/w0vncserver/wayland/objects/Display.cxx
```diff
@@ -50,20 +52,20 @@ Display::Display(const char* name)
 {
   display = wl_display_connect(name);
   if (!display)
-    fatal_error("Failed to connect to wayland display");
+    throw std::runtime_error("Failed to connect to wayland display");
 
   registry = wl_display_get_registry(display);
   if (!registry)
-    fatal_error("Failed to get registry");
+    throw std::runtime_error("Failed to get registry");
```
CendioOssman, line 60: You need to clean up `display` here.

CendioHalim: Fixed

PR #2045, opened 2026-01-21
Fix socket cleanup in GSocketSource

CendioHalim:
This matches the cleanup that happens in Xvnc and x0vncserver and should have been part of cd1521f0af2437505fe415ffff088e7c5dad017d

### DIFF
--- a/unix/w0vncserver/GSocketSource.cxx
+++ b/unix/w0vncserver/GSocketSource.cxx
@@ -175,7 +175,7 @@ int GSocketSource::prepare(int* timeout)
         g_source_remove_unix_fd(source, state->tag);
         server->removeSocket(sock);
         delete sock;
-        assert(fdMap.erase(fd));
+        fdMap.erase(fd);
       }
       continue;
     }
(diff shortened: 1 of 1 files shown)


unix/w0vncserver/GSocketSource.cxx
```diff
@@ -154,12 +156,28 @@ int GSocketSource::prepare(int* timeout)
...
+          break;
+        }
+      }
+
+      if (done) {
+        vlog.debug("Client gone, sock %d", fd);
+        g_source_remove_unix_fd(source, state->tag);
+        server->removeSocket(sock);
+        delete sock;
+        assert(fdMap.erase(fd));
```
CendioOssman: Overlooked this the first time, but this is dangerous. Everything in `assert()` will be removed if you compile without assertions. So you cannot have side effects in there.

CendioHalim: Didn't know that. Had a look at all asserts in `unix/w0vncserver/` and only found two, removed them.

CendioOssman (approved):

PR #2048, opened 2026-01-23
Decouple WaylandPixelBuffer and ScreencopyManager

CendioHalim:
The `WaylandPixelBuffer` is implemented a bit too much knowledge of the `ScreencopyManager`.

This PR tries to decouple them, with the intention of making the `WaylandPixelBuffer` usable with `ext-image-copy-capture` without large modifications.

### DIFF
--- a/unix/w0vncserver/wayland/objects/ScreencopyManager.h
+++ b/unix/w0vncserver/wayland/objects/ScreencopyManager.h
@@ -24,10 +24,9 @@
 #include <functional>
 
 #include <core/Region.h>
+#include <rfb/PixelFormat.h>
 
 #include "Object.h"
-
-namespace rfb { class PixelFormat; }
 
 struct wl_buffer;
 struct zwlr_screencopy_manager_v1;
@@ -49,8 +48,6 @@ namespace wayland {
 
     // Called when the remote output is resized
     void resize();
-
-    rfb::PixelFormat getPixelFormat();
 
     uint8_t* getBufferData();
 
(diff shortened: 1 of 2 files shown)


unix/w0vncserver/wayland/objects/ScreencopyManager.h
```diff
@@ -21,12 +21,13 @@
 
 #include <stddef.h>
 
+#include <functional>
+
 #include <core/Region.h>
+#include <rfb/PixelFormat.h>
```
CendioOssman, line 27: Is this really needed? You don't seem to be using the type in the header.

unix/w0vncserver/wayland/objects/ScreencopyManager.h
```diff
@@ -41,26 +42,25 @@ namespace wayland {
...
+    void resize();
+
+    rfb::PixelFormat getPixelFormat();
+
     uint8_t* getBufferData();
 
+  private:
     // Capture the next frame. This function is asynchronous.
     // Framebuffer data is available after handleScreencopyReady() has
     // been called, in which case captureFrameDone() will be called.
```
CendioOssman, line 57: This looks like it needs to be removed or modified.

unix/w0vncserver/wayland/objects/ScreencopyManager.h
CendioOssman: Seems like it might have just been in the wrong commit?

CendioHalim: Yeah, wrong commit. Fixed now

PR #2049, opened 2026-01-26
w0vncserver: Add Clipboard support to WaylandDesktop

CendioHalim:
This adds clipboard support using ext-data-control-v1.

### DIFF
--- a/unix/w0vncserver/parameters.cxx
+++ b/unix/w0vncserver/parameters.cxx
@@ -45,6 +45,14 @@ core::BoolParameter
   rawKeyboard("RawKeyboard",
               "Send keyboard events straight through and avoid mapping "
               "them to the current keyboard layout", false);
+core::BoolParameter
+  setPrimary("SetPrimary",
+             "Set the primary as well as the selection clipboard",
+             true);
+core::BoolParameter
+  sendPrimary("SendPrimary",
+              "Send the primary as well as the selection clipboard",
+              true);
 
 
 static const char* defaultDesktopName()
--- a/unix/w0vncserver/parameters.h
+++ b/unix/w0vncserver/parameters.h
@@ -28,5 +28,7 @@ extern core::BoolParameter localhostOnly;
 extern core::StringParameter desktopName;
 extern core::StringParameter interface;
 extern core::BoolParameter rawKeyboard;
+extern core::BoolParameter setPrimary;
+extern core::BoolParameter sendPrimary;
 
 #endif // __W0VNCSERVER_PARAMETERS_H__
--- a/unix/w0vncserver/w0vncserver.man
+++ b/unix/w0vncserver/w0vncserver.man
@@ -258,6 +258,15 @@ Default is \fBTLSVnc,VncAuth\fP.
 .B \-SendCutText
 Send clipboard changes to clients. Default is on.
 .
+.B \-SendPrimary
+Send the primary as well as the selection clipboard to clients. Note
+that this does not work on all compositors. Default is on.
+.
+.TP
+.B \-SetPrimary
+Set the primary as well as the selection clipboard. Note that this does
+not work on all compositors. Default is on.
+.
 .TP
 .B \-UseBlacklist
 Temporarily reject connections from a host if it repeatedly fails to
--- a/unix/w0vncserver/wayland/objects/DataControl.cxx
+++ b/unix/w0vncserver/wayland/objects/DataControl.cxx
@@ -1,0 +1,531 @@
+/* Copyright 2025 Adam Halim for Cendio AB
+ *
+ * This is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this software; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ * USA.
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include <string.h>
+#include <unistd.h>
+#include <assert.h>
+#include <fcntl.h>
+
+#include <stdexcept>
+
+#include <gio-unix-2.0/gio/gunixinputstream.h>
+#include <gio/gio.h>
+#include <ext-data-control-v1.h>
+
+#include <core/string.h>
+#include <core/LogWriter.h>
+
+#include "../../parameters.h"
+#include "Display.h"
+#include "Seat.h"
+#include "DataControl.h"
+
+using namespace wayland;
+
+static const char* MIME_TEXT_PLAIN = "text/plain";
+static const char* MIME_TEXT_PLAIN_UTF8 = "text/plain;charset=utf-8";
+
+static const char* MIME_TYPES[] {
+  MIME_TEXT_PLAIN_UTF8,
+  MIME_TEXT_PLAIN,
+};
+
+/* FIXME: There is currently no way of knowing if it is us or some
+*  other application that owns the clipboard. This means we can't
+*  distinguish if a clipboard change was initiated by us, or by
+*  someone else. A workaround for now is to use a unique mime type to
+*  indicate ownership.
+*  https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/issues/111
+*/
+static std::string tigervncMimeType = "application/x-tigervnc-data-control-" + std::to_string(getpid());
+
+struct PendingData {
+  int32_t fd;
+  std::string mimeType;
+};
+
+static core::LogWriter vlog("WaylandDataControl");
+
+const ext_data_control_device_v1_listener DataControl::deviceListener = {
+  .data_offer = [](void* data, ext_data_control_device_v1*,
+                   ext_data_control_offer_v1* offer) {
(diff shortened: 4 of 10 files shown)


unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,485 @@
...
+#include "DataControl.h"
+
+core::BoolParameter
+  setPrimary("SetPrimary",
+             "Set the PRIMARY as well as the CLIPBOARD selection",
+             true);
+core::BoolParameter
+  sendPrimary("SendPrimary",
+              "Send the PRIMARY as well as the CLIPBOARD selection",
+              true);
```
CendioOssman: Given that we have a `parameters.cxx`, don't we want to put everything there?

unix/w0vncserver/w0vncserver.man
```diff
@@ -257,6 +257,12 @@ Default is \fBTLSVnc,VncAuth\fP.
 .TP
 .B \-SendCutText
 Send clipboard changes to clients. Default is on.
+.B \-SendPrimary
```
CendioOssman, line 261: Missing `.`

unix/w0vncserver/w0vncserver.man
```diff
@@ -257,6 +257,12 @@ Default is \fBTLSVnc,VncAuth\fP.
...
+Send the PRIMARY as well as the CLIPBOARD selection to clients. Default is on.
+.
+.TP
+.B \-SetPrimary
+Set the PRIMARY as well as the CLIPBOARD selection. Default is on.
```
CendioOssman: We probably need a note that this isn't supported with all compositors.

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,485 @@
...
+
+  /* FIXME: There is currently no way of knowing if it is us or some
+  *  other application that owns the clipboard. This means we can't
+  *  distinguish if a clipboard change was initiated by us, or by
+  *  someone else. A workaround for now is to use a unique mime type to
+  *  indicate ownership.
+  *  https://gitlab.freedesktop.org/wlroots/wlr-protocols/-/issues/111
+  */
+  pid = getpid();
+  tigervncMimeType = "application/x-tigervnc-data-control-" + std::to_string(pid);
```
CendioOssman: Could be moved to initilization? The process pid is static and known on startup.

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,485 @@
...
+   readInProgress(false), pendingReadMimeType(nullptr),
+   readStream(nullptr), readCancellable(nullptr), readBuffer(""),
+   clipboardAnnounceCb(clipboardAnnounceCb_),
+   clipboardRequestCb(clipboardRequestCb_),
+   sendClipboardDataCb(sendClipboardDataCb_)
+{
+  pid_t pid;
+
+  manager = (ext_data_control_manager_v1*)boundObject;
+  device = ext_data_control_manager_v1_get_data_device(manager, seat->getSeat());
```
CendioOssman, line 119: Any error management needed when creating Wayland objects?

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,485 @@
...
+  if (!selectionSource) {
+    vlog.error("Failed to create data source for selection");
+    return;
+  }
+
+  ext_data_control_source_v1_add_listener(selectionSource, &sourceListener, this);
+
+  ext_data_control_source_v1_offer(selectionSource, tigervncMimeType.c_str());
+  for (const char* mimeType : MIME_TYPES)
+    ext_data_control_source_v1_offer(selectionSource, mimeType);
```
CendioOssman: Lots of duplication between primary and "regular" selection. Should we share some code?

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,485 @@
...
+
+  // Nothing to receive
+  if (!lastOffer)
+    return;
+
+  if (readInProgress)
+    return;
+
+  mimeType = nullptr;
+  for (uint i = 0; i < sizeof(MIME_TYPES) / sizeof(MIME_TYPES[0]); i++) {
```
CendioOssman: Why not range based for loop like the other place you use `MIME_TYPES`?

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,485 @@
...
+void DataControl::handleSelection(ext_data_control_offer_v1* offer)
+{
+  if (!offer) {
+    clipboardAnnounceCb(false);
+    return;
+  }
+
+  // Ignore our own requests
+  if (offer == ourOffer)
+    return;
```
CendioOssman, line 335: When does this happen? I thought the funky mime type was to handle this case?

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,485 @@
...
+
+  clipboardAnnounceCb(true);
```
CendioOssman: We should check that there is a compatible mime type first.

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,485 @@
...
+}
+
+void DataControl::handleDataOffer(ext_data_control_offer_v1* offer)
+{
+  ext_data_control_offer_v1_add_listener(offer, &offerListener, this);
+  lastOffer = offer;
+  availableMimeTypes.clear();
+}
+
+void DataControl::handleSelection(ext_data_control_offer_v1* offer)
```
CendioOssman, line 319: Should we check that the offer here matches `lastOffer`?

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,485 @@
...
+    primarySource = nullptr;
+  if (source == selectionSource)
+    selectionSource = nullptr;
+}
+
+void DataControl::handleOffer(const char* mimeType)
+{
+  if (strcmp(mimeType, tigervncMimeType.c_str()) == 0) {
+    assert(lastOffer);
+    ourOffer = lastOffer;
```
CendioOssman: A bit confusing with all these different offer pointers. Can't we do the mime check later and avoid this variable?

unix/w0vncserver/wayland/objects/DataControl.cxx
CendioHalim: `wayland::Object` (parent class) will throw an exception if it fails, but yeah we should throw an exception if `ext_data_control_manager_v1_get_data_device()` fails.

unix/w0vncserver/parameters.h
```diff
@@ -28,5 +28,8 @@ extern core::BoolParameter localhostOnly;
 extern core::StringParameter desktopName;
 extern core::StringParameter interface;
 extern core::BoolParameter rawKeyboard;
+extern core::BoolParameter rawKeyboard;
```
CendioOssman: Duplication

unix/w0vncserver/parameters.cxx
```diff
@@ -45,6 +45,14 @@ core::BoolParameter
   rawKeyboard("RawKeyboard",
               "Send keyboard events straight through and avoid mapping "
               "them to the current keyboard layout", false);
+core::BoolParameter
+  setPrimary("SetPrimary",
+             "Set the PRIMARY as well as the CLIPBOARD selection",
```
CendioOssman: I just realized, these are X11 terms. We should probably use the terminogy that wayland uses. Which is probably similar, but likely not CAPS.

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,510 @@
...
+    ext_data_control_offer_v1_destroy(lastOffer);
+    lastOffer = nullptr;
+  }
+}
+
+void DataControl::setSelection()
+{
+  if (selectionSource)
+    ext_data_control_source_v1_destroy(selectionSource);
+  primarySource = nullptr;
```
CendioOssman: Should be `selectionSource`.

unix/w0vncserver/parameters.cxx
CendioHalim: Seems like they are using the terminology "Selection", and "Primary selection"

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,519 @@
...
+
+void DataControl::clearSelection()
+{
+  if (selectionSource)
+    ext_data_control_source_v1_destroy(selectionSource);
+  selectionSource = nullptr;
+
+  ext_data_control_device_v1_set_selection(device, nullptr);
+
+  clearPendingWrites();
```
CendioOssman, line 204: Should probably clear any pending reads as well.

unix/w0vncserver/wayland/objects/DataControl.cxx
```diff
@@ -0,0 +1,519 @@
...
+  if (readCancellable)
+    g_object_unref(readCancellable);
+
+  if (lastOffer) {
+    ext_data_control_offer_v1_destroy(lastOffer);
+    lastOffer = nullptr;
+  }
+}
+
+void DataControl::setSelection()
```
CendioOssman, line 156: Should clear any pending reads and writes here as well.

CendioHalim: Fixed

CendioOssman: I can't see any stopping of pending writes in this code path?

CendioHalim: Fixed

PR #2053, opened 2026-01-29
w0vncserver: General error handling improvements

CendioHalim:
General clean up of the error handling in w0vncserver.

Removed most instances of `fatal_error()` and instead call `server->closeClients()`. Also use exceptions in more places.

### DIFF
--- a/unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
+++ b/unix/w0vncserver/wayland/objects/VirtualKeyboard.cxx
@@ -21,6 +21,7 @@
 #endif
 
 #include <time.h>
+#include <assert.h>
 
 #include <stdexcept>
 
@@ -55,7 +56,7 @@ VirtualKeyboard::VirtualKeyboard(Display* display, Seat* seat_)
   if (!keyboard)
     throw std::runtime_error("Failed to create virtual keyboard");
 
-  if (seat->getKeyboard()->hasKeymap())
+  if (seat->getKeyboard() && seat->getKeyboard()->hasKeymap())
     setupKeyboard();
   else
     vlog.debug("Keyboard keymap is not set - keyboard will not work until it is set");
@@ -136,11 +137,16 @@ void VirtualKeyboard::key(uint32_t keysym, uint32_t keycode, bool down)
 
 bool VirtualKeyboard::keymapUpdated()
 {
+  if (!seat->getKeyboard())
+    return false;
+
   return keyboardFd != seat->getKeyboard()->getFd();
 }
 
 void VirtualKeyboard::setupKeyboard()
 {
+  assert(seat->getKeyboard());
+
   if (!keyboardFd)
     vlog.debug("Keymap set - keyboard will now work");
 


unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
```diff
@@ -103,7 +104,10 @@ ScreencopyManager::ScreencopyManager(Display* display, Output* output_,
   shm = new Shm(display);
 
   size = output_->getWidth() * output_->getHeight() * 4;
-  assert(size);
+  if (!size)
+    throw std::runtime_error(core::format("Invalid output size %dx%d",
+                                          output_->getWidth(),
+                                          output_->getHeight()));
```
CendioOssman: Need to delete `shm` here.

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
```diff
@@ -103,7 +104,10 @@ ScreencopyManager::ScreencopyManager(Display* display, Output* output_,
...
 
   initBuffers(size);
```
CendioOssman, line 111: This also needs cleanup handling as it can throw exceptions.

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
```diff
@@ -194,6 +221,9 @@ void ScreencopyManager::initBuffers(size_t size)
 
 rfb::PixelFormat ScreencopyManager::convertPixelformat(uint32_t format)
 {
+  if (!active)
+    return rfb::PixelFormat();
+
```
CendioOssman: Doesn't look like all of these checks are needed. Some of these functions do not rely on any active state?

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
```diff
@@ -297,8 +297,15 @@ void ScreencopyManager::handleScreencopyFailed()
   if (!active)
     return;
 
-  vlog.error("Frame could not be copied");
-  captureFrameDone();
+  // We will get a failure if the output size has changed, which is
+  // expected during a resize.
+  if (output->getWidth() != info->width || output->getHeight() != info->height) {
+    vlog.status("Frame could not be copied due to resize - skipping frame");
```
CendioOssman: If this is normal during every resize, then I don't think it needs to be logged.

unix/w0vncserver/pipewire/PipeWireStream.cxx
```diff
@@ -62,11 +62,13 @@ const pw_stream_events PipeWireStream::streamEventsHandler {
...
+  : pipeWireFd(pipeWireFd_), active(true)
 {
   source = new PipeWireSource();
 
   context = pw_context_new(source->getLoop(), nullptr, 0);
+  if (!context)
+    throw std::runtime_error("Failed to create PipeWire context");
   core = pw_context_connect_fd(context, pipeWireFd_, nullptr, 0);
   if (!core)
     throw std::runtime_error("Failed to connect to PipeWire FD");
```
CendioOssman: Cleanup of objects needed here.

unix/w0vncserver/wayland/objects/ShmPool.cxx
```diff
@@ -60,5 +60,12 @@ wl_buffer* ShmPool::createBuffer(int32_t offset, int32_t width,
                                  int32_t height, int32_t stride,
                                  uint32_t format)
 {
-  return wl_shm_pool_create_buffer(pool, offset, width, height, stride, format);
+  wl_buffer* buffer;
+
+  buffer = wl_shm_pool_create_buffer(pool, offset, width, height, stride, format);
+
+  if (!buffer)
+    throw std::runtime_error("Failed to create wl_buffer from pool");
```
CendioOssman: This seems odd if it does nothing more than the allocation. Wouldn't it be better to return `nullptr` and let the caller decide what to do?

unix/w0vncserver/portals/PortalDesktop.cxx
```diff
@@ -116,9 +116,14 @@ void PortalDesktop::start()
...
-                                    cancelStart, initClipboard,
-                                    clipboardSubscribe);
+  try {
+    remoteDesktop = new RemoteDesktop(restoreToken, startPipewire,
+                                      cancelStart, initClipboard,
+                                      clipboardSubscribe);
+  } catch (std::exception& e) {
+    vlog.error("error initializing RemoteDesktop: %s", e.what());
+    server->closeClients("Failed to start remote desktop session");
+  }
```
CendioOssman, line 127: This is going to crash as `remoteDesktop` is used right below this.

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
CendioHalim: From what I've seen, it's not every resize, more like ~50%. I'll remove the logging.

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
```diff
@@ -103,9 +104,19 @@ ScreencopyManager::ScreencopyManager(Display* display, Output* output_,
...
+    delete shm;
+    throw std::runtime_error(core::format("Invalid output size %dx%d",
+                                          output_->getWidth(),
+                                          output_->getHeight()));
+  }
 
-  initBuffers(size);
+  try {
+    initBuffers(size);
+  } catch (std::exception&) {
```
CendioOssman: Should probably be `catch (...)` as this is more of a "finally" construct

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
```diff
@@ -103,9 +104,19 @@ ScreencopyManager::ScreencopyManager(Display* display, Output* output_,
   shm = new Shm(display);
```
CendioOssman: Given the mess of cleaning this up, have you considered making this a normal member rather than a pointer? It would automate the cleanup.

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
```diff
@@ -169,23 +183,31 @@ void ScreencopyManager::resize()
...
   fd = memfd_create("w0vncserver-shm", FD_CLOEXEC);
-  if (fd < 0) {
-    fatal_error("Failed to allocate shm: %s", strerror(errno));
-    return;
-  }
+  if (fd < 0)
+    throw std::runtime_error(core::format("Failed to allocate shm: %s", strerror(errno)));
+
+  if (ftruncate(fd, size) < 0)
+    throw std::runtime_error(core::format("Failed to truncate shm: %s", strerror(errno)));
```
CendioOssman: Need to close `fd` here, I guess?

CendioOssman (changes requested):

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
```diff
@@ -87,25 +87,32 @@ ScreencopyManager::ScreencopyManager(Display* display, Output* output_,
...
+                                          output_->getWidth(),
+                                          output_->getHeight()));
+  }
 
-  initBuffers(size);
+  try {
+    initBuffers(size);
+  } catch (...) {
+    throw;
+  }
```
CendioOssman: The try/catch is a bit of a no-op now and should be removed.

CendioOssman (changes requested):

PR #2065, opened 2026-02-11
w0vncserver: add image copy capture support

tobfah:
Adds Wayland screen capture support to w0vncserver using the ext-image-copy-capture-v1 protocol (the successor to the deprecated wlr-screencopy-unstable-v1). The latter remains as a fallback for compositors that do not support the new protocol yet.

Also adds support for local cursor capture.

Tested on Fedora43 with Labwc (xfce4 session).

### DIFF
--- a/unix/w0vncserver/CMakeLists.txt
+++ b/unix/w0vncserver/CMakeLists.txt
@@ -17,6 +17,7 @@ add_executable(w0vncserver
   wayland/objects/DataControl.cxx
   wayland/objects/Display.cxx
   wayland/objects/Keyboard.cxx
+  wayland/objects/Pointer.cxx
   wayland/objects/Seat.cxx
   wayland/objects/Shm.cxx
   wayland/objects/ShmPool.cxx
@@ -25,6 +26,7 @@ add_executable(w0vncserver
   wayland/objects/ImageCaptureSource.cxx
   wayland/objects/ImageCopyCaptureManager.cxx
   wayland/objects/ImageCopyCaptureSession.cxx
+  wayland/objects/ImageCopyCaptureCursorSession.cxx
   wayland/objects/ScreencopyManager.cxx
   wayland/WaylandDesktop.cxx
   wayland/WaylandPixelBuffer.cxx
--- a/unix/w0vncserver/CMakeLists.txt
+++ b/unix/w0vncserver/CMakeLists.txt
@@ -22,6 +22,9 @@ add_executable(w0vncserver
   wayland/objects/ShmPool.cxx
   wayland/objects/VirtualPointer.cxx
   wayland/objects/VirtualKeyboard.cxx
+  wayland/objects/ImageCaptureSource.cxx
+  wayland/objects/ImageCopyCaptureManager.cxx
+  wayland/objects/ImageCopyCaptureSession.cxx
   wayland/objects/ScreencopyManager.cxx
   wayland/WaylandDesktop.cxx
   wayland/WaylandPixelBuffer.cxx
@@ -40,6 +43,9 @@ add_executable(w0vncserver-forget
 # Generate protocol headers
 set(GEN_WAYLAND_SRCS "")
 set(WAYLAND_PROTOCOLS
+  ext-foreign-toplevel-list-v1
+  ext-image-capture-source-v1
+  ext-image-copy-capture-v1
   wlr-screencopy-unstable-v1
   wlr-virtual-pointer-unstable-v1
   virtual-keyboard-unstable-v1
--- a/unix/w0vncserver/wayland/WaylandDesktop.cxx
+++ b/unix/w0vncserver/wayland/WaylandDesktop.cxx
@@ -115,7 +115,7 @@ void WaylandDesktop::start()
   };
 
   try {
-    pb = new WaylandPixelBuffer(display, output, server, desktopReadyCb);
+    pb = new WaylandPixelBuffer(display, output, seat, server, desktopReadyCb);
   } catch (std::exception& e) {
     vlog.error("Error initializing pixel buffer: %s", e.what());
     server->closeClients("Failed to start remote desktop session");
--- a/unix/w0vncserver/wayland/WaylandDesktop.cxx
+++ b/unix/w0vncserver/wayland/WaylandDesktop.cxx
@@ -1,4 +1,4 @@
-/* Copyright 2025 Adam Halim for Cendio AB
+/* Copyright 2025-2026 Adam Halim for Cendio AB
  *
  * This is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -224,7 +224,12 @@ bool WaylandDesktop::available()
 {
   wayland::Display display;
 
-  return display.interfaceAvailable("zwlr_screencopy_manager_v1") &&
+  // We need either wlr-screencopy OR ext-image-copy-capture
+  return (display.interfaceAvailable("zwlr_screencopy_manager_v1") ||
+         (
+            display.interfaceAvailable("ext_image_copy_capture_manager_v1") &&
+            display.interfaceAvailable("ext_output_image_capture_source_manager_v1")
+         )) &&
          display.interfaceAvailable("zwlr_virtual_pointer_manager_v1") &&
          display.interfaceAvailable("zwp_virtual_keyboard_manager_v1");
 }
--- a/unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
+++ b/unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
@@ -48,6 +48,7 @@ static core::LogWriter vlog("WaylandPixelBuffer");
 
 WaylandPixelBuffer::WaylandPixelBuffer(wayland::Display* display,
                                        wayland::Output* output_,
+                                       wayland::Seat* seat,
                                        rfb::VNCServer* server_,
                                        std::function<void()> desktopReadyCallback_)
   : firstFrame(true), desktopReadyCallback(desktopReadyCallback_),
@@ -59,6 +60,15 @@ WaylandPixelBuffer::WaylandPixelBuffer(wayland::Display* display,
     std::bind(&WaylandPixelBuffer::bufferEvent, this, std::placeholders::_1,
               std::placeholders::_2, std::placeholders::_3);
 
+  std::function<void(int, int, const core::Point&, uint32_t, const uint8_t*)>
+    cursorImageCb = std::bind(&WaylandPixelBuffer::cursorImageEvent, this,
+                              std::placeholders::_1, std::placeholders::_2,
+                              std::placeholders::_3, std::placeholders::_4,
+                              std::placeholders::_5);
+
+  std::function<void(const core::Point&)> cursorPosCb =
+    std::bind(&WaylandPixelBuffer::cursorPosEvent, this, std::placeholders::_1);
+
   std::function<void()> stoppedCb = [this]() {
     server->closeClients("The remote session stopped");
   };
@@ -69,7 +79,10 @@ WaylandPixelBuffer::WaylandPixelBuffer(wayland::Display* display,
 
     imageCopyCaptureManager = new wayland::ImageCopyCaptureManager(display,
                                                                    imageCaptureSource,
+                                                                   seat,
                                                                    bufferEventCb,
+                                                                   cursorImageCb,
+                                                                   cursorPosCb,
                                                                    stoppedCb);
     imageCopyCaptureManager->createSession();
   } else {
@@ -83,6 +96,29 @@ WaylandPixelBuffer::~WaylandPixelBuffer()
   delete screencopyManager;
   delete imageCopyCaptureManager;
   delete imageCaptureSource;
+}
+
+void WaylandPixelBuffer::cursorImageEvent(int width, int height,
+                                          const core::Point& hotspot,
+                                          uint32_t shmFormat,
+                                          const uint8_t* src)
+{
+  uint8_t* cursorData;
+
+  cursorData = nullptr;
+  try {
+    cursorData = convertCursorBuffer(src, shmFormat, width, height);
+    server->setCursor(width, height, hotspot, cursorData);
+  } catch (std::exception& e) {
+    vlog.error("Failed to set cursor: %s", e.what());
+  }
+
+  delete cursorData;
+}
+
+void WaylandPixelBuffer::cursorPosEvent(const core::Point& pos)
+{
+  server->setCursorPos(pos, true);
 }
 
 void WaylandPixelBuffer::bufferEvent(uint8_t* buffer, core::Region damage,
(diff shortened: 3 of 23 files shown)


unix/w0vncserver/CMakeLists.txt
```diff
@@ -41,6 +44,8 @@ add_executable(w0vncserver-forget
 set(GEN_WAYLAND_SRCS "")
 set(WAYLAND_PROTOCOLS
   ext-foreign-toplevel-list-v1
+  ext-image-capture-source-v1
```
CendioOssman, line 49: This is in the wrong commit

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.h
```diff
@@ -0,0 +1,59 @@
...
+
+  class ImageCaptureSource;
+  class ImageCopyCaptureSession;
+
+  class ImageCopyCaptureManager : public Object {
+  public:
+    ImageCopyCaptureManager(Display* display,
+                            ImageCaptureSource* source,
+                            std::function<void(uint8_t*, core::Region, uint32_t)>
+                              bufferEventCb,
```
CendioOssman, line 44: This is called `frameReadyCb` in `ImageCopyCaptureSession`. Could we try to find a consistent name for this? At least within the same protocol.

unix/w0vncserver/wayland/objects/ImageCopyCaptureSession.cxx
```diff
@@ -0,0 +1,329 @@
...
+  if ((width && height) && (width_ != width || height_ != height))
+    cleanupBuffers();
+
+  width = width_;
+  height = height_;
+}
+
+void ImageCopyCaptureSession::handleShmFormat(uint32_t format_)
+{
+  formats.push_back((uint32_t)format_);
```
CendioOssman: What happens if the formats change during execution? We need to reset formats at some point.

unix/w0vncserver/wayland/objects/ImageCopyCaptureSession.cxx
```diff
@@ -0,0 +1,329 @@
...
+
+  ext_image_copy_capture_session_v1_destroy(session);
+  session = nullptr;
+
+  stoppedCb();
+}
+
+void ImageCopyCaptureSession::handleFrameTransform(uint32_t /* transform */)
+{
+}
```
CendioOssman: What is this for? Can we really ignore it?

unix/w0vncserver/wayland/objects/ImageCopyCaptureSession.cxx
```diff
@@ -0,0 +1,329 @@
...
+    vlog.error("Could not capture frame: unknown reason - trying again");
+    createFrame();
+  } else if (reason == EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_BUFFER_CONSTRAINTS) {
+    vlog.error("Invalid buffer - trying again");
+    cleanupBuffers();
+    createFrame();
+  } else if (reason == EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_STOPPED) {
+    vlog.error("Could not capture frame, stopping capture session");
+    stoppedCb();
+  }
```
CendioOssman, line 294: What if it's not any of the listed reasons?

unix/w0vncserver/wayland/objects/ImageCopyCaptureSession.cxx
```diff
@@ -0,0 +1,329 @@
...
+{
+  assert(frame == nullptr);
+
+  frame = ext_image_copy_capture_session_v1_create_frame(session);
+  ext_image_copy_capture_frame_v1_add_listener(frame, &frameListener, this);
+
+  damage.clear();
+
+  ext_image_copy_capture_frame_v1_attach_buffer(frame, buffer);
+  ext_image_copy_capture_frame_v1_capture(frame);
```
CendioOssman, line 313: The spec states:
> When a wl_buffer is captured for the first time, or when the client doesn't track damage, the
> client must damage the whole buffer.

unix/w0vncserver/wayland/objects/ImageCopyCaptureSession.cxx
```diff
@@ -0,0 +1,329 @@
...
+}
+
+void ImageCopyCaptureSession::handleDone()
+{
+  size_t size;
+  int fd;
+
+  assert(!shm);
+  assert(!pool);
+  assert(!buffer);
```
CendioOssman, line 168: These could be set if any constraints change while capturing. It's only on an actual resize that we free these. Any other change will result in a crash.

unix/w0vncserver/wayland/objects/ImageCopyCaptureSession.cxx
```diff
@@ -0,0 +1,329 @@
...
+    frame = nullptr;
+  }
+
+  if (reason == EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_UNKNOWN) {
+    vlog.error("Could not capture frame: unknown reason - trying again");
+    createFrame();
+  } else if (reason == EXT_IMAGE_COPY_CAPTURE_FRAME_V1_FAILURE_REASON_BUFFER_CONSTRAINTS) {
+    vlog.error("Invalid buffer - trying again");
+    cleanupBuffers();
+    createFrame();
```
CendioOssman: This doesn't look like it will work. `cleanupBuffers()` will free everything, but `createFrame()` will not allocate new buffers.

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
```diff
@@ -227,7 +227,6 @@ void ScreencopyManager::handleScreencopyBuffer(uint32_t format,
     .height = height,
     .stride = stride
   };
-
```
CendioOssman, line 251: This looks like unrelated noise.

unix/w0vncserver/wayland/WaylandDesktop.cxx
```diff
@@ -224,7 +224,12 @@ bool WaylandDesktop::available()
...
 
-  return display.interfaceAvailable("zwlr_screencopy_manager_v1") &&
+  // We need either wlr-screencopy OR ext-image-copy-capture
+  return (display.interfaceAvailable("zwlr_screencopy_manager_v1") ||
+         (
+            display.interfaceAvailable("ext_image_copy_capture_manager_v1") &&
+            display.interfaceAvailable("ext_output_image_capture_source_manager_v1")
+         )) &&
          display.interfaceAvailable("zwlr_virtual_pointer_manager_v1") &&
          display.interfaceAvailable("zwp_virtual_keyboard_manager_v1");
```
CendioOssman, line 234: Not really related to this PR, but didn't we make these interfaces optional?

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2025 Adam Halim for Cendio AB
+/* Copyright 2026 Adam Halim for Cendio AB
```
CendioOssman: `2025-2026`. We generally want the entire span.

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
```diff
@@ -21,34 +21,37 @@
...
 
 #include <assert.h>
-#include <unistd.h>
-
 #include <stdexcept>
 
 #include <pixman.h>
-#include <wayland-client-core.h>
 #include <wayland-client-protocol.h>
-#include <wayland-client.h>
```
CendioOssman, line 29: These changes look a bit unrelated?

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
```diff
@@ -21,34 +21,37 @@
...
 
 #include "../w0vncserver.h"
 #include "objects/Output.h"
 #include "objects/Display.h"
 #include "objects/ScreencopyManager.h"
+#include "objects/ImageCaptureSource.h"
+#include "objects/ImageCopyCaptureManager.h"
 #include "WaylandPixelBuffer.h"
 
+using namespace wayland;
```
CendioOssman: This is a gray area. Normally we only allow `using namespace` for defining objects within said namespace. And `WaylandPixelBuffer` is not in `wayland`. But it's in a directory specific to the wayland backend, so I could be convinced to ignore this instance.

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
```diff
@@ -21,34 +21,37 @@
...
+
 static core::LogWriter vlog("WaylandPixelBuffer");
 
-WaylandPixelBuffer::WaylandPixelBuffer(wayland::Display* display,
-                                       wayland::Output* output_,
+WaylandPixelBuffer::WaylandPixelBuffer(Display* display_,
```
CendioOssman: Why did `display` change?

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.h
```diff
@@ -33,24 +33,34 @@ namespace wayland {
...
                             std::function<void(uint8_t*, core::Region, uint32_t)>
                               bufferEventCb,
+                            std::function<void(int, int, const core::Point&, uint32_t, const uint8_t*)>
+                              cursorImageCb,
+                            std::function<void(const core::Point&)> cursorPosCb,
                             std::function<void()> stoppedCb);
     ~ImageCopyCaptureManager();
 
     void createSession();
+    void createPointerCursorSession();
```
CendioOssman: Why not do this as part of `createSession`? This object already represents all of the handling.

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.cxx
```diff
@@ -72,3 +80,22 @@ void ImageCopyCaptureManager::createSession()
                                         bufferEventCb, stoppedCb);
 }
 
+void ImageCopyCaptureManager::createPointerCursorSession()
+{
+  Pointer* pointer;
+
+  pointer = seat->getPointer();
```
CendioOssman, line 85: `pointer` can be `nullptr`.

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
```diff
@@ -46,6 +46,7 @@ static core::LogWriter vlog("WaylandPixelBuffer");
 
 WaylandPixelBuffer::WaylandPixelBuffer(Display* display_,
                                        Output* output_,
+                                       Seat* seat_,
```
CendioOssman: No `_` needed here.

CendioOssman (changes requested):

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
CendioHalim: No you're right, this was a mistake that has been reverted.

unix/w0vncserver/wayland/objects/ImageCopyCaptureSession.cxx
CendioHalim: Should be fixed now

unix/w0vncserver/wayland/objects/ImageCopyCaptureSession.cxx
CendioHalim: We've now added support transformed outputs

unix/w0vncserver/wayland/WaylandDesktop.cxx
CendioHalim: Yeah we kind of did, if we fail to initialize `VirtualPointer` or `VirtualKeyboard`, the `WaylandDesktop` will work without them. But we still require the interfaces to be available (there is an assert in `Object.cxx`). We could change the code to only require one of the imaging protocols instead.

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.h
CendioHalim: Figured we'd let the Manager just be a dumb object that implements the protocol, and let the caller do whatever invocations they need. Would you prefer it if the cursor session always be created implicitly?

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.h
tobfah: We made the method names consistent

unix/w0vncserver/wayland/objects/ScreencopyManager.cxx
tobfah: Addressed this

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
tobfah: Addressed

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
tobfah: Reverted this

unix/w0vncserver/wayland/WaylandPixelBuffer.cxx
tobfah: Addressed

unix/w0vncserver/CMakeLists.txt
tobfah: We squashed the two commits

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.h
CendioOssman: We've already merged the two concepts in the fact that you have to specify cursor callbacks to the construct. So the caller is always forced to account for the cursor in some way. It doesn't seem like anything is simpler or more flexible for the caller with this split?

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.cxx
```diff
@@ -0,0 +1,107 @@
...
+  // FIXME: Make it possible to re-enable this if a cursor is added in
+  // the future like we do in VirtualKeyboard
+  if (!pointer)
+    throw std::runtime_error(core::format("Unable to create cursor capture "
+                                          "session: no pointer found"));
```
CendioOssman: `core::format()` seems like a copy-and-paste error here

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.h
```diff
@@ -51,6 +51,9 @@ namespace wayland {
     void createSession();
     void createPointerCursorSession();
 
+    ImageCopyCaptureSession* getSession() { return session; }
+    ImageCopyCaptureCursorSession* getCursorSession() { return cursorSession; }
```
CendioOssman: This isn't ideal as it breaks the abstraction. Why not include the transformation in the buffer event, like we do other attributes?

unix/w0vncserver/wayland/objects/ImageCopyCaptureSession.h
```diff
@@ -52,6 +53,7 @@ namespace wayland {
     uint32_t getWidth() const { return width; }
     uint32_t getHeight() const { return height; }
     uint32_t getFormat() const { return format; }
+    wl_output_transform getTransform() const { return (wl_output_transform)transform; }
```
CendioOssman: If we can store this as an `uint32_t`, why not keep this format in the entire interface? Avoids having extra headers included.

CendioOssman (changes requested):

CendioHalim:
All comment should now have been addressed.

Also realized that we can make `ScreenCopyManager` support transformed outputs, added that as well.

CendioOssman (approved):

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.h
```diff
@@ -0,0 +1,68 @@
...
+#ifndef __WAYLAND_IMAGE_COPY_CAPTURE_MANAGER_H__
+#define __WAYLAND_IMAGE_COPY_CAPTURE_MANAGER_H__
+
+#include <functional>
+
+#include "Object.h"
+
+struct ext_image_copy_capture_manager_v1;
+
+namespace core { class Region; class Point; }
```
CendioOssman: clang gets upset about `class Point` when it is in fact `struct Point`

CendioOssman (changes requested):

unix/w0vncserver/wayland/objects/ImageCopyCaptureManager.h
CendioHalim: Whoops, fixed :^)

PR #2074, opened 2026-03-09
Make desktop resizing interface async

CendioHalim:
This change makes it possible to implement an SDesktop that resizes asynchronously.

The current implementation only allows one resize to be happening at the same time. Requests that are received while a resize is already happening will be denied.

This commit is done in preparation for adding resize support to the WaylandDesktop.

Leaving this as draft since the timeout handling isn't robust, and the interface needs some slight changes:  the `VNCServer` probably shouldn't have `setScreenLayoutDone()` be public.

### DIFF
--- a/common/rfb/SDesktop.h
+++ b/common/rfb/SDesktop.h
@@ -89,11 +89,9 @@ namespace rfb {
 
     // setScreenLayout() requests to reconfigure the framebuffer and/or
     // the layout of screens.
-    virtual unsigned int setScreenLayout(int /*fb_width*/,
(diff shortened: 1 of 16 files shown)


CendioOssman:
Having `setScreenLayoutDone()` be part of the public interface is the correct approach. Compare it with `approveConnection()`.

What I'm more concerned about is how it coexists with `setPixelBuffer()` and `setScreenLayout()`. I think it would be more robust if we could make sure that is atomic (i.e. a single call), rather than multiple calls.

PR #2075, opened 2026-03-10
fix: vncserver xinit invocation

step-:
**Issue Description:** For an xsession .desktop file containing:

```
[Desktop Entry]
Exec=/etc/X11/xinitrc jwm
```

the xinit invocation was (wrong):

    xinit /etc/X11/Xsession '/etc/X11/xinitrc jwm' -- ...

instead of (correct):

    xinit /etc/X11/Xsession /etc/X11/xinitrc jwm -- ...

according to the xinit manual page (with omissions):

```
SYNOPSIS
       xinit [[client] options ...] [ -- [server] [display] options ...]

       An alternate client and/or server may be specified on the command line.
       The  desired  client  program  and its arguments should be given as the
       first command line arguments to xinit.  To specify a particular  server
       command  line, append a double dash (--) to the xinit command line (af‐
       ter any client and arguments) followed by the desired server command.

       Both the client program name and the server  program  name  must  begin
       with  a  slash  (/) or a period (.).

X Version 11                      xinit 1.4.2                         XINIT(1)
```

### DIFF
--- a/unix/vncserver/vncserver.in
+++ b/unix/vncserver/vncserver.in
@@ -242,7 +242,10 @@ $ENV{XAUTHORITY} = $xauthorityFile;
 
 @cmd = ("xinit");
 
-push(@cmd, $Xsession, $session{'Exec'});
+push(@cmd, $Xsession);
+foreach my $word (split(/\s+/, $session{'Exec'})) {
+  push(@cmd, $word);
+}
 
 push(@cmd, '--');
 


CendioOssman:
I'm afraid this change would break other setups. Other `Xsession` implementations expect the entire command line in the first argument, and not as a series of arguments.

What `Xsession` implementation do you have in this instance?

I'm also curious about your use of `xinitrc`. That is something that should be run implicitly, e.g. by `Xsession`. It is not part of the desktop environment definition, like in your example. I don't think it is standard that it can be executed with a command like that.

step-:
> I'm afraid this change would break other setups. Other `Xsession` implementations expect the entire command line in the first argument, and not as a series of arguments.

In major distros your mean? I don't know. Perhaps they don't even pass arguments hence they have not been affected.

> What `Xsession` implementation do you have in this instance?

None of the major distros'. Mine is a Linux From Scratch system with minimal Desktop Environment.
 
> I'm also curious about your use of `xinitrc`. That is something that should be run implicitly, e.g. by `Xsession`. It is not part of the desktop environment definition, like in your example. I don't think it is standard that it can be executed with a command like that.

- The command above was generated by vncserver spawned by vncsession. I replaced the rest of the line concerning /usr/bin/Xvnc arguments with ellipsis to be concise. Here's the unedited line:
  ```
   xinit /etc/X11/Xsession /etc/X11/xinitrc jwm -- /usr/bin/Xvnc :1 -auth /root/.Xauthority -desktop 'ipc:1 (root)' -fp /usr/share/fonts/X11//misc,/usr/share/fonts/X11//75dpi,/usr/share/fonts/X11//100dpi,/usr/share/fonts/X11//Type1 -pn -rfbauth /root/.config/tigervnc/passwd -rfbport 5901
  ```
- I included excerpts from xinit's manpage. The way vncserver.in constructed the xinit command line violates xinit's synopsis.

CendioOssman:
Xsession is basically always used with an argument. One major implementation is GDM's, which only supports a single argument.

I don't think there's a formal standard for Xsession. But if you write your own, it will need to be compatible with the de facto standard. Which would be set by things like GDM.

My point about `xinitrc` was that it doesn't seem to belong in a .desktop file. It's part of the infrastructure, and not specific to just one single desktop. So I think you should just drop it and do `exec=jwm`. If you look at e.g. MATE, it's just `Exec=mate-session`.

And to be clear, we are not violating xinit's usage. We are giving it a command (Xsession) and the arguments that command expects.

step-:
Thank you for your comments.

> Xsession is basically always used with an argument.

Well, not in the distribution I'm using.

> One major implementation is GDM's, which only supports a single argument.

OK.

> I don't think there's a formal standard for Xsession. But if you write your own, it will need to be compatible with the de facto standard. Which would be set by things like GDM.

OK. I looked up how GDM sources the session command from .desktop files that specify the `Exec=` command: `gnome-session`. That's similar to TigerVNC vncserver script, fetching the `Exec=` command from .desktop files in /usr/share/xsessions. A user can select one such file by writing `session=<filename>` for `/usr/share/xsessions/<filename>.desktop`.

I'm not aware that, in an `Exec=<command>` line, `<command>` must be a single word. I think you are saying that it is, a de-facto standard,because GDM expects a single word (`gnome-session`). I understand your point.

However, this commit isn't concerned with Xsession. It's concerned with the way the vncserver script passes arguments to the Xsession file. It is a defacto standard on \*nix systems that programs receive arguments as separate list elements (argc, arv[]), not as a single string concatenating all elements together. This is what this commit is about.

> My point about xinitrc was that it doesn't seem to belong in a .desktop file. It's part of the infrastructure, and not specific to just one single desktop. So I think you should just drop it and do exec=jwm. If you look at e.g. MATE, it's just Exec=mate-session.

Perhaps my original message wasn't clear enough, and you thought that I'm dealing with a .desktop file in ~/.local/share/applications? No, I'm dealing with a system .desktop file in /usr/share/xsessions, as I explained in more detail above. These are pre-existing xsession files I cannot edit.

> And to be clear, we are not violating xinit's usage. We are giving it a command (Xsession) and the arguments that command expects.

I agree. However, you are also passing the arguments for Xsession as a single string, concatenating whatever is on the `Exec=` line, and assuming that either there is only one argument (like for GDM), or that the Xsession script will parse the string to split it into the actual argument list.

Anyway, taking a step back, I can understand your hesitation about this patch. After all, no one has complained before me in six years ref c67a5f2, and changing things at this point is risky. The patch is available, I can use it for my own rig, and if anyone else will run into this problem, they can find the patch here.

CendioOssman:
> I'm not aware that, in an `Exec=<command>` line, `<command>` must be a single word. I think you are saying that it is, a de-facto standard,because GDM expects a single word (`gnome-session`). I understand your point.

There can be multiple words in there. In fact, GNOME relies on that fact for its "classic" mode:

```
Exec=env GNOME_SHELL_SESSION_MODE=classic gnome-session
```

As you can see. It's not just multiple arguments. It's a shell expression.

> I agree. However, you are also passing the arguments for Xsession as a single string, concatenating whatever is on the Exec= line, and assuming that either there is only one argument (like for GDM), or that the Xsession script will parse the string to split it into the actual argument list.

Xsession will indeed parse that string to do what's needed. Or rather, it delegates it to the shell. E.g.

```
/bin/sh -c "exec -l $SHELL -c \"$1\""
```

The proposed patch would break this, as everything in $2 and onwards would just be dropped.

step-:
> There can be multiple words in there. In fact, GNOME relies on that fact for its "classic" mode:
>
>     Exec=env GNOME_SHELL_SESSION_MODE=classic gnome-session
>As you can see. It's not just multiple arguments. It's a shell expression.

> Xsession will indeed parse that string to do what's needed. Or rather, it delegates it to the shell. E.g.
>
>     /bin/sh -c "exec -l $SHELL -c \"$1\""
> The proposed patch would break this, as everything in $2 and onwards would just be dropped.

Absolutely, I agree. To which I add that there is a specification for how arguments within `$1` are to be quoted (spoiler: within escaped double quotes):
https://specifications.freedesktop.org/desktop-entry/latest/exec-variables.html

The quoting rules go hand-in-hand with TigerVNC's implementation and the shell-based parsing mechanism.  Thank you for taking the time to discuss all my points. I will withdraw this PR.

PR #2078, opened 2026-03-23
w0vncserver: Memory related cleanups

CendioHalim:


unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -306,15 +311,17 @@ void Clipboard::selectionRead()
...
     return;
   }
 
   if (!g_variant_is_of_type(response, G_VARIANT_TYPE("(h)"))) {
     g_object_unref(fdList);
     vlog.error("Could not read clipboard: invalid response type: %s, "
                "expected (h)", g_variant_get_type_string(response));
-    readInProgress = false;
+    delete readContext;
+    readContext = nullptr;;
```
CendioOssman: Double ; in a couple of places here.

unix/w0vncserver/portals/Clipboard.cxx
```diff
@@ -531,8 +545,12 @@ void Clipboard::handleReadDataCallback(GAsyncResult* res)
   g_input_stream_read_bytes_async(
     readStream, 4096, G_PRIORITY_DEFAULT, readCancellable,
     [](GObject*, GAsyncResult* result, void* userData) {
-      ((Clipboard*)userData)->handleReadDataCallback(result);
-    }, this);
+      ReadContext* ctx = (ReadContext*)userData;
+      if (ctx->isAlive)
```
CendioOssman, line 557: `isAlive` is never changed. Is this change really relevant for `Clipboard`?

unix/w0vncserver/portals/RemoteDesktop.cxx
```diff
@@ -523,9 +523,8 @@ void RemoteDesktop::handleStart(GVariant* parameters)
 
   streams = g_variant_lookup_value(result, "streams",
                                     G_VARIANT_TYPE_ARRAY);
-  g_variant_unref(result);
```
CendioOssman, line 526: Not sure the commit message matches the change. This looks like a use-after-free issue?

CendioOssman (changes requested):

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: Fixed

unix/w0vncserver/portals/Clipboard.cxx
CendioHalim: I missed to sync all changes from `DataControl` to `Clipboard`, they should be identical now. Tested this by reading 10 bytes at a time with a 1s sleep and could verify with valgrind + debug prints that cancellation and clean-up works

unix/w0vncserver/portals/RemoteDesktop.cxx
CendioHalim: You're right, fixed.

CendioOssman (approved):

PR #2105, opened 2026-05-21
Improve reproducibility of package builds

hyder365:
This makes two small changes that help downstream distributions produce reproducible TigerVNC packages.

- Add a `JAVA_SIGN_JAR` CMake option, defaulting to enabled, so release builds keep the current behavior while distro/package builds can disable signing of the Java viewer jar.
- Replace `__DATE__`/`__TIME__` in native build-time strings with a deterministic Unix epoch timestamp.

The automatically generated/self-signed jar signature and compile-time date/time macros otherwise embed data that can vary between builds.

### DIFF
--- a/unix/vncconfig/buildtime.c
+++ b/unix/vncconfig/buildtime.c
@@ -15,4 +15,9 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
  * USA.
  */
-char buildtime[] = __DATE__ " " __TIME__;
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+char buildtime[] = BUILD_TIMESTAMP;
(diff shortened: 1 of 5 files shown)


unix/vncconfig/buildtime.c
```diff
@@ -15,4 +15,4 @@
  * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
  * USA.
  */
-char buildtime[] = __DATE__ " " __TIME__;
+char buildtime[] = "Jan  1 1970 00:00:00";
```
CendioOssman: This just completely breaks this function and just displays this static value. That's hardly a fix.

CendioOssman (changes requested):

unix/vncconfig/buildtime.c
hyder365: How would you like it to be done?

CendioOssman: You're the one with the feature request, so you'll need to be clear what you need. Just yanking out stuff that's in the way isn't a productive way forward. You'll need to find where there is an overlap with the existing functionality and what you need.

`SOURCE_DATE_EPOCH` might be something to use.

hyder365:
I've adjusted it to use SOURCE_DATE_EPOCH instead—should handle both the reproducible builds case and preserve the actual timestamp when it's not set.

CendioOssman:
Is this vibe coded? Because this isn't even close to the proper fix. The patch now adds the check in runtime, rather than build time.

There should be examples out there of how to use this value properly.

hyder365:
Thanks for reviewing. I'm honestly not an expert at this, but just trying to help further reproducible builds in various distros. How does this version look?

CendioOssman:
Thanks. This looks much more reasonable. Although there seems to be one instance left in `win/winvnc/buildTime.cxx`.

@bphinz, what are your thoughts on the signing changes?

hyder365:
> Although there seems to be one instance left in `win/winvnc/buildTime.cxx`.

Okay, I've put a change there too.

bphinz:
Looks good to me

CendioOssman (approved):

PR #2111, opened 2026-05-31
Minor compatibility fixes for macOS

### LINKED ISSUE #2110: macOS: unconditional usage of NSRunningApplication breaks build on 10.5
Recent commit https://github.com/TigerVNC/tigervnc/commit/a4b444806a8e5e8ea957038ee4767ab051d98ee5 has broken the build on 10.5. Adding that code could be done conditionally, and then the app will work on Leopard.

barracuda156:
1. Fix https://github.com/TigerVNC/tigervnc/issues/2110
2. Fix wrong macro for 10.7.
3. Include `AvailabilityMacros.h` explicitly to ensure consistency, use numerical macros for comparison uniformly.

### DIFF
--- a/vncviewer/KeyboardMacOS.mm
+++ b/vncviewer/KeyboardMacOS.mm
@@ -29,9 +29,18 @@
 
 #include <IOKit/hidsystem/IOHIDLib.h>
 #include <IOKit/hidsystem/IOHIDParameter.h>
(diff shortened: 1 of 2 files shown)


CendioOssman:
As mentioned on #2110, I'm afraid we will not be supporting such old versions of macOS.

PR #2113, opened 2026-05-31
Work around macOS NSWindowStyleMaskBorderless

derekschrock:
When FLTK puts a window fullscreen on macOS it has the style of
NSWindowStyleMaskBorderless.  This disables the minimize feature of the
window.  Masking NSWindowStyleMaskMiniaturizable allows the boarder less
window to be minimize.

This should be fixed in FLTK however given that we're pinned to 1.13.x
only security fixes will be accepted upstream?  Possible 1.14 or 1.15 do
not have this issue.

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -1007,7 +1007,7 @@ void Viewport::popupContextMenu()
     // FIXME: Workaround for not being able to minimize in fullscreen
     // https://github.com/TigerVNC/tigervnc/pull/1813
     if (window()->fullscreen_active())
-      window()->fullscreen_off();
+      cocoa_enable_minimize(window());
 #endif
     window()->iconize();
     break;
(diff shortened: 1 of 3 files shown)


vncviewer/cocoa.h
```diff
@@ -37,4 +37,6 @@ CGColorSpaceRef cocoa_win_color_space(Fl_Window *win);
 bool cocoa_win_is_zoomed(Fl_Window *win);
 void cocoa_win_zoom(Fl_Window *win);
 
+void cocoa_minimize(Fl_Window *win);
```
CendioOssman: This sounds like it will actually do the minimize operation. Perhaps something like  `cocoa_enable_minimize()`?

vncviewer/Viewport.cxx
```diff
@@ -1004,10 +1004,8 @@ void Viewport::popupContextMenu()
     break;
   case ID_MINIMIZE:
 #ifdef __APPLE__
-    // FIXME: Workaround for not being able to minimize in fullscreen
```
CendioOssman, line 1007: Might want to keep some comment on why this magic is necessary.

CendioOssman (changes requested):
Nice find!

Some minor comments, then I can go ahead and merge this fix.

CendioOssman (approved):

CendioOssman:
Sorry for the delay in retesting this. Works well on macOS 26 here.

Nice work!

PR #2116, opened 2026-06-03
Avoid re-registering notification for deleted client.

0-wiz-0:
After disconnecting a client, the log file quickly filled up with lots of copies this line:
```
 XserverDesktop: Cannot find file descriptor for socket event
```

A new block was introduced with https://github.com/TigerVNC/tigervnc/commit/9198ccd1b28d4e05b9ede87da386adc64499e241 where `vncSetNotifyFd(fd, screenIndex, true, (*i)->outStream().hasBufferedData());` is called after `delete(*i);`


CendioOssman:
Thanks! No idea why I cannot see the problem here, but your diagnosis and fix looks correct.

PR #2124, opened 2026-06-29
Fix protocol desynchronization in readSetDesktopSize

hyder365:
readSetDesktopSize() returned true instead of false when insufficient data was available, unlike every other message handler in SMsgReader. This caused the readMsg() state machine to reset to MSGSTATE_IDLE, leading to message boundary desynchronization — the first byte of subsequent data was consumed as a new message type rather than part of the pending SetDesktopSize message.


CendioOssman (approved):
Good catch! Thanks!

PR #2125, opened 2026-06-29
Prevent PAM error message disclosure to VNC clients

hyder365:
The PAM conversation callback was forwarding PAM_ERROR_MSG and PAM_TEXT_INFO messages verbatim to unauthenticated clients. Depending on PAM configuration, this could reveal whether a username exists (e.g. 'User not known' vs 'Authentication failure'), enabling username enumeration without valid credentials.


CendioOssman:
That would be a bug in your PAM modules. The messages they present are supposed to be shown to the connecting user. Compare with how sshd behaves.

PR #2126, opened 2026-06-29
Reject unsolicited clipboard provides from clients

hyder365:
Track whether a clipboard request is pending and reject clipboardProvide messages that were not explicitly requested. An authenticated client could previously push arbitrary clipboard content that would reach the desktop clipboard without any request, enabling clipboard spoofing attacks against other connected clients.\n\nAlso fixes clipboard store-before-check by moving the access rights check before data storage in both clientCutText and handleClipboardProvide paths.


CendioOssman:
A malicious client can do that anyway by pushing announce messages. So I'm not sure what the practical difference is here.

What scenario are you targeting that this PR would improve?

hyder365:
You're right, a malicious client can just use clipboardNotify to trigger the request and still get its provide accepted. This only blocks the direct push, which isn't much of a win.

PR #2127, opened 2026-06-29
Rate-limit framebuffer resize requests to prevent memory DoS

hyder365:
An authenticated client could request the maximum framebuffer size (16384x16384) repeatedly without any rate limiting, triggering ~1 GiB memory allocations each time and eventually causing the server to run out of memory.\n\nAdd a 1-second minimum interval between SetDesktopSize requests to prevent rapid memory exhaustion attacks.


CendioOssman (changes requested):
We very much do not want to rate limit this. We recently improved on the responsiveness in these resizes.

It should not be possible to accumulate an indefinite number of framebuffers. That sounds like the real bug here.

Have you investigated where that leak comes from?

hyder365:
You're right, I see the old framebuffer gets freed at xvnc.c:619. There's no accumulation. I'll close this.

PR #2132, opened 2026-07-07
TightDecoder: bound rectangle area to prevent OOM from a malicious server

acts-1631:
Tight rectangle dimensions are controlled by the server. The existing width cap permits a rectangle up to 2048x65535, causing `decodeRect()` to allocate a very large decompression buffer and, when pixel translation is required, another large temporary buffer.

A fixed pixel-area limit would reject legitimate narrow, tall rectangles from other servers. Instead, this limits decoded rectangle storage to 32 MiB, matching `BufferedOutStream`. The limit is calculated from the rectangle dimensions and negotiated pixel format, which conservatively bounds both decompression and translation buffers while retaining support for tall, low-memory updates.

### DIFF
--- a/common/rfb/TightDecoder.cxx
+++ b/common/rfb/TightDecoder.cxx
@@ -46,6 +46,10 @@ using namespace rfb;
 static const int TIGHT_MAX_WIDTH = 2048;
 static const int TIGHT_MIN_TO_COMPRESS = 12;
 
+// Keep temporary decompression and pixel translation buffers bounded.
+// This matches the size limit used by BufferedOutStream.
+static const uint64_t TIGHT_MAX_DATA_SIZE = 32 * 1024 * 1024;
+
 TightDecoder::TightDecoder()
   : Decoder(DecoderPartiallyOrdered), readState(IDLE)
 {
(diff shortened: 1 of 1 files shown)


CendioOssman (changes requested):
I'm afraid this will not work. TigerVNC needs to be compatible with other servers as well, which might send larger rectangles.

Some kind of sanity limit is probably appropriate, though. `BufferedOutStream` has a limit of 32 MiB. Perhaps we should set the same for the `MemOutStream` used here.

acts-1631:
Updated the PR to remove the fixed area limit. It now limits decoded rectangle memory to 32 MiB based on the rectangle dimensions and negotiated pixel format. This preserves narrow, tall rectangles while bounding the decompression and pixel-translation allocations.

CendioOssman (changes requested):
This would only prevent massive allocations for Tight rectangles. The server could just switch to another encoding and do the same thing.

So the check needs to be somewhere more general, which is why I thought `MemOutStream` might be a candidate.

CendioOssman:
Hmm... We have a maximum screen size of 16834×16834. Have you actually tested that a rectangle height of 2048×65535 is accepted? It would be outside the framebuffer and should ideally be dropped.

acts-1631:
You’re right, I missed the bounds check in `CMsgReader`. That rectangle is rejected before it reaches `TightDecoder`. I’ll leave this closed, since any limit needs to apply generally rather than only to Tight.

Issue #2129, opened 2026-07-01
copy from client to server via selection fails if clipboard does not contain text


gsabdiv:
**Describe the bug**
Between tigervnc-1.13.1 and tigervnc-1.15.0 some toolkit got introduced for handling copy&paste and it came along with some bug

**To Reproduce**
1. Be sure the clipboard is empty, for example by doing this right after logging in
2. vncviewer -Log *:stderr:100 to some Server (TigerVNC on RHEL 8 in my case)
3. select some plain text on client side, in my case in an xterm
4. See "Got non-plain text in local clipboard, ignoring." being output from the client

**Expected behavior**
Selection is being copied to server

**Screenshots**

**Client (please complete the following information):**
 - OS: RedHat Entperprise Linux 8.10
 - VNC client: TigerVNC
 - VNC client version: 1.15.0
 - Client downloaded from: Supplied by RedHat AppStream repository

**Server (please complete the following information):**
 - OS: RedHat Entperprise Linux 8.10
 - VNC server: TigerVNC
 - VNC server version: 1.15.0 (but shouldn't matter)
 - Server downloaded from: Supplied by RedHat AppStream repository
 - Server was started using: vncserver (but doesn't matter)

**Additional context**
This probably is what was experienced in #2060, but got closed due to a lack of being able to reproduce it.
Bug still exists in 1.16.2 (downloaded from git and compiled by myself)
Viewport::handleClipboardChange() does a "if (!Fl::clipboard_contains(Fl::clipboard_plain_text))", which apparently causes a problem if the clipboard is empty. A workaround: Ctrl+middle_click the xterm and chose "select to clipboard", then select something and revert it to not serlect to clipboard. From now on copy&paste works fine for selection - but maybe it breaks by putting binary stuff into the clipboard. I do not have a suggested patch, since I don't know who this toolkit is supposed to handle the two ways of doing copy and paste. Everything worked fine up to viewer 1.13.something.

CendioOssman:
Yeah, that's a bug. We should only do that check for the regular clipboard, not selections. Thanks for reporting!

grulja:
Hi @gsabdiv, can you please give #2135 a try?

PR #2134, opened 2026-07-25
Defer clipboard inspection until focused

### LINKED ISSUE #2128: Windows viewer opens the local clipboard on every clipboard change even when unfocused, causing intermittent clipboard contention with other apps
**Describe the bug**
With at least one viewer running, the **Windows** clipboard becomes unreliable in an intermittent way:

- Copying an image from a third-party tool (e.g. Flameshot) with Ctrl+C sometimes fails to place anything on the clipboard: the item also never appears in the Windows clipboard history (Win+V).
- Copying text in another Windows app (e.g. a browser) and then pasting it into the viewer sometimes does nothing.

Both happen only when one or more viewers are running, and the probability of failure increases with the number of viewers. With no viewer running, the Windows clipboard works reliably.

 - OS: Windows 11
 - VNC TigerVNC 1.16.2

 - OS: Oracle linux 9
 - VNC server: TigerVNC XVnc
 - VNC server version: 1.15.0
 - Server downloaded from: OL9 repo
 - Server was started using: XVnc

**Additional context**

_Warning : I asked Claude to help me with this analysis_

The Windows clipboard is a single-owner resource: only one process may hold it open via `OpenClipboard()` at a time; concurrent `OpenClipboard()` calls fail with `ERROR_ACCESS_DENIED`.

In the constructor, each `Viewport` registers a clipboard listener:
https://github.com/TigerVNC/tigervnc/blob/v1.16.2/vncviewer/Viewport.cxx#L112
```
Fl::add_clipboard_notify(handleClipboardChange, this);
```

On Windows, FLTK implements this with `AddClipboardFormatListener`, so every viewer receives `WM_CLIPBOARDUPDATE` whenever *any* application changes the clipboard — regardless of focus. The notify callback then inspects the clipboard:
https://github.com/TigerVNC/tigervnc/blob/v1.16.2/vncviewer/Viewport.cxx#L598
```
void Viewport::handleClipboardChange(int source, void *data)
{
...
  if (!Fl::clipboard_contains(Fl::clipboard_plain_text)) {
    vlog.debug("Got non-plain text in local clipboard, ignoring.");
    // Reset the state as if we don't have any clipboard data at all
    self->pendingClientClipboard = false;
   ...
  if (!self->hasFocus()) {
    vlog.debug("Local clipboard changed whilst not focused, will notify server later");
    self->pendingClientClipboard = true;
    // Clear any older client clipboard from the server
    try {
      self->cc->announceClipboard(false);
    } catch (std::exception& e) {
      vlog.error("%s", e.what());
      abort_connection_with_unexpected_error(e);
    }
    return;
  }
```
FLTK opens the clipboard:
https://github.com/fltk/fltk/blob/release-1.4.5/src/Fl.cxx#L2201-L2204
```
int Fl::clipboard_contains(const char *type)
{
  return Fl::screen_driver()->clipboard_contains(type);
}
```
https://github.com/fltk/fltk/blob/release-1.4.5/src/Fl_win32.cxx#L933-L944
```
int Fl_WinAPI_Screen_Driver::clipboard_contains(const char *type) {
  int retval = 0;
  if (!OpenClipboard(NULL))
    return 0;
  if (strcmp(type, Fl::clipboard_plain_text) == 0 || type[0] == 0) {
    retval = IsClipboardFormatAvailable(CF_UNICODETEXT);
  } else if (strcmp(type, Fl::clipboard_image) == 0) {
    retval = IsClipboardFormatAvailable(CF_DIB) || IsClipboardFormatAvailable(CF_ENHMETAFILE);
  }
  CloseClipboard();
  return retval;
}
```

`Fl::clipboard_contains()` calls `OpenClipboard()` under the hood. Crucially, this happens **before** the focus check:
https://github.com/TigerVNC/tigervnc/blob/v1.16.2/vncviewer/Viewport.cxx#L613

So on every clipboard change, *every running viewer* (focused or not, one per process) immediately tries to open the Windows clipboard. This races against:
- the application that just wrote the clipboard (which may still hold it, or use delayed rendering / `WM_RENDERFORMAT`), and
- the Windows Clipboard History service (itself a listener that re-opens the clipboard to snapshot it into Win+V).

When the viewers win the race, the writer or the history service gets `ERROR_ACCESS_DENIED`, so the data is lost or never enters Win+V (image case). When a viewer loses the race, `Fl::clipboard_contains()` returns `false`, the callback takes the "non-plain text → ignoring" branch and announces no clipboard to the server, so a later paste into the viewer does nothing (text case). The intermittency is the timing race; the scaling with viewer count is because each viewer is an independent contender.

**Proposition**
Notably, `announceClipboard(true)` is **already** deferred to focus regain via `pendingClientClipboard` / `flushPendingClipboard()`:
https://github.com/TigerVNC/tigervnc/blob/v1.16.2/vncviewer/Viewport.cxx#L636
https://github.com/TigerVNC/tigervnc/blob/v1.16.2/vncviewer/Viewport.cxx#L519

The only thing that currently happens at change-time while unfocused is the `OpenClipboard()` from `Fl::clipboard_contains()`. So the fix is to **defer that inspection to focus regain too**: when the viewer does not have focus, just set the pending flag and return *without opening the clipboard*; perform `Fl::clipboard_contains()` (and the corresponding `announceClipboard()`) inside `flushPendingClipboard()` when focus is regained — at which point the writing application has long released the clipboard.

Sketch (Windows-scoped, to avoid changing X11 primary-selection behavior):

`handleClipboardChange()`:
```cpp
#if defined(WIN32)
  if (!self->hasFocus()) {
    // Do NOT open the clipboard here — it races with whatever app
    // just wrote it (and with the clipboard history service).
    // Defer inspection until we regain focus.
    self->pendingClientClipboard = true;
    return;
  }
#endif

  if (!Fl::clipboard_contains(Fl::clipboard_plain_text)) {
    ... announce(false); return;
  }
  self->clipboardSource = source;
  ... // focused: inspect + announce as today
```

`flushPendingClipboard()`:
```cpp
void Viewport::flushPendingClipboard()
{
  if (pendingClientClipboard) {
    if (Fl::clipboard_contains(Fl::clipboard_plain_text)) {
      try { cc->announceClipboard(true); } catch (...) { ... }
    } else {
      try { cc->announceClipboard(false); } catch (...) { ... }
    }
  }
  pendingClientClipboard = false;
}
```

This is behavior-preserving from the server's perspective (it only ever learns about the client clipboard at focus time anyway) and removes the unfocused `OpenClipboard()` calls entirely. In the failing workflows above (screenshot tool, browser copy) no viewer is focused, so the viewer-induced contention drops to zero.

gpertea:
## Summary

Defer plain-text clipboard inspection while the viewer is unfocused. TigerVNC already delays advertising the client clipboard until focus returns; this change also delays `Fl::clipboard_contains()` until that point.

This avoids unnecessary platform clipboard access from every unfocused viewer and prevents a transient failed format check from immediately being treated as a non-text clipboard.

## Scope

This addresses the unfocused clipboard-inspection and contention path described in #2128. It does not change how FLTK delivers clipboard-change notifications, retry failed clipboard access, or recover notifications lost by the Windows clipboard-viewer chain.

The `source != 0` check is retained, preserving the PRIMARY-selection behavior fixed by #2129.

Addresses #2128.

### DIFF
--- a/vncviewer/Viewport.cxx
+++ b/vncviewer/Viewport.cxx
@@ -595,7 +595,22 @@ void Viewport::handleClipboardChange(int source, void *data)
     return;
 #endif
 
-  if (source != 0 && !Fl::clipboard_contains(Fl::clipboard_plain_text)) {
+  if (!self->hasFocus()) {
+    vlog.debug("Local clipboard changed whilst not focused, will notify server later");
+    self->clipboardSource = source;
+    self->pendingClientClipboard = true;
+    // Clear any older client clipboard from the server
+    try {
+      self->cc->announceClipboard(false);
+    } catch (std::exception& e) {
+      vlog.error("%s", e.what());
+      abort_connection_with_unexpected_error(e);
+    }
+    return;
+  }
+
+  if (source != 0 &&
+      !Fl::clipboard_contains(Fl::clipboard_plain_text)) {
     vlog.debug("Got non-plain text in local clipboard, ignoring.");
     // Reset the state as if we don't have any clipboard data at all
     self->pendingClientClipboard = false;
@@ -610,19 +625,6 @@ void Viewport::handleClipboardChange(int source, void *data)
 
   self->clipboardSource = source;
 
-  if (!self->hasFocus()) {
-    vlog.debug("Local clipboard changed whilst not focused, will notify server later");
-    self->pendingClientClipboard = true;
-    // Clear any older client clipboard from the server
-    try {
-      self->cc->announceClipboard(false);
-    } catch (std::exception& e) {
-      vlog.error("%s", e.what());
-      abort_connection_with_unexpected_error(e);
-    }
-    return;
-  }
-
   vlog.debug("Local clipboard changed, notifying server");
   try {
     self->cc->announceClipboard(true);
@@ -636,6 +638,13 @@ void Viewport::handleClipboardChange(int source, void *data)
 void Viewport::flushPendingClipboard()
 {
   if (pendingClientClipboard) {
+    if (clipboardSource != 0 &&
+        !Fl::clipboard_contains(Fl::clipboard_plain_text)) {
+      vlog.debug("Pending local clipboard has no plain text, ignoring.");
+      pendingClientClipboard = false;
+      return;
+    }
+
     vlog.debug("Focus regained after local clipboard change, notifying server");
     try {
       cc->announceClipboard(true);


CendioOssman (changes requested):
Thank you for your suggestion.

I think we just add needless complexity and confusion by making this Windows-only. Please adjust the commit so that this behaviour is the standard one.

gpertea:
Thanks, agreed.  I've removed the Windows-specific branch, so clipboard inspection is now deferred consistently across platforms. I also rebased onto current `master` while preserving the new PRIMARY-selection handling from #2129. Tested on Windows and headless X11/Linux.

gpertea:
Please hold off on reviewing this for now - I've converted the PR back to draft.

Further live testing exposed a remaining Windows clipboard failure. The patch addresses clipboard inspection/contention after a notification is received, but it does not address loss of notifications caused by FLTK’s Windows clipboard-viewer chain. I need to separate those behaviors and reassess the scope before requesting review again.

gpertea:
Further testing confirmed that this patch has a narrower scope than I initially thought: it avoids immediate clipboard inspection by an unfocused viewer, reducing contention and preventing a transient failed check from discarding a notification that was received.

However it does not address missed Windows clipboard notifications caused by FLTK's own clipboard-viewer chain. I've updated the PR description accordingly, rebased onto current master, preserved the #2129 behavior and tightened the clipboardSource assignment. So this is intended as a targeted improvement, not a complete fix for Windows clipboard reliability, which requires a platform specific fix in FLTK.

CendioOssman (approved):
Looks good. Thanks for your work here!

PR #2136, opened 2026-08-02
Allow building with MSVC

jose-pr:
This allows TigerVNC to be built with MSVC. MSVC lacks a number of POSIX headers and functions that the sources rely on. These are provided in `contrib/msvc/include`, which is only used for MSVC. That way the normal sources don't need to know about the compiler. MinGW remains the supported toolchain on Windows.

The series:

1. **Remove obsolete NOMINMAX definition** — unused since e0e4f4ca, and the block referred to the `rfb` target before it existed.
2. **Put WINAPI inside the SendSAS function pointer type** — the calling convention belongs inside the parentheses; gcc accepts it outside, MSVC doesn't.
3. **Use a private copy of CLSID_VideoProcessorMFT** — whether a header declares the GUID depends on the toolchain and the targeted Windows version. The Windows SDK only declares it for Windows 8.1 and newer, whilst current MinGW always does. The decoder now keeps its own copy under a name that can't clash, so there is nothing left to detect.
4. **Share hook state with MSVC as well** — MSVC has no per-variable equivalent of `__attribute__((section("shared"), shared))`, so the variables are grouped in a data segment that the linker is told to share.
5. **Set NDEBUG and _DEBUG with definitions** — `add_definitions()` rather than editing the per-configuration flags, which works the same for both compilers.
6. **Select compiler flags separately for MSVC** — each section of gcc flags gets its MSVC equivalent (`/std:c11`, `/std:c++14`, `/W3`, `/WX`), or is skipped.
7. **Keep preprocessor directives out of the usage text** — translation tools don't run the preprocessor, so only the Unix form of the text was extracted and the Windows text could never be translated. MSVC also rejects directives inside macro arguments. The whole format string is now selected per platform. The output is unchanged on both.
8. **Build with MSVC using compatibility headers** — the include directory, a force-included `msvc_compat.h`, and a short README.
9. **Add manually triggered build with MSVC** — `.github/workflows/msvc.yml`, started from the Actions tab. It builds with vcpkg for x64, with translations but without GnuTLS. The dependencies take close to an hour to build without a cache, so it doesn't run on every push and pull request.

### Testing

Locally on Windows ARM64, with MSVC 14.44 and MSYS2 clang (MinGW):

- Each commit builds with MinGW.
- The whole tree, including the viewer, builds in Debug and Release with both compilers.
- All unit tests pass in each configuration.
- The viewer negotiates and plays audio with both compilers.

Debug builds were run with the existing warnings made non-fatal.

The MSVC job passes on my fork, including the unit tests: https://github.com/jose-pr/tigervnc/actions/runs/34989583759

Note that hook state is still not shared when building with clang, as clang ignores the `shared` attribute. That was already the case before this PR.


CendioOssman:
I'm afraid that I'm extremely sceptical about merging this. This is a very invasive change that affects a lot of files. For a compiler that is not used by any of the developers.

You say that this is primarily for ARM. What is the current state of MinGW there? I saw that gcc got support a while ago. So I would hope that the latest versions would work just fine.

jose-pr:
Fair, and thanks for looking at it properly rather than just closing it.

First, the CI failure on this PR is mine and it is a straightforward bug, not
something inherent to the approach. `common/core/os.h` guarded its POSIX shims
on `_WIN32` when they are MSVC-only, so mingw picked up replacements for things
it already provides — `mode_t` (which it declares as `unsigned short`, against
my `int`), plus `strcasecmp`, `gettimeofday` and `dirname`. That broke the
existing mingw build at `common/core/Logger_file.cxx`, which is worse than my
own PR not working. The `ffs()` shim right beside them already had the correct
`_MSC_VER` guard for exactly this reason; I just didn't apply it to its
neighbours. Fixed in the latest push, and verified against a real mingw build
rather than by inspection this time.

On mingw for ARM: you're right, and I should not overstate this. MSYS2's
`CLANGARM64` environment works today — that is in fact what I used to test the
above, on an ARM64 Windows box. It is clang rather than gcc, but "mingw doesn't
work on ARM" is not a claim I can make and I'm not going to pretend otherwise.

The main reason is a preference, not a necessity: I would rather build on
Windows with the toolchain the OS itself ships and officially supports. The
practical part of that is not having to wait on mingw to catch up — for a
future architecture, or for a Windows API that lands in the SDK first — and
having the option of integrating more deeply with the platform if that is ever
wanted. But mostly it is just wanting to use the native compiler where the
platform has one.

That is not an argument that anything is broken today, and I recognise it is a
weak reason to ask you to carry an invasive change for a compiler none of you
use. If the answer is that the maintenance cost outweighs a contributor's
toolchain preference, that is completely reasonable and I won't argue it.

If there is a version of this you would consider, I would guess it is a much
smaller one: keep `common/core/os.h` and the genuinely portable fixes, drop
everything that only exists to satisfy MSVC, and leave the build files alone —
so the tree becomes *more* portable without the project taking on a second
Windows toolchain. Happy to cut it down to that, or to close this if you would
rather not have the surface area at all. Either is fine; I'd just rather know
which before spending more time on it.

CendioOssman:
One idea that might be less intrusive is if you could provide some transparent wrapper headers that hide the differences. I.e. a `contrib/msvc/include` with the missing headers and functionality. We could then just add that include directory if MSVC is detected, and it would provide compatibility shims similar to MinGW. Avoiding modifying any of the normal files.

Would that be feasible?

jose-pr:
Yes, and it works out better than I expected. I've built it rather than
guessing, so here are the numbers.

`contrib/msvc/include` holds five headers — `strings.h`, `unistd.h`,
`libgen.h`, `sys/time.h`, and one `msvc_compat.h` that is force-included
(`/FI`) because the GCC attributes can't be reached any other way. It goes
on the include path only inside the existing `if(MSVC)` block.

**The whole tree now builds with MSVC — client libraries, the Windows
server, and the tests — with nothing under `common/`, `vncviewer/`,
`tests/` or `unix/` touched at all.**

| | before | after |
|---|---|---|
| files changed | 51 | 4 |
| lines changed | +477 −80 | +75 −9 |
| files touched in `common/` + `vncviewer/` + `tests/` | 46 | **0** |

The four that remain, and why a header can't absorb them:

1. **`CMakeLists.txt`** — where the include directory gets added, plus
   guarding the GCC warning flags (`-Wextra` and friends) behind
   `if(NOT MSVC)` with MSVC spellings on the other side. Unavoidable by
   construction.
2. **`win/wm_hooks/wm_hooks.cxx`** (+18) — the one attribute in the tree
   that isn't a diagnostic. `__attribute__((section("shared"), shared))`
   puts the hook state in a shared section; compiling it away gives you a
   DLL that builds and silently doesn't work, so it's bracketed by
   `data_seg` pragmas with the `/SECTION:shared,RWS` linker directive.
   I checked the built DLL rather than trusting the build: the section
   carries `IMAGE_SCN_MEM_SHARED` under both toolchains.
3. **`win/winvnc/VNCServerService.cxx`** (1 line) —
   `typedef void WINAPI (*SendSAS_proto)(BOOL)` has the calling convention
   outside the parentheses. gcc accepts it because `WINAPI` is an
   attribute there; it's arguably malformed either way, so this is a
   portability fix worth having regardless of MSVC.
4. **`win/rfb_win32/CMakeLists.txt`** (1 line) — it includes
   `core/i18n.h` but didn't link `core`, so it only found `libintl` when
   that happened to be on the default include path.

Two things I found on the way that are yours whichever way this goes:

- **The `CLSID_VideoProcessorMFT` probe asks the wrong question.**
  `check_variable_exists` is a link test, but the fallback definition it
  feeds is guarded on whether a *header declares* the GUID. The symbol is
  in `mfuuid.lib` either way; the declaration depends on the toolchain
  and on `_WIN32_WINNT`, and `add_definitions()` doesn't reach
  `try_compile()`. So on MSVC it said "found" and the compile then failed.
  It's now a compile test over the same headers, in the same order, with
  the same target version — which correctly says no on MSVC at the Win7
  target and yes on MinGW.
- **`ENABLE_NLS=OFF` doesn't build**, and not only with MSVC. `i18n.cxx`
  calls `dgettext()` and friends, which `i18n.h` has redefined by then,
  and with NLS off there's nothing left to call. I hit this with
  clang/MinGW too, so it's independent of this PR. Happy to send it
  separately.

I verified MinGW still builds throughout — that was the failure mode of my
last push here and I didn't want to repeat it. Tested with MSVC 14.44
(ARM64, VS 2022 Build Tools) and MSYS2 CLANGARM64.

If this shape looks right I'll rewrite the PR onto it.

The branch is at https://github.com/jose-pr/tigervnc/tree/msvc-wrappers if you want to look before deciding — one commit on current master. I have deliberately not touched this PR's branch yet.

CendioOssman:
That looks very promising. Please update the PR and we can sort out the details here.

jose-pr:
Updated — the branch is now the wrapper approach, rebased onto current master (so it has the audio merge under it), as one commit.

Final shape is 9 files: the five headers under `contrib/msvc/include`, and four normal files — `CMakeLists.txt`, `win/rfb_win32/CMakeLists.txt`, `win/winvnc/VNCServerService.cxx` (1 line) and `win/wm_hooks/wm_hooks.cxx`. Still nothing under `common/`, `tests/` or `unix/`. The normal-file total is +80 −10 rather than the +75 −9 I quoted, because rebasing onto master with the viewer enabled turned up two more GCC-isms that belong in the same bucket:

- The FLTK version check sets `CMAKE_REQUIRED_FLAGS "-Wno-error"`, which `cl` rejects outright, so the check failed and the build stopped at "Incompatible version of FLTK" with a perfectly good FLTK 1.3.11 installed. It now gets `/WX-` on MSVC.
- A few places use the alternative spellings (`and`, `not`). MSVC only recognises those via `<iso646.h>`, which the compatibility header now includes.

I should be precise about scope, since I was loose about it last time. What I have verified is **the client libraries, the Windows server and the tests** — `BUILD_VIEWER=OFF`. The viewer does not build with MSVC yet, and the two things stopping it are both pre-existing and independent of MSVC:

1. `OptionsDialog::handleX509()` is defined unguarded, but the members it touches (`encX509Checkbox`, `caInput`, `crlInput`) are declared under `#ifdef HAVE_GNUTLS`. Any build without GnuTLS fails, on any compiler.
2. `vncviewer.cxx:393` puts `#ifndef WIN32` inside the argument list of the `_()` macro. That is undefined behaviour rather than something MSVC is being awkward about; gcc happens to accept it.

Both are one-liners and I am happy to send them, but they touch `vncviewer/` so I have deliberately left them out of this PR rather than quietly widen it. Say which you would prefer.

Verified on MSVC 14.44 (ARM64, VS 2022 Build Tools) and MSYS2 CLANGARM64, with the viewer and audio enabled on the MinGW side so the merge is covered there too.

jose-pr:
Correction to my last comment, and a better result than I reported.

I listed two things as stopping the viewer under MSVC. Only one of them actually is.

The `OptionsDialog::handleX509()` problem is real, but it is a *no-GnuTLS* bug rather than an MSVC one — any build without GnuTLS fails on any compiler, and it simply happened to be how my first MSVC attempt was configured. With GnuTLS present it does not arise. I had wrongly presented it as an MSVC blocker.

That leaves exactly one: `vncviewer.cxx:393`, where `#ifndef WIN32` sits inside the argument list of the `_()` macro. With that one construct restructured, **the viewer builds with MSVC and runs** — I checked it rather than stopping at the link:

- `vncviewer.exe` starts and prints its usage.
- Against a small RFB server streaming the audio extension, it negotiates S16/2ch/48 kHz and consumes a full stream (768000 bytes for four seconds) with no protocol error. So the audio code that just landed compiles and works under MSVC too.

I have **not** put that fix in this PR. Splitting the format string into a WIN32 and a non-WIN32 copy is the obvious way to do it, but it changes a translated string and so has a `.po` cost, and that seems like your call rather than something I should decide inside a portability PR. Happy to add it here, send it separately, or leave it alone — whichever you prefer.

Also worth flagging while I am in here, and independent of this PR: `wm_hooks`'s shared section is not actually shared when built with clang. `__attribute__((section("shared"), shared))` — clang honours the `section` half and ignores the `shared` half, with a `-Wignored-attributes` warning for each variable. I checked the section characteristics in the built DLLs:

```
gcc/MinGW    (your CI)      MEM_SHARED = true
clang/MinGW                 MEM_SHARED = false
MSVC (this PR's data_seg)   MEM_SHARED = true
```

I confirmed this on unmodified master, so it is not something this PR introduces, and your CI uses gcc so released builds are fine. But anyone building with clang gets a `wm_hooks.dll` where each injected process has its own copy of the hook state, which fails quietly rather than loudly.

jose-pr:
Added a `contrib/msvc/README` alongside the headers, since the compatibility layer is only half of what someone repeating this needs.

The other half is the dependencies, and that part is genuinely obscure: GnuTLS's vcpkg port is marked `"supports": "!windows | mingw"`, so a Windows triplet correctly contains no `gnutls.lib` and it reads as a dead end rather than an expected gap. It has to be built separately. The ShiftMediaProject MSVC solutions cover zlib/gmp/nettle/gnutls but ship Win32 and x64 only, so ARM64 needs adding to each — worth recording that this needs no library source changes at all, only msbuild project files, plus an ARM64 case in ShiftMediaProject/VSYASM's `yasm.props` and installer.

The note also records the clang/`wm_hooks` behaviour from my last comment, so it is findable by whoever hits it rather than only living in this thread.

Happy to drop the file if you would rather not carry build notes for a toolchain that is not supported — it is one commit on its own and trivial to remove.

jose-pr:
A correction, and then the thing I should have provided in the first place.

I said the ARM64 work on the GnuTLS stack needed no library source changes, only msbuild project files. That was wrong, and wrong because I described it from a saved set of patches rather than from what actually built. Regenerating from the real trees turned up several files those patches never captured. The msbuild half is as I described — ARM64 configurations, and excluding x86 assembly — but there is a second half:

- `gmp/longlong.h`: `<immintrin.h>` is a hard `#error` on ARM64, and `_umul128` is x64-only, so `umul_ppmm` has to be built from `__umulh` plus an ordinary 64-bit multiply.
- `gmp/SMP/config.h`, `nettle/SMP/config.h`: the fork ships no aarch64 assembly, so the `HAVE_NATIVE_mpn_*` declarations have to go on ARM64 — left in place they make the library skip its own portable C and then fail to link against routines that do not exist.
- `gnutls/SMP/config.h`: PadLock is a VIA x86 instruction-set extension; disabled, and its sources excluded.
- `gmp/SMP/gmp.h`: `__aarch64__` / `_M_ARM64` added to the 64-bit limb detection.

The note in `contrib/msvc/README` is corrected accordingly — it now points at the patches rather than paraphrasing them, which is what I should have done to begin with.

The patches themselves are on a branch rather than in this PR, since nothing in them is part of TigerVNC's build:

https://github.com/jose-pr/tigervnc/tree/msvc-deps — `contrib/msvc/deps`

Five patches (zlib, gmp, nettle, gnutls, VSYASM) against pinned ShiftMediaProject commits, each verified to apply cleanly to a fresh checkout at the commit named in the README, with notes on what the non-msbuild parts do and why. It is a record of one working configuration, not something maintained — but it should save the next person the day it cost.

win/rfb_win32/CMakeLists.txt
```diff
@@ -38,4 +38,7 @@ if(BUILD_WINVNC)
   target_sources(rfb_win32 PRIVATE ${RFB_WIN32_SOURCES} CleanDesktop.cxx)
 endif()
 
+# core for its headers as much as its code: rfb_win32 includes
+# core/i18n.h, and libintl is only on the include path via core
+target_link_libraries(rfb_win32 core)
```
CendioOssman: This is a separate fix. Fortunately, it is fixed on `master` for unrelated reasons.

win/winvnc/VNCServerService.cxx
```diff
@@ -43,7 +43,7 @@ const char* winvnc::VNCServerService::Name = "TigerVNC";
 
 // SendSAS is not available until Windows 7, and missing from MinGW
 static HMODULE sasLibrary = nullptr;
-typedef void WINAPI (*SendSAS_proto)(BOOL AsUser);
+typedef void (WINAPI *SendSAS_proto)(BOOL AsUser);
```
CendioOssman, line 46: This seems to be a general fix. Please put it as a separate commit.

win/wm_hooks/wm_hooks.cxx
```diff
@@ -26,7 +26,21 @@
...
+#ifdef _MSC_VER
+#pragma comment(linker, "/SECTION:shared,RWS")
+#define SHARED
+#define SHARED_BEGIN __pragma(data_seg("shared"))
+#define SHARED_END __pragma(data_seg())
+#else
 #define SHARED __attribute__((section ("shared"), shared))
+#define SHARED_BEGIN
+#define SHARED_END
+#endif
```
CendioOssman: For clarity, it's probably best to do this as a separate commit from all the include file magic.

CMakeLists.txt
```diff
@@ -54,6 +59,11 @@ message(STATUS "CMAKE_BUILD_TYPE = ${CMAKE_BUILD_TYPE}")
 message(STATUS "VERSION = ${VERSION}")
 message(STATUS "BUILD_TIMESTAMP = ${BUILD_TIMESTAMP}")
 
+# The flags below are GCC's spelling of things MSVC either spells
+# differently or does by default, so it is given its own equivalents
+# rather than these.
+if(NOT MSVC)
+
```
CendioOssman: For clarity, I think it is best if we do this section by section. It's annoying to match things with this large block.

And this is probably also best in a separate commit.

CMakeLists.txt
```diff
@@ -102,11 +132,6 @@ if(ENABLE_TSAN AND NOT WIN32 AND NOT APPLE AND CMAKE_SIZEOF_VOID_P MATCHES 8)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=thread")
 endif()
 
-if(MSVC)
-  # undef min and max macro
-  target_compile_definitions(rfb PRIVATE NOMINMAX)
-endif()
-
```
CendioOssman, line 109: This should have been removed back in e0e4f4ca67dbb2db2cf547431fc1c60192aad21a. Please do a separate cleanup commit to not confuse it with the rest.

CMakeLists.txt
```diff
@@ -186,8 +212,27 @@ if(ENABLE_H264)
...
+      #include <mfapi.h>
+      #include <mferror.h>
+      #include <wmcodecdsp.h>
+      #include <windows.h>
+      #include <mfidl.h>
+      #include <mftransform.h>
+      int main() { return (int)CLSID_VideoProcessorMFT.Data1; }"
+      HAVE_VIDEO_PROCESSOR_MFT)
+    set(CMAKE_REQUIRED_DEFINITIONS)
     set(CMAKE_REQUIRED_LIBRARIES)
```
CendioOssman: This is a general fix, so should also be separate.

contrib/msvc/README
```diff
@@ -0,0 +1,99 @@
...
+
+win/wm_hooks/wm_hooks.cxx keeps its state in a shared section so that
+every process the DLL is injected into sees one copy. That attribute is
+not a diagnostic, so compiling it away would leave a DLL that builds and
+silently does not work. It is bracketed by data_seg pragmas instead.
+
+Worth knowing on the MinGW side as well: clang honours the section() half
+of that attribute and ignores the shared half, so a clang-built DLL has
+the section without IMAGE_SCN_MEM_SHARED (0x10000000) set. Only gcc
+honours both. Check the section characteristics if it matters to you.
```
CendioOssman: These are fixed issues, so we don't need to document that.

CendioOssman (changes requested):

jose-pr:
Thanks for the review. I've split it up as suggested and rebased on current master:

1. Remove obsolete NOMINMAX definition
2. Put WINAPI inside the SendSAS function pointer type
3. Check that CLSID_VideoProcessorMFT is declared
4. Share hook state with MSVC as well
5. Select compiler flags separately for MSVC
6. Keep preprocessor directives out of the usage text
7. Build with MSVC using compatibility headers

- The `rfb_win32` change is dropped, as master now has it.
- The compiler flags are handled section by section, next to the existing comments, rather than as one large block.
- The README no longer documents issues that are fixed.

Commit 6 is new. The viewer didn't build with MSVC because `usage()` has `#ifndef WIN32` inside the `_()` argument list, which MSVC rejects. It now selects the whole format string per platform instead. I checked that the output is byte for byte identical to before on both Windows and Unix.

I've tested locally on Windows ARM64 with MSVC and MinGW (clang). Each commit builds with MinGW, and the whole tree, including the viewer, builds in Debug and Release with both compilers. All unit tests pass in each configuration.

CMakeLists.txt
```diff
@@ -174,8 +175,19 @@ if(ENABLE_H264)
   if(WIN32)
     set(HAVE_H264 1)
 
+    # Same headers and target version as H264WinDecoderContext.cxx
```
CendioOssman, line 178: We don't have to keep them in sync. We can keep this as minimal as possible. Any false positives will result in compilation errors, which are then easily caught.

This also means we don't have to add any odd extra variables.

CMakeLists.txt
```diff
@@ -174,8 +175,19 @@ if(ENABLE_H264)
...
+    set(CMAKE_REQUIRED_DEFINITIONS -D_WIN32_WINNT=${WIN32_WINNT})
     set(CMAKE_REQUIRED_LIBRARIES ole32 mfplat mfuuid wmcodecdspuuid)
-    check_variable_exists(CLSID_VideoProcessorMFT HAVE_VIDEO_PROCESSOR_MFT)
```
CendioOssman, line 178: Do we even need this test? This might be a workaround for something that has been fixed since ages.

CMakeLists.txt
```diff
@@ -55,37 +55,65 @@ message(STATUS "VERSION = ${VERSION}")
...
+  set(CMAKE_C_FLAGS_MINSIZEREL "${CMAKE_C_FLAGS_MINSIZEREL} /UNDEBUG")
+  set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} /UNDEBUG")
+else()
+  set(CMAKE_C_FLAGS_RELEASE "${CMAKE_C_FLAGS_RELEASE} -UNDEBUG")
+  set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -UNDEBUG")
+  set(CMAKE_C_FLAGS_RELWITHDEBINFO "${CMAKE_C_FLAGS_RELWITHDEBINFO} -UNDEBUG")
+  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -UNDEBUG")
+  set(CMAKE_C_FLAGS_MINSIZEREL "${CMAKE_C_FLAGS_MINSIZEREL} -UNDEBUG")
+  set(CMAKE_CXX_FLAGS_MINSIZEREL "${CMAKE_CXX_FLAGS_MINSIZEREL} -UNDEBUG")
+endif()
```
CendioOssman, line 72: There's a lot of duplication here now, unfortunately. Another approach is probably warranted.

Can something like `add_definitions(-UNDEBUG)` work instead?

CMakeLists.txt
```diff
@@ -55,37 +55,65 @@ message(STATUS "VERSION = ${VERSION}")
...
+  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Og")
+endif()
 
 # Make sure we get a sane C and C++ version
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
+if(NOT MSVC)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
+endif()
```
CendioOssman, line 93: We often rely on this, so it would be good if MSVC alternatives are added.

CMakeLists.txt
```diff
@@ -55,37 +55,65 @@ message(STATUS "VERSION = ${VERSION}")
...
+if(MSVC)
+  set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} /WX")
+  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /WX")
+else()
+  set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Werror")
+  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Werror")
+endif()
 # clang doesn't support format_arg, which breaks this warning
-if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+if(NOT MSVC AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
CendioOssman, line 116: This seems unnecessary as the id won't match "Clang" anyway?

CMakeLists.txt
```diff
@@ -55,37 +55,65 @@ message(STATUS "VERSION = ${VERSION}")
...
 # But extra debug checks are still gated by this custom define
-set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -D_DEBUG")
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG")
+if(MSVC)
+  set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} /D_DEBUG")
+  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} /D_DEBUG")
+else()
+  set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -D_DEBUG")
+  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG")
+endif()
```
CendioOssman, line 81: This could also be converted to use `add_definitions()`. We'd need to check `CMAKE_BUILD_TYPE`, though.

CMakeLists.txt
```diff
@@ -314,7 +342,11 @@ if(BUILD_VIEWER)
   endif()
 
   if(FLTK_FOUND)
-    set(CMAKE_REQUIRED_FLAGS "-Wno-error")
+    if(MSVC)
+      set(CMAKE_REQUIRED_FLAGS "/WX-")
```
CendioOssman, line 346: I'm uncertain what bug this resolved. Have you checked if it is needed for MSVC?

vncviewer/vncviewer.cxx
```diff
@@ -390,20 +390,24 @@ static void usage(const char *programName)
...
+          "       %s [parameters] [host][::port]\n"
+          "       %s [parameters] -listen [port]\n"
+          "       %s [parameters] [.tigervnc file]\n"),
+          programName, programName, programName, programName);
+#else
   fprintf(stderr, _(
           "\n"
           "Usage: %s [parameters] [host][:displayNum]\n"
           "       %s [parameters] [host][::port]\n"
-#ifndef WIN32
```
CendioOssman, line 397: This is actually a bug you found, not just an incompatibility with MSVC. The translations won't understand this magic either, and won't generate the two different texts we actually need.

CendioOssman (changes requested):
Looking real good. There are some last tweaks to consider to make this more maintainable. But otherwise I'd say it's ready.

Did you consider adding a GitHub action to build this as well? It will help us avoid breaking things.

CMakeLists.txt
jose-pr: Agreed. I've ended up removing the test entirely, see the other comment.

CMakeLists.txt
jose-pr: It's still needed for MSVC, unfortunately. The Windows SDK only declares CLSID_VideoProcessorMFT when targeting Windows 8.1 or newer, and we target Windows 7. Current MinGW always declares it, though.

So instead of testing for it, the decoder now has its own copy of the GUID under a different name. That works everywhere, and there's nothing left to detect.

CMakeLists.txt
jose-pr: Yes, that works. I checked that NDEBUG ends up undefined in release builds with both MSVC and MinGW. It's a separate commit now, as it isn't really MSVC specific.

CMakeLists.txt
jose-pr: Done, in the same commit. With Visual Studio's generator CMAKE_BUILD_TYPE is empty, but the debug runtime defines _DEBUG there anyway, so it still ends up set.

CMakeLists.txt
jose-pr: Added /std:c11 and /std:c++14, and /std:c++17 for the unit tests. MSVC has no C99 or C++11 modes, so these are the closest it has.

CMakeLists.txt
jose-pr: Right, reverted.

CMakeLists.txt
jose-pr: The -Wno-error there came with the FLTK version check. MSVC rejects that option outright, so the check always failed and claimed the wrong FLTK version. MSVC doesn't need an equivalent, so it's now just skipped there.

vncviewer/vncviewer.cxx
jose-pr: Good point. I checked, and the catalog only has the Unix version of the text today. With this change xgettext picks up both. I've updated the commit message to describe it as a bug.

jose-pr:
Thanks! Yes, I've added one in the last commit. It's a separate workflow that only runs when started from the Actions tab, as MSVC isn't the main compiler for TigerVNC and I didn't want it slowing down or adding to the regular checks on every push and pull request. If you'd rather have it run automatically, it's a one line change to the trigger.

It's not quick without a cache. On my fork, building the dependencies with vcpkg took about 49 minutes (gettext and gmp are most of that), and the whole job about 52 minutes. With the vcpkg cache in place it was down to about 4 minutes. Here's an example run of the same job: https://github.com/jose-pr/tigervnc/actions/runs/34989583759 (the first attempt is without the cache, the second with it).

It builds for x64 with translations, but without GnuTLS, as vcpkg only has that for MinGW on Windows. It's a release build for now, as there are still a number of warnings with MSVC.

jose-pr:
The rocky9 package build failure looks unrelated to this change. It fails in `dnf builddep`, before anything is compiled:

```
nothing provides openssl-libs(x86-64) = 1:3.5.8-1.el9_8 needed by openssl-devel-1:3.5.8-1.el9_8.x86_64 from appstream
```

So the mirror has the new `openssl-devel` but not the matching `openssl-libs` yet. The same job passed for the previous version of this branch, and it should pass again once the mirror catches up.

Issue #369, opened 2016-10-16
support for HiDPI


alekna:
On a high DPI (HiDPI) screen TigerVNC is unusable. Would be it possible to implement scaling factor option in TigerVNC client? Remmina seems to support HiDPI out of the box.

Please find a screenshot below. TigerVNC is on the left, Remmina - on the right. Both are connected to the same VNC server.
![screenshot from 2016-10-17 00-07-49](https://cloud.githubusercontent.com/assets/10272121/19420918/5855a816-93fe-11e6-9e02-0e4d7e1b17f8.png)

hildred:
If you are using Xvnc (or vnc server) and you only connect from high resolution displays, the -dpi option to Xvnc may be useful.

alekna:
I tried running it with `-dpi 192`. not much of an improvement.
![screenshot from 2016-10-19 02-10-22](https://cloud.githubusercontent.com/assets/10272121/19499856/67423bba-95a1-11e6-97af-2f08ae2ef5ce.png)

hildred:
try larger numbers like 500 or 1000, also make sure you have a scalable fixed width font.

alekna:
It looks really bad with higher `-dpi`. In my opinion scaling should be done on the client side. It should be font and application agnostic. Just like Reminna does that.
![screenshot from 2016-10-24 22-28-42](https://cloud.githubusercontent.com/assets/10272121/19660708/8d4bd3e2-9a39-11e6-9b13-3a7cec6f23c2.png)

zeroimpl:
FYI I got good results by setting the Linux desktop to a high resolution (3840x1960) and configured for DPI of 192, then just using the **TigerVNC Java client** to scale locally by 50%.

imperative:
In addition to the scaling of the remote view, there seem to be issues with the interface of the C++ client. 

Does it have any interface scaling at all? It seems to use the library FLTK, which does support scaling, and should even read an environment variable FLTK_SCALING_FACTOR. However in practice, changing it does not seem to have any effect. Why is that? Doesn't the client use the scaling features provided by FLTK? Does it override the environment variable setting with some hard-coded value?

CendioOssman:
No, there is no interface scaling at all. That FLTK setting is unfortunately only for the next version of FLTK. We are still targeting the stable 1.3.x branch of FLTK.

johann-petrak:
[...]

PR #2137, opened 2026-08-02
Declare the viewer DPI-aware on Windows

jose-pr:
Windows renders the viewer to a virtualised logical desktop and
bitmap-stretches the result unless the application declares itself
DPI-aware. On a 200% display a 3270x2180 panel is reported as 1635x1090, so
a 1920x1200 remote framebuffer no longer fits — `DesktopWindow` logs
"Reducing window size to fit on current monitor" — and every pixel that is
drawn is then upscaled by the compositor. Blur on a remote desktop is the
one thing worth avoiding.

### Why "true" and not PerMonitorV2

Per-monitor awareness obliges the application to rescale itself on
`WM_DPICHANGED` when it is dragged between differently scaled monitors, and
FLTK 1.3 has no HiDPI support to build that on. Claiming it would replace
stretching with a viewer that draws at the wrong size, which is worse.
System DPI awareness is what the toolkit can actually honour today.

If FLTK ever grows per-monitor support, the manifest is the only thing that
needs revisiting.

### Why the manifest is delivered two ways

The linkers differ. MSVC synthesises its own application manifest, so a
second `RT_MANIFEST` compiled in through the `.rc` collides with it; CMake
merges a `.manifest` listed as a source instead. MinGW's linker synthesises
nothing, so there it has to go through the `.rc`. Both paths are in the
`CMakeLists.txt` change and the comment says which is which.

### Scope

Behind `ENABLE_DPI_AWARE`, default ON, Windows only. Nothing changes on any
other platform, and turning it off restores exactly the current behaviour.

### Testing

Built and run on Windows. I have not exercised the MinGW path your CI uses —
I build this tree with MSVC through a local change that is not part of this
pull request — so the `.rc` branch is the one that would benefit from a
second pair of eyes. It is the path upstream builds today, and the manifest
resource id it uses (`1 24`) is the standard `CREATEPROCESS_MANIFEST_RESOURCE_ID`
/ `RT_MANIFEST` pair.

### DIFF
--- a/vncviewer/CMakeLists.txt
+++ b/vncviewer/CMakeLists.txt
@@ -30,10 +30,28 @@ if(WIN32 AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
   set_target_properties(vncviewer PROPERTIES WIN32_EXECUTABLE TRUE)
 endif()
 
+option(ENABLE_DPI_AWARE "Declare the viewer DPI-aware on Windows (sharp 1:1 rendering; see vncviewer.exe.manifest)" ON)
+
 if(WIN32)
   # Since vncviewer.rc is generated, local includes will be looking
   # in the wrong directory. We need to help it out.
   target_include_directories(vncviewer PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+
+  # The application manifest carries the DPI-awareness declaration. MSVC's
+  # linker synthesises its own manifest, so a second RT_MANIFEST resource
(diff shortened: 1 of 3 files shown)


CendioOssman:
I'm afraid this is not a fix. It simply declares we can handle high DPI when that isn't the case. The GUI needs to be able to scale. And session handling is even more complex, as we don't know what DPI the remote end is using.

Please see #369 for the main issue tracking this feature.

PR #2138, opened 2026-08-02
Add support for QEMU's audio extension

jose-pr:
QEMU can send a guest's audio to a VNC client over a vendor extension of its
own. This teaches the viewer to ask for it and to play it, so that connecting
to a QEMU guest gives you sound as well as a picture.

**This is @mkupchik's work.** He opened #1478 in June 2022 with a complete
implementation of this feature, got a review saying "This is definitely
something we are interested in" and a list of changes, and then it stalled.
It is still open and now conflicts with master. What is here is his design,
rebased onto current master, with the outstanding review comments addressed.
The playback engine in particular is his — the circular buffer, the silence
played ahead of a new stream, the lock-free handoff out of the device
callback — and his copyright is on the files that carry it.

I have opened this separately only because #1478 has had no commits from its
author since July 2022. **If he is still around, or if you would rather this
went in as his, I would much prefer to send these as commits on top of #1478
and close this.**

### The review comments on #1478

Five things were asked for. Two were done, one partly, and two were not. The
two that were not are done here.

- *"please rename this to something like `supportsAudio`"* — partly done
  there. Finished here as far as I think it should go: everything a subclass
  sees is generic (`CConnection::supportsAudio`, `getAudioFormat()`,
  `handleAudioBegin()`, `handleAudioEnd()`, `handleAudioData()`). I did
  **not** rename `supportsQEMUAudio()` on `CMsgHandler`, because it reports
  one specific QEMU pseudo encoding arriving and sits directly beside the
  existing `supportsQEMUKeyEvent()`. Say the word if you would rather it were
  generic there too.
- *"Please try to define constants for things such as this"* — done, in
  `qemuTypes.h`.
- *"Can this be pushed to CConnection instead? Starting the audio on first
  detection of QEMU audio support should be generic"* — done;
  `CConnection::framebufferUpdateEnd()` starts the handshake.
- *"This complexity is something that doesn't really fit well in CMsgReader
  ... Can't we let the input stream handle the buffering? ... We don't need
  to support the theoretical extremes"* — **done.** `readQEMUServerMessage()`
  is 41 lines and holds no state of its own: read the submessage and the
  operation, return early for anything that is not audio data, read the
  length, refuse anything over 1 MiB with a `protocol_error`,
  `hasDataOrRestore()`, hand `getptr()` straight to the handler, `skip()`.
  `MSGSTATE_AUDIO_DATA`, `nAudioBytesLeft` and `readAudioData()` are gone.
- *"Doesn't Windows resample for us? Do we really need to try all of these?
  ... Have you actually seen it failing to give you the requested format?"* —
  **done.** The search over 16 combinations of sample width, channel count
  and frequency is gone. One format is asked for, and if the device refuses
  it the viewer reports that it has no audio and never advertises the
  encoding.

  That one format is **48 kHz stereo S16**, which is what #1478 itself
  arrived at — its last commit was "Switched to 48 kHz output sample rate
  ... to avoid downsampling in QEMU for modern Windows guests". Narrowing the
  search had quietly undone that; it is restored.

### Shape

`CMsgHandler` gains two methods rather than five: one saying the server has
advertised audio, and one passing a QEMU submessage on undecoded. Working out
that submessage 1, operation 2 means audio data happens in `CConnection`,
which is the layer that already knows what the extension is. The do-nothing
defaults live on `CConnection` beside the clipboard ones, since `CMsgHandler`
is pure virtual throughout.

Playback sits behind an `AudioOutput` interface shaped like `Keyboard`.
There are two backends:

- **Windows**, on `waveOut` — @mkupchik's, essentially unchanged.
- **Linux and friends**, on **libpulse**, which covers both PulseAudio and
  PipeWire systems since PipeWire's pulse server is what answers there.

The Pulse backend is the asynchronous API driven by a `pa_threaded_mainloop`
rather than `pa_simple`, because `AudioOutput.h`'s own contract is that no
method may block and `pa_simple_write()` blocks until the server has taken
the data. Its buffering is deliberately the same as the waveOut backend's,
constant for constant — the same circular buffer, the same drop when it
fills, the same silence ahead of a stream. Only the handover differs, since
`pa_stream_write()` copies the samples out and PulseAudio reports starvation
directly rather than by inference.

libpulse is optional, via `trioption(ENABLE_AUDIO)` in the shape
`ENABLE_H264` already uses: a build without it compiles exactly as before and
`create()` keeps returning nullptr. On Windows it is unconditional, waveOut
being part of the OS.

There is an `Audio` parameter, on by default, with a checkbox and a man page
entry. #1478 enabled audio unconditionally, with no way to turn it off.

### What this does not do

- Playback only. The extension only goes one way.
- The setting takes effect at connect time, not mid-session.
- No macOS backend. `AudioOutput::create()` returns nullptr there and the
  viewer never advertises the encoding, exactly as on a Linux box built
  without libpulse.

### Reproducing it

Start QEMU with an audio device and tell the VNC server about it:

```
qemu-system-x86_64 ... \
  -audiodev none,id=aud0 \
  -device ich9-intel-hda -device hda-duplex,audiodev=aud0 \
  -vnc :0,audiodev=aud0
```

`-audiodev none` is the point: the host plays nothing, it only captures for
VNC. Then connect and make the guest play something.

Worth knowing if you go looking: QEMU only offers the extension when it was
started with an audio device, and it disconnects a client that sends an audio
message before the server has acknowledged the pseudo encoding. So the
acknowledgement rectangle has to gate the handshake.

### Testing

The handshake, on both platforms:

```
CConnection: Requesting audio (format 3, 2 channels, 48000 Hz)
```

**Windows — heard.** A 1 kHz sine tone, the ALSA spoken-channel samples, and
thirty seconds of continuous synthesised stereo music were played in a Linux
guest under QEMU and heard from the client's speakers. The music was panned
slowly between channels, so it also confirms the two channels carry different
content rather than one duplicated. No dropouts, and the viewer was still
connected at the end.

**Linux — measured, not heard.** The viewer ran on Rocky Linux 9 against the
same QEMU audio display, playing into a synthetic sink whose PCM was captured
and analysed. Against a server sending a sine it generated itself, the peak
was **0.610352 — 20000/32768 to the last digit**. Halving the source
amplitude halved it to 0.305176; moving the tone moved the dominant frequency
bin. Against the real QEMU display the captured tone read 439.96–439.98 Hz at
0.610260 and 0.305084 for source amplitudes of 20000 and 10000, while the
kernel's own socket byte counter read 194,708 B/s against 192,000 B/s of
audio. No dropouts in any run.

The control matters as much as the signal: with `-Audio=0` the capture is
exact zeros for its whole length, the `-259` pseudo encoding is absent from
`SetEncodings`, and the socket counter reads 0 B/s.

**Nobody listened on Linux**, and I am not claiming otherwise — that run
supports "the backend handed the sound server exactly the samples it was
given, at the right rate, with no gaps", which is a narrower claim than the
Windows one.

Two paths in the Pulse backend are written and reviewed but never fired,
because the local network never starved the stream: the adaptive widening of
the pre-roll silence after an underflow, and the drop when the buffer fills.
The bounded connect timeout was likewise never hit, since the sound server
always answered immediately.

One caveat on the Windows side of this branch: it is based on upstream, which
declines to configure under MSVC, so the Windows build here was compiled
through a local opt-in that is not part of this pull request, and I have not
exercised the MinGW path your CI uses. Nothing in the audio code is compiler
specific — `<mmsystem.h>` and `waveOut*`, which mingw-w64 has — but I have
not proven that.


CendioOssman:
Thanks for having another look at this. I don't normally use QEMU. Do you have a quick guide to point to to quickly get a test machine running?

jose-pr:
Of course — and thanks for taking a look.

The one thing worth saying up front, because it looks exactly like the feature being broken: the audio comes from **QEMU's own `-vnc` display**, not from a VNC server running inside the guest. If you boot a desktop guest and connect to `vncserver`/`Xvnc` inside it, you'll get a picture and no sound, and nothing is wrong. You need an `audiodev` attached to a sound card on QEMU's `-vnc` display.

A minimal guest that is enough to test with — no OS install needed, any bootable ISO with a startup sound, or just run `speaker-test` once it's up:

```
qemu-system-x86_64 \
  -m 2048 \
  -audiodev alsa,id=snd0 \
  -device intel-hda -device hda-duplex,audiodev=snd0 \
  -vnc :1,audiodev=snd0 \
  -cdrom <any-live.iso>
```

The load-bearing parts are `-audiodev …,id=snd0`, a sound card bound to it (`hda-duplex,audiodev=snd0`), and **`audiodev=snd0` on the `-vnc` option itself** — that last one is what makes QEMU offer the audio pseudo-encoding to the client. Without it everything else still works and there is simply no audio extension advertised.

Two things that cost me a few minutes, in case they save you the same: `-audiodev help` lists what your build actually has (mine has no `pa`, so `alsa` it is), and **`hda-duplex` needs a backend with a capture side** — pointing it at `wav` fails with ``Could not create a backend for voice `adc'``. If you only care about playback into the viewer, `-device hda-output` avoids that entirely.

Then `vncviewer :1` with this branch built with `-DENABLE_AUDIO=ON`. Inside the guest, `speaker-test -c2 -t sine` is the quickest way to get a continuous tone.

If you'd rather not run QEMU at all, I've put a small RFB server on a branch off this one that does the same job deterministically:

https://github.com/jose-pr/tigervnc/tree/qemu-audio-bench/tests/audio

```
python tests/audio/qemu_audio_server.py --port 5900   # then: vncviewer :0
python tests/audio/qemu_audio_server.py --port 5900 --no-audio
```

It speaks 3.8/None, answers one FramebufferUpdateRequest with a rectangle carrying pseudo-encoding -259, reads `SetAudioFormat`/`EnableAudio` and then streams a sine it generates itself. Because nothing in that path resamples or mixes, the amplitude a correct client emits is an exact number rather than just "nonzero" — at `--amplitude 20000` it's 20000/32768 = 0.610352, and that's what I measure with this branch.

The second line is the control arm: the identical session with the pseudo-encoding withheld, where the viewer must play nothing at all. Worth running, because without it a bench can't distinguish "audio works" from "the capture rig reports sound whatever happens". It also exits non-zero if audio was offered and never enabled, so a run that streamed nothing can't be mistaken for a green one.

Standard library only, no dependencies, and written from the wire format — it shares no code with TigerVNC or any other VNC implementation, which is what makes agreement between it and the viewer worth something. Happy to fold it into this PR, send it as its own, or leave it on the branch as a testing aid — whichever you prefer.

CendioOssman:
A quick test here just gives me a crash with `CConn:       Input stream used without underrun check`.

Have you tried this with a debug build? The above guard rail isn't present on release builds.

jose-pr:
Thanks — that's a real bug, and a good catch.

You're right that I'd been testing a release build, which is exactly why I didn't see it: `RFB_INSTREAM_CHECK` is only defined under `_DEBUG`, so the guard rail that caught this was compiled out on my side. My fault for not building both.

The cause is in `CMsgReader::readQEMUServerMessage()` — I spend the underrun budget twice:

```cpp
if (!is->hasDataOrRestore(length))   // checkedBytes = length
  return false;
is->clearRestorePoint();

handler->handleQEMUServerMessage(submessage, operation,
                                 is->getptr(length), length);  // check(length) -> checkedBytes = 0
is->skip(length);                                              // check(length) -> throws
```

`getptr()` and `skip()` each call `check(length)`, but only one `hasData()` paid for it. `JPEGDecoder.cxx` already documents precisely this — "getptr() resets the amount of assured data", followed by a re-assert — I just didn't follow it here. The fix is to re-assert between the two:

```cpp
  handler->handleQEMUServerMessage(submessage, operation,
                                   is->getptr(length), length);

  // getptr() resets the amount of assured data
  is->hasData(length);
  is->skip(length);
```

For what it's worth on severity: the bytes really were assured by the preceding `hasData()`, so this was a spurious assertion rather than an actual overrun — the release build was reading valid data. Not an argument for leaving it, though; a handler that can't survive a debug build isn't finished.

I've verified it against a debug build this time, driving the reader with synthesised AudioData messages: 8 of 9 cases fail before the change and all 9 pass after. That includes payloads dribbled in 1/3/7/64/1000-byte chunks, to force the re-entrant path where a message spans several reads — which I'd previously not been exercising at all, and which is the part I'd consider most likely to still hold a genuine bug. If it would help I can fold those cases in as a proper test rather than a throwaway harness.

Pushed as 3fc7648.

common/rfb/CConnection.cxx
```diff
@@ -589,6 +597,63 @@ void CConnection::framebufferUpdateEnd()
 
     firstUpdate = false;
   }
+
+  // The server tells us it can send audio by sending us a rectangle
+  // with that pseudo encoding, so this is the earliest point at which
+  // we can safely ask for it
+  if (server.supportsQEMUAudio && !audioRequested)
+    requestAudio();
```
CendioOssman, line 612: It's good that we don't have to explicitly do protocol stuff in the frontend. But audio requires cooperation from the frontend, so we might still need some signalling there.

Perhaps a `supportsAudio`, like we have `supportsLocalCursor`?

common/rfb/CConnection.cxx
```diff
@@ -589,6 +597,63 @@ void CConnection::framebufferUpdateEnd()
...
+  uint32_t frequency;
+
+  assert(!audioRequested);
+
+  // Only ask once, whatever the answer. A server that offered audio
+  // and then had it declined will not offer it again.
+  audioRequested = true;
+
+  if (!getAudioFormat(&sampleFormat, &channels, &frequency))
+    return;
```
CendioOssman: These settings seem fixed, so can't we just hard code everything here?

common/rfb/CConnection.cxx
```diff
@@ -1021,6 +1106,9 @@ void CConnection::updateEncodings()
     encodings.push_back(pseudoEncodingLEDState);
     encodings.push_back(pseudoEncodingVMwareLEDState);
   }
+  if (supportsAudio) {
```
CendioOssman: We avoid `{}` on single statements.

common/rfb/CConnection.cxx
CendioOssman: I see that we have that variable. We might want to check that here in case the server does silly things?

common/rfb/CConnection.h
```diff
@@ -297,8 +323,14 @@ namespace rfb {
     bool supportsCursorPosition;
     bool supportsDesktopResize;
     bool supportsLEDState;
+    // Unlike the others, leaving this off is not just a matter of
+    // ignoring something the server sends anyway. The server will
+    // encode and send audio purely because we asked for it, so do not
+    // ask unless there is somewhere for it to go.
```
CendioOssman: This is an implementation detail that we don't need to explain in the interface.

common/rfb/CConnection.h
```diff
@@ -297,8 +323,14 @@ namespace rfb {
...
+    bool supportsAudio;
 
   private:
+    void requestAudio();
```
CendioOssman: We try to match the order in the header with the order in the .cxx file. So this should probably be further down and not with the initialization helpers.

common/rfb/CMsgHandler.h
```diff
@@ -84,6 +85,16 @@ namespace rfb {
                                         const size_t* lengths,
                                         const uint8_t* const* data) = 0;
 
+    // A submessage of QEMU's vendor extension, with the payload that
+    // follows the operation code, if any. The submessage is left
+    // undecoded here as QEMU multiplexes several unrelated things on
+    // this message type, and only some of them concern any given
+    // handler.
```
CendioOssman: We don't comment the other ones, so we don't need to comment this one either.

common/rfb/CMsgReader.cxx
```diff
@@ -94,6 +100,48 @@ bool CMsgReader::readServerInit()
...
+  operation = is->readU16();
+
+  // Only audio data has a payload. Everything else is fixed size, and
+  // there is no generic length field we could use to skip an operation
+  // we do not know about.
+  if ((submessage != qemuAudio) || (operation != msgFromQemuAudioData)) {
+    is->clearRestorePoint();
+    handler->handleQEMUServerMessage(submessage, operation, nullptr, 0);
+    return true;
+  }
```
CendioOssman: We're getting a mix of abstractions here. If `CMsgReader` has to look at the types for some things, then it's probably best if it does all of that checking.

I.e. the `CMsgHandler` interface should have the subtypes as callbacks rather than the generic message.

*Or* it behaves like data rects and leaves it completely up to the callback to figure out how much data to read (if any).

CendioOssman: The latter requires more complex state handling, like data rects.

common/rfb/msgTypes.h
```diff
@@ -47,5 +47,12 @@ namespace rfb {
   const int msgTypeSetDesktopSize = 251;
 
   const int msgTypeQEMUClientMessage = 255;
+
+  // Same number, opposite direction. Unlike every other message type
+  // here, 255 is not direction specific: QEMU multiplexes both halves
+  // of its extension on it, and only the submessage id inside says
+  // which. Named separately so that a reader and a writer each state
+  // their intent.
```
CendioOssman: Again, no comment needed

common/rfb/qemuTypes.h
```diff
@@ -21,5 +21,26 @@
...
+  const int msgFromQemuAudioBegin = 1;
+  const int msgFromQemuAudioData = 2;
+
+  const int msgToQemuEnableAudio = 0;
+  const int msgToQemuDisableAudio = 1;
+  const int msgToQemuSetAudioFormat = 2;
+
+  // Sample formats for msgToQemuSetAudioFormat. The extension carries
+  // uncompressed PCM only, so these select width, signedness and (for
+  // the wider ones) host endianness, never a codec.
```
CendioOssman: Could consider dropping some of these comments as well. We rely on the rfbproto documentation, which means we don't have to duplicate protocol details in code comments.

vncviewer/AudioOutput.cxx
```diff
@@ -0,0 +1,50 @@
...
+
+AudioOutput* AudioOutput::create()
+{
+#ifdef WIN32
+  AudioOutputWin32* audio;
+
+  audio = new AudioOutputWin32();
+  if (!audio->isAvailable()) {
+    delete audio;
+    vlog.info("No audio playback device available");
```
CendioOssman: This should be translated.

vncviewer/AudioOutput.cxx
```diff
@@ -0,0 +1,50 @@
...
+    return nullptr;
+  }
+
+  return audio;
+#else
+  // No playback implementation for this platform yet
```
CendioOssman: Might want to have a log line here as well? At least on debug.

vncviewer/AudioOutput.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <stddef.h>
+#include <stdint.h>
+
+// An audio playback device, of which we only ever use one at a time.
+//
+// Everything here is called from the main thread, in the middle of
+// handling a protocol message, so no method may block. An
+// implementation is expected to hand the samples to the system and
+// return, not to wait for them to be played.
```
CendioOssman: We do (almost) everything single threaded. So this is not something we need to explicitly state.

vncviewer/AudioOutputWin32.cxx
```diff
@@ -0,0 +1,376 @@
...
+ * along with this software; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ * USA.
+ */
+
+// The waveOut playback design here, in particular the circular buffer,
+// the silence played ahead of a new stream, and the way finished
+// buffers are returned from the device callback, comes from Mikhail
+// Kupchik's implementation in the audio work that has been pending as
+// pull request #1478 since 2022.
```
CendioOssman: We don't need the development history permanently here. We have the PR for such details.

vncviewer/AudioOutputWin32.cxx
```diff
@@ -0,0 +1,376 @@
...
+
+#include "AudioOutputWin32.h"
+
+static core::LogWriter vlog("AudioOutputWin32");
+
+// How much audio to buffer ahead, and how much silence to play before a
+// new stream, to hide variations in when samples arrive over the
+// network.
+static const unsigned maxJitterMs = 1000;
+static const unsigned minStreamDelayMs = 20;
```
CendioOssman: Have you seen that this needs to be different per backend? Or should we coordinate this somewhere?

vncviewer/AudioOutputWin32.cxx
```diff
@@ -0,0 +1,376 @@
...
+    return true;
+  if (!available)
+    return false;
+
+  fillWaveFormat(&wfx, sampleFormat, channels, frequency);
+
+  if (waveOutOpen(&device, WAVE_MAPPER, &wfx,
+                  (DWORD_PTR)&AudioOutputWin32::waveOutCallback,
+                  (DWORD_PTR)this, CALLBACK_FUNCTION) != MMSYSERR_NOERROR) {
+    vlog.error("Could not open audio playback device");
```
CendioOssman: This should be translated

vncviewer/AudioOutputWin32.cxx
```diff
@@ -0,0 +1,376 @@
...
+    writtenHead(0), pendingHead(0),
+    streamId(0), extraDelayMs(0),
+    doneBuffers(nullptr), buffersInFlight(0)
+{
+  WAVEFORMATEX wfx;
+
+  fillWaveFormat(&wfx, sampleFormat, channels, frequency);
+
+  if (waveOutOpen(nullptr, WAVE_MAPPER, &wfx, 0, 0,
+                  CALLBACK_NULL | WAVE_FORMAT_QUERY) != MMSYSERR_NOERROR)
```
CendioOssman: Should we log something here?

vncviewer/CConn.cxx
```diff
@@ -104,6 +106,14 @@ CConn::CConn()
   supportsDesktopResize = true;
   supportsLEDState = true;
 
+  // Only ask the server to send audio if there is a device that can
+  // play it. Opening that device is deferred until the server actually
+  // has something for us.
+  if (::audio) {
```
CendioOssman: We try to avoid global namespace markers. Probably better to give this parameter a better variable name to avoid conflicts.

vncviewer/OptionsDialog.cxx
```diff
@@ -1251,6 +1253,12 @@ void OptionsDialog::createMiscPage(int tx, int ty, int tw, int th)
                                                   _("Ask to reconnect on connection errors")));
   ty += CHECK_HEIGHT + TIGHT_MARGIN;
 
+  audioCheckbox = new Fl_Check_Button(LBLRIGHT(tx, ty,
+                                              CHECK_MIN_WIDTH,
+                                              CHECK_HEIGHT,
+                                              _("Play audio from the server")));
+  ty += CHECK_HEIGHT + TIGHT_MARGIN;
```
CendioOssman: Looks like the indentation is off by one here?

vncviewer/parameters.cxx
```diff
@@ -283,6 +288,9 @@ static core::VoidParameter* parameterArray[] = {
   &emulateMiddleButton,
   &alwaysCursor,
   &cursorType,
+  /* Audio */
+  &audio,
```
CendioOssman: The titles here should match the options window. So put it with the other misc stuff.

vncviewer/AudioOutputWin32.cxx
```diff
@@ -46,11 +46,17 @@ static const unsigned minStreamDelayMs = 20;
...
+// 48 kHz rather than 44.1 kHz, and that is deliberate: it is what #1478
+// settled on in its last commit, "Switched to 48 kHz output sample rate
+// ... to avoid downsampling in QEMU for modern Windows guests". QEMU
+// converts from whatever the GUEST produces, so asking for 48 kHz is what
+// spares the common guest a resample -- there is no server-side default
+// that 44.1 kHz would match.
 static const uint8_t preferredSampleFormat = rfb::qemuAudioFormatS16;
 static const uint8_t preferredChannels = 2;
-static const uint32_t preferredFrequency = 44100;
+static const uint32_t preferredFrequency = 48000;
```
CendioOssman: This commit should be merged in to the other one. We don't need this history in the final merge.

vncviewer/AudioOutputPulse.cxx
```diff
@@ -0,0 +1,535 @@
...
+
+// Connecting is the only thing here that waits, because create() has no
+// way to say "ask me again later"
+bool AudioOutputPulse::connect()
+{
+  pa_mainloop_api* api;
+  pa_time_event* timeout;
+  bool ready;
+
+  mainloop = pa_threaded_mainloop_new();
```
CendioOssman, line 117: Normally we avoid threads as they are a never ending source of weird bugs. But I think we can make an exception here as PulseAudio will completely hide the fact that a thread is being used.

vncviewer/AudioOutputPulse.cxx
```diff
@@ -0,0 +1,535 @@
...
+    pa_threaded_mainloop_free(mainloop);
+    mainloop = nullptr;
+    return false;
+  }
+
+  pa_threaded_mainloop_lock(mainloop);
+
+  api = pa_threaded_mainloop_get_api(mainloop);
+
+  context = pa_context_new(api, "TigerVNC Viewer");
```
CendioOssman: We've dropped the "Viewer" part from the description.

CendioOssman: I didn't get the correct icon here, so we might need to provide more context to PulseAudio to map to our .desktop file.

vncviewer/AudioOutputPulse.cxx
```diff
@@ -0,0 +1,535 @@
...
+    return false;
+  }
+
+  pa_context_set_state_callback(context, contextStateCallback, this);
+
+  // Starting a sound server that isn't running is not our place
+  if (pa_context_connect(context, nullptr,
+                         PA_CONTEXT_NOAUTOSPAWN, nullptr) < 0) {
+    vlog.debug("Could not connect to sound server: %s",
+               pa_strerror(pa_context_errno(context)));
```
CendioOssman: Shouldn't this be an error?

vncviewer/AudioOutputPulse.cxx
```diff
@@ -0,0 +1,535 @@
...
+  }
+
+  bufferSize = bufferFree = samples * sampleSize;
+  pendingSize = writtenHead = pendingHead = 0;
+
+  pa_threaded_mainloop_lock(mainloop);
+
+  stream = pa_stream_new(context, "Remote audio", &spec, nullptr);
+  if (stream == nullptr) {
+    vlog.error("Could not create audio playback stream: %s",
```
CendioOssman: This should be translated

CendioOssman: And other calls to `vlog.info()` or `vlog.error()`.

vncviewer/AudioOutputPulse.cxx
```diff
@@ -0,0 +1,535 @@
...
+  AudioOutputPulse* self = (AudioOutputPulse*)userdata;
+
+  // Only FAILED is worth saying anything about: before we are available
+  // this is just the connect above making progress, whose failures
+  // connect() reports itself, and TERMINATED is our own disconnect,
+  // which pa_context_disconnect() reports back to us from right here
+  if (self->available &&
+      (pa_context_get_state(context) == PA_CONTEXT_FAILED))
+    vlog.error("Lost connection to sound server: %s",
+               pa_strerror(pa_context_errno(context)));
```
CendioOssman: You can't touch anything non-PulseAudio as the rest of the TigerVNC code is not thread safe.

vncviewer/AudioOutputPulse.cxx
```diff
@@ -0,0 +1,535 @@
...
+
+  switch (pa_stream_get_state(stream)) {
+  case PA_STREAM_READY:
+    // Anything buffered while the stream was still connecting can go
+    // out now
+    self->submit();
+    break;
+  case PA_STREAM_FAILED:
+    vlog.error("Audio playback stream failed: %s",
+               pa_strerror(pa_context_errno(self->context)));
```
CendioOssman: Is this also on the second thread? If so, this isn't safe either.

CMakeLists.txt
```diff
@@ -208,6 +208,25 @@ if(ENABLE_H264)
   endif()
 endif()
 
+trioption(ENABLE_AUDIO "Enable audio playback in the viewer")
```
CendioOssman, line 211: Please also update the automated builds and packages to enable this option where relevant, and to install the PulseAudio dependency.

common/rfb/CMsgReader.cxx
```diff
@@ -137,6 +137,9 @@ bool CMsgReader::readQEMUServerMessage()
 
   handler->handleQEMUServerMessage(submessage, operation,
                                    is->getptr(length), length);
+
+  // getptr() resets the amount of assured data
+  is->hasData(length);
   is->skip(length);
```
CendioOssman: It's `setptr()` that should be called after `getptr()`.

And please merge this commit with the one it fixes.

CendioOssman (changes requested):
Everything seems to work nicely here. I tested both a Linux client and a Windows client. Nice work! :)

If any of the commits are mainly Mykhailo's work, then it's probably best to tag him as the author in git.

Please also avoid any development backstory in the commits. The message should focus on explaining why things are done the way they are in the commit.

common/rfb/CMsgReader.cxx
jose-pr: Gone with the first of your two options: `CMsgReader` now does all the subtype checking, and `CMsgHandler` has `audioBegin()`, `audioEnd()` and `audioData()` instead of the generic message.

The data-rect shape doesn't buy anything here. Data rects work because the callback can be told how much is left; this extension has no generic length field, so an operation neither side recognises cannot be stepped over by the reader *or* by the handler — the stream is lost either way. Moving the layouts into the handler would have bought the extra state handling you mentioned and no extensibility in return.

Related: an unknown submessage or operation is now a protocol error rather than something to ignore. Ignoring it left the reader positioned wrongly for the next message, which is worse than failing.

jose-pr:
Thanks for the review — that was a lot of detail and all of it useful.

I've rewritten the branch rather than pushing fixups on top, since you asked for the two fixup commits to be merged into the ones they fix. It is now three commits on current master, and the first two are authored to Mikhail, as his work is what they are:

- `Add support for QEMU's audio extension` — Mikhail Kupchik
- `Play audio in the viewer` — Mikhail Kupchik
- `Play audio on Linux, via PulseAudio` — me

The old `Ask for 48 kHz rather than 44.1 kHz` and `Restore the underrun check` commits are folded into the commits they belonged to, and the commit messages no longer carry any development history.

The comments that needed a decision rather than an edit are answered in their own threads. The rest are done as asked, and are listed here rather than as eighteen separate notifications:

**Comments dropped** — `CConnection.h:329`, `CMsgHandler.h:92`, `msgTypes.h:55`, `qemuTypes.h:38`, `AudioOutput.h:31`, `AudioOutputWin32.cxx:24`.

**Style** — no braces on the single statement (`CConnection.cxx:1109`); `requestAudio()` moved so the header matches the .cxx order (`CConnection.h:333`); the indentation (`OptionsDialog.cxx:1260`); the parameter is now `playAudio`, so no namespace marker is needed (`CConn.cxx:112`); the option is listed under Misc with the other window titles (`parameters.cxx:292`).

**Translated** — `AudioOutput.cxx:41`, `AudioOutputWin32.cxx:133`, `AudioOutputPulse.cxx:259` and the other `vlog.info()` and `vlog.error()` calls. Failing to reach the sound server is now an error rather than debug (`AudioOutputPulse.cxx:176`), since it leaves the user with no audio and nothing to explain it.

**Logging added** — `AudioOutput.cxx:47` (debug), and `AudioOutputWin32.cxx:83` for a device that cannot play the format we need, as that is why audio is never offered.

**Squashed** — `AudioOutputWin32.cxx:59`.

I also checked it end to end against a small RFB server that streams the extension without QEMU in the loop, so that the numbers are exact: the client negotiates S16/2ch/48 kHz and consumes a full five second stream (960000 bytes) with no protocol error, and with the pseudo-encoding withheld it never asks for audio at all. The underrun check is live in that build, which is what catches the `getptr()`/`setptr()` pairing below.

common/rfb/CConnection.cxx
jose-pr: Done — the format is now three constants in `CConnection.cxx`, and `getAudioFormat()` is gone. Both backends were returning the same values because both convert whatever they are given, so the virtual only spread one constant across three files.

The frontend still signals with `supportsAudio` alone, per your other comment, and the request is still built in `requestAudio()`, so if a format ever does need to vary that is a change there rather than to the frontend interface.

common/rfb/CConnection.cxx
jose-pr: Added. `framebufferUpdateEnd()` now requires `supportsAudio` as well as `server.supportsQEMUAudio` before asking, so a server that sends the pseudo-encoding rectangle we never advertised is ignored rather than obeyed.

vncviewer/AudioOutputWin32.cxx
jose-pr: Coordinated. Nothing about how far ahead to buffer, or how much silence to play before a stream, depends on which sound system plays the samples, and both backends had identical copies. They are in `AudioOutput.h` now, next to the format constants, and both backends use them.

vncviewer/AudioOutputPulse.cxx
jose-pr: Fixed. The PulseAudio callbacks no longer touch anything outside PulseAudio — `vlog` included. A failure is recorded under the mainloop lock and the main thread reports it on its next call in.

vncviewer/AudioOutputPulse.cxx
jose-pr: It was, yes — and so were the `vlog` calls in `submit()` and in the context state callback, which I had missed. All three go through the same deferred reporting now.

vncviewer/AudioOutputPulse.cxx
jose-pr: Dropped — it is just "TigerVNC" now.

jose-pr: The context now carries `PA_PROP_APPLICATION_ID`, set to `vncviewer` to match the installed `vncviewer.desktop`, which is what PulseAudio maps against. `PA_PROP_APPLICATION_ICON_NAME` is set as well, so the icon still resolves where the desktop file is not installed.

common/rfb/CMsgReader.cxx
jose-pr: Right, thanks. `getptr()` spends the underrun check it makes, so the `skip()` had nothing left to check against, and re-asserting `hasData()` was the wrong way round it. It is a single `setptr(length)` now, and the commit is folded into the one it was fixing.

CMakeLists.txt
jose-pr: Done: `ENABLE_AUDIO=ON` in the Linux and Windows CI jobs, `libpulse-dev` added to the Linux job and to the three Debian control files, and `pulseaudio-libs-devel` plus `-DENABLE_AUDIO=ON` in the three RPM specs. macOS is left alone, as there is no backend for it yet.

vncviewer/AudioOutputPulse.cxx
jose-pr: Thanks. I have kept the thread to PulseAudio's own mainloop rather than starting one of ours, so nothing else in the viewer has to become thread aware.

CendioOssman (approved):
Looks great! I've also retested with a Linux client.

Thanks for doing the work to get this finished!

any1:
I'm a bit late to the party, but I had a look at the already merged code and I have some comments. This should not be taken as criticism; rather suggestions for improvements that could be made in the future, either by myself or anyone else who might want to take this on.

1. Architecture. The platform abstraction layer could be thinner. There is a lot of duplication between Win32 and Pulse.
2. Clock drift is only accounted for in one direction. As far as I can see, if the server sends faster than the receiver consumes audio samples, delay will build up. This is all fine and dandy if you're using this via QEMU on the same host as the clock doesn't drift, but this may become problematic for long running sessions over network.
3. Excessive buffering. The jitter allowance starts at 1 second and only ever goes up. A whole second is rather excessive. The jitter should be measured instead, or at the very least lowered to 200 ms (which is a common figure for this sort of thing).

To properly account for clock drift, the only viable solution that I know of is to measure the source and destination rates and resample the incoming audio using the ratio between the two. One approach to achieve this is to constantly vary the ratio using the target delay error as an input to a PI controller.

For measuring jitter, I've found the most effective method to be a windowed quantile filter. I.e. you pass all latency measurements through a window and find the value below which a certain percentage (e.g. 99%) of the samples lies. I have found this to be slightly more effective than envelope detection and it's certainly more effective than a plain maximum or standard deviation over the window. The quantile method has very good transient rejection.

PR #2146, opened 2026-08-22
Aadvertise servers over DNS-SD using Avahi

### LINKED ISSUE #2143: Advertise servers with MDNS/DNS-SD
I would like to add support for advertising servers over local network using MDNS/DNS-SD. My plan is to use Avahi library for this. Main motivation is to allow easy discovery of the servers in environments where IP address may not be static.

Would this be acceptable/welcome?

gujjwal00:
TODO:
- [x] Add support for publishing services through Avahi
- [x] Advertise x0vncserver
- [x] Advertise w0vncserver
- [x] Advertise Xvnc

@CendioOssman Since one would most likely want a command line argument to control this, how should  I handle that (in code & in man page) when servers are not compiled with Avahi support? One option is to  keep the parameter always visible in help and just log an error at runtime.

And I may have underestimated the work required to integrate with event loops of servers :)
All three servers are using different schemes: for loop in x0vncservers, Glib in w0vncserver, and I am still figuring out how everything is wired up in Xvnc. Using threaded poll api from Avahi would have simplified the implementation a lot.  

Closes #2143


CendioOssman:
Yeah, socket management is currently a pain. I've been planning to write some abstraction in the core libraries to hide the details. Some day. :)

common/network/DnsSD.h
```diff
@@ -0,0 +1,53 @@
...
+ * USA.
+ */
+
+#ifndef __NETWORK_DNS_SD_H__
+#define __NETWORK_DNS_SD_H__
+
+#include <list>
+#include <string>
+
+#include <network/Socket.h>
```
CendioOssman: This doesn't seem to be directly used? Is it because of `SocketListener`? That can have a forward declaration.

common/network/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
 endif()
 
 add_library(network STATIC
+  DnsSD.cxx
```
CendioOssman: Only add this conditionally. That avoids having a bunch of `#ifdef` in the source code.

common/network/DnsSD.h
```diff
@@ -0,0 +1,53 @@
...
+
+struct AvahiPoll;
+
+namespace network
+{
+  namespace dnssd
+  {
```
CendioOssman: Why an extra namespace?

CMakeLists.txt
```diff
@@ -446,6 +446,30 @@ if(UNIX AND NOT APPLE)
...
+# Check for Avahi library to enable DNS-SD support
+if(UNIX AND NOT APPLE)
+  trioption(ENABLE_AVAHI "Enable DNS-SD support using Avahi")
+  if(ENABLE_AVAHI)
+    if(ENABLE_AVAHI STREQUAL "AUTO")
+      find_package(AvahiClient)
+      if(ENABLE_WAYLAND)
+        find_package(AvahiGlib)  
+      endif()
+    else()
```
CendioOssman: It's probably more clear to include this in the commit that actually uses it.

PR #2153, opened 2026-09-02
Size the Windows H.264 decoded buffer for the actual stream

heidrickla:
## Summary

On Windows, H.264 rects larger than about 1920x1088 decode to nothing: the
viewer connects, `FramebufferUpdate`s arrive at full rate, and the window
stays black. 1920x1080 streams render normally.

`H264WinDecoderContext` allocates `decoded_buffer` once, in the
constructor, from `GetOutputStreamInfo()` on the decoder's initial
placeholder output type. That buffer is sized for the placeholder
(4,147,200 bytes on Windows 11), not for the stream that arrives later. A
2560x1440 NV12 frame needs 5,529,600 bytes, so once the real stream is
known every `ProcessOutput()` fails with `E_FAIL` (0x80004005), the drain
loop has no case for that result, and no frame is ever converted or drawn.

This change:

- re-sizes `decoded_buffer` when the decoder reports
  `MF_E_TRANSFORM_STREAM_CHANGE`, from the output stream info it reports at
  that point;
- releases the previous converter output buffer before creating the new
  one on a stream change, instead of adding a second buffer to the sample;
- breaks out of the output loop on an unexpected `ProcessOutput()` result
  and logs it, rather than looping forever;
- logs a failed `ProcessInput()` at debug level instead of returning silently.

## How it was found and verified

Server: kvmd-vnc (PiKVM) on a GL.iNet RM10, an RV1126-based KVM whose
capture is 2560x1440@60 (and 3840x2160@30 with a 4K EDID), H.264 High
profile level 5.0/5.1.

- Replayed the unit's captured stream to `vncviewer` 1.16.2 from a small
  RFB server that frames it as kvmd-vnc does: black. A 1920x1080 x264
  transcode of the same content: renders. A 2560x1440 Main-profile
  transcode: black. So the profile is irrelevant and the size is not.
- With logging added at the swallowed error paths, the 1440p run shows the
  stream change reporting 2560x1440 and then `ProcessOutput()` failing
  with `E_FAIL` (0x80004005) on every frame.
- With this patch: the 1440p replay renders, a live 2560x1440@60 session
  renders (~59 updates/s), and a live 3840x2160@30 session renders (the
  buffer grows to 16,588,800 bytes). Raising
  `CODECAPI_AVDecVideoMaxCodedWidth/Height` on the MFT was tried first and
  makes no difference; the decoder accepted the larger stream all along.

## Regression test

`tests/unit/h264decoder.cxx` (second commit) decodes a flat grey IDR frame
at 1920x1080 and at 2560x1440 through `H264Decoder` into a black
`ManagedPixelBuffer` and checks three pixels. The frames are embedded (458
and 751 bytes, SPS+PPS+IDR from x264). Built with `ENABLE_H264` on Windows:

- on master without the fix: `decode1080p` passes, `decode1440p` fails
  with all three pixels still 0;
- with the fix: both pass (about 40 ms each).

The test is portable and runs wherever `ENABLE_H264` is on; I have only
run it against the Media Foundation path here.

Built with MSYS2 MinGW-w64 (GCC 16.2) on Windows 11 24H2, FLTK 1.3.9.


common/rfb/H264WinDecoderContext.cxx
```diff
@@ -31,11 +31,15 @@
...
 #include <core/i18n.h>
 
+#include <core/LogWriter.h>
+
 #include <rfb/PixelBuffer.h>
 #include <rfb/H264WinDecoderContext.h>
 
 using namespace rfb;
 
+static core::LogWriter vlog("H264WinDecoderContext");
```
CendioOssman, line 41: Unfortunately, decoders can't currently safely log because of threading.

common/rfb/H264WinDecoderContext.cxx
```diff
@@ -161,9 +165,11 @@ void H264WinDecoderContext::decode(const uint8_t* h264_buffer,
...
 
-  if (FAILED(decoder->ProcessInput(0, input_sample, 0)))
+  HRESULT hr = decoder->ProcessInput(0, input_sample, 0);
+  if (FAILED(hr))
   {
-    // Silently ignore errors, hoping its a temporary encoding glitch
+    // Hopefully a temporary encoding glitch
+    vlog.debug("ProcessInput() failed: 0x%08lx", (unsigned long)hr);
     return;
   }
```
CendioOssman, line 174: This change looks like a no-op?

common/rfb/H264WinDecoderContext.cxx
CendioOssman, line 118: I don't quite understand where the initial size comes from. We don't give `decoder` any information more than input and output types.

Why hasn't thin been an issue before?

CendioOssman (changes requested):
Thanks for this fix! There is one build failure and some minor adjustments. Then we can merge this.

common/rfb/H264WinDecoderContext.cxx
heidrickla: Removed, along with the two log calls.

common/rfb/H264WinDecoderContext.cxx
heidrickla: It was. Reverted. Its extra `hr` was also what tripped -Werror=shadow on the loop’s own `hr` in the Windows build.

common/rfb/H264WinDecoderContext.cxx
heidrickla: The constructor calls GetOutputStreamInfo() before any input has been seen, and at that point the MFT’s output type still has its default 1920x1080 frame size, so cbSize is 4,147,200 (1920×1080×2 on Windows 11). Anything up to 1080p fits, which is why it never showed up. At 2560x1440 the frame doesn’t fit, ProcessOutput() returns E_FAIL on every call and nothing is drawn. Reworded the comment to say that.

heidrickla:
Pushed the revised commits (same two, amended): logging removed, the ProcessInput() change reverted, comment reworded. Built with the CI’s Debug flags locally: the previous revision reproduces the -Werror=shadow error, master’s decoder fails the 1440p test, this revision passes both.

PR #2156, opened 2026-09-07
Add fadeout before idle timer is triggered

CendioHalim:
This PR adds a fadeout timer if the `idleTimer` parameter is used. It will start fading the screen to black 10 seconds before the idle timer is triggered. The idea is to give users a chance to prevent being disconnected if they notice the screen fading.

Tested with `Xvnc`, `x0vncserver` and `w0vncserver`. During the fading animation, if nothing else is happening on screen, I could see ~20% CPU usage with `Xvnc` on my machine.

I had a quick look at what durations other desktop environments have for this, and it seemed to range between 5-10 seconds.

I left out fading the cursor in this PR. Instead, the cursor will be invisible during the fading animation.

### DIFF
--- a/common/rfb/FadingPixelBuffer.cxx
+++ b/common/rfb/FadingPixelBuffer.cxx
@@ -1,0 +1,178 @@
+/* Copyright 2026 Adam Halim for Cendio AB
+ *
+ * This is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this software; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ * USA.
+ */
+
+#include <assert.h>
+
+#include <pixman.h>
+#include <stdexcept>
+#include <vector>
+
+#include <core/LogWriter.h>
+#include <core/Region.h>
+#include <core/i18n.h>
+
+#include "FadingPixelBuffer.h"
+
+using namespace rfb;
+
+static const PixelFormat pfRGBX(32, 24, false, true, 255, 255, 255, 0, 8, 16);
+static const PixelFormat pfBGRX(32, 24, false, true, 255, 255, 255, 16, 8, 0);
+static const PixelFormat pfXRGB(32, 24, false, true, 255, 255, 255, 8, 16, 24);
+static const PixelFormat pfXBGR(32, 24, false, true, 255, 255, 255, 24, 16, 8);
+
+static core::LogWriter vlog("FadingPixelBuffer");
+
+FadingPixelBuffer::FadingPixelBuffer(const PixelBuffer* pb)
+  : PixelBuffer(pb->getPF(), pb->width(), pb->height()),
+    parent(pb), fadeLevel(1.0f),
+    fadedBuffer(new uint8_t[width() * height() * (format.bpp/8)])
+{
+}
+
+FadingPixelBuffer::~FadingPixelBuffer()
+{
+  delete[] fadedBuffer;
+}
+
+void FadingPixelBuffer::setFadeLevel(float level)
+{
+  fadeLevel = level < 0.0f ? 0.0f : (level > 1.0f ? 1.0f : level);
+}
+
+const uint8_t* FadingPixelBuffer::getBuffer(const core::Rect& r,
+                                            int* stride_) const
+{
+  *stride_ = width();
+  return fadedBuffer + (r.tl.y * width() + r.tl.x) * format.bpp / 8;
+}
+
+void FadingPixelBuffer::syncBuffers(const core::Region& r)
+{
+  const uint8_t* buffer;
+  int stride_;
+  int bpp;
+  pixman_format_code_t pixmanFormat;
+  pixman_image_t* srcImg;
+  pixman_image_t* dstImg;
+  uint16_t alpha;
+  pixman_color_t black;
+  pixman_image_t* overlay;
+  std::vector<core::Rect> rects;
+
+  assert(fadeLevel <= 1.0f && fadeLevel >= 0.0f);
+
+  if (!r.get_rects(&rects))
+    return;
+
+  if (parent->getRect().width() != width() || parent->getRect().height() != height())
+    throw std::runtime_error(_("Parent PixelBuffer size mismatch"));
+
+  if (format != parent->getPF())
+    throw std::runtime_error(_("Parent PixelBuffer format mismatch"));
+
+  try {
+    pixmanFormat = rfbToPixmanFormat(format);
+  } catch (std::runtime_error&) {
+    syncBuffersFallback(r);
+    return;
+  }
+
+  buffer = parent->getBuffer(parent->getRect(), &stride_);
+  bpp = format.bpp / 8;
+
+  alpha = (1 - fadeLevel) * 0xffff;
+  black = {0, 0, 0, alpha};
+  overlay = pixman_image_create_solid_fill(&black);
+
+  srcImg = pixman_image_create_bits(pixmanFormat, width(), height(),
+                                    (uint32_t *)buffer, stride_ * bpp);
+  dstImg = pixman_image_create_bits(pixmanFormat, width(), height(),
+                                    (uint32_t *)fadedBuffer, width() * bpp);
+  for (core::Rect &rect : rects) {
+    pixman_image_composite32(PIXMAN_OP_SRC, srcImg, nullptr, dstImg,
+                             rect.tl.x, rect.tl.y, 0, 0, rect.tl.x,
+                             rect.tl.y, rect.width(), rect.height());
+
+    pixman_image_composite32(PIXMAN_OP_OVER, overlay, nullptr, dstImg,
+                             rect.tl.x, rect.tl.y, 0, 0, rect.tl.x,
+                             rect.tl.y, rect.width(), rect.height());
+  }
+
+  pixman_image_unref(overlay);
+  pixman_image_unref(srcImg);
+  pixman_image_unref(dstImg);
+}
+
+void FadingPixelBuffer::syncBuffersFallback(const core::Region& r)
+{
+  const uint8_t* buffer;
+  int stride_;
+  int bpp;
+  std::vector<core::Rect> rects;
+
+  assert(fadeLevel <= 1.0f && fadeLevel >= 0.0f);
+
+  if (!r.get_rects(&rects))
+    return;
+
+  if (parent->getRect().width() != width() || parent->getRect().height() != height())
+    throw std::runtime_error(_("Parent PixelBuffer size mismatch"));
+
+  if (format != parent->getPF())
+    throw std::runtime_error(_("Parent PixelBuffer format mismatch"));
+
+  buffer = parent->getBuffer(parent->getRect(), &stride_);
+  bpp = format.bpp / 8;
+
+  for (core::Rect &rect : rects) {
+    std::vector<uint8_t> rgb(rect.width() * 3);
+
+    for (int y = 0; y < rect.height(); y++) {
+      const uint8_t* srcRow;
+      uint8_t* dstRow;
+
+      srcRow = buffer + ((rect.tl.y + y) * stride_ + rect.tl.x) * bpp;
+      dstRow = fadedBuffer + ((rect.tl.y + y) * width() + rect.tl.x) * bpp;
+
+      format.rgbFromBuffer(rgb.data(), srcRow, rect.width());
+
+      for (int x = 0; x < rect.width(); x++) {
+        rgb[x*3 + 0] = (uint8_t)(rgb[x*3 + 0] * fadeLevel);
+        rgb[x*3 + 1] = (uint8_t)(rgb[x*3 + 1] * fadeLevel);
+        rgb[x*3 + 2] = (uint8_t)(rgb[x*3 + 2] * fadeLevel);
+      }
+
+      format.bufferFromRGB(dstRow, rgb.data(), rect.width());
+    }
+  }
+}
+
+pixman_format_code_t FadingPixelBuffer::rfbToPixmanFormat(PixelFormat pf)
+{
+  if (pf == pfBGRX)
+    return PIXMAN_x8r8g8b8;
+  if (pf == pfRGBX)
+    return PIXMAN_x8b8g8r8;
+  if (pf == pfXBGR)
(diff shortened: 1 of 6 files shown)


CendioHalim:
I also made it trigger on `maxIdleTime`. It will use the lowest/closes timeout for each connection (presumably `idleTimeout`...).

Tested with multiple clients and combinations of `idleTimout` and `maxIdleTime` and did not see any issues.

common/rfb/FadingPixelBuffer.cxx
```diff
@@ -0,0 +1,178 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this software; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307,
+ * USA.
+ */
+
+#include <assert.h>
+
+#include <pixman.h>
```
CendioOssman, line 21: We haven't been using pixman directly in the rfb library until this. It's best if it's also added to `CMakeLists.txt` so we don't rely on side-effects from other libraries.

common/rfb/FadingPixelBuffer.cxx
```diff
@@ -0,0 +1,178 @@
...
+static const PixelFormat pfBGRX(32, 24, false, true, 255, 255, 255, 16, 8, 0);
+static const PixelFormat pfXRGB(32, 24, false, true, 255, 255, 255, 8, 16, 24);
+static const PixelFormat pfXBGR(32, 24, false, true, 255, 255, 255, 24, 16, 8);
+
+static core::LogWriter vlog("FadingPixelBuffer");
+
+FadingPixelBuffer::FadingPixelBuffer(const PixelBuffer* pb)
+  : PixelBuffer(pb->getPF(), pb->width(), pb->height()),
+    parent(pb), fadeLevel(1.0f),
+    fadedBuffer(new uint8_t[width() * height() * (format.bpp/8)])
```
CendioOssman, line 43: This can be a non-trival amount of memory for large sessions. Did you explore any method of reducing this cost?

CendioOssman, line 43: Why not `ManagedPixelBuffer` if we're doing our own allocation anyway?

common/rfb/FadingPixelBuffer.cxx
```diff
@@ -0,0 +1,178 @@
...
+  assert(fadeLevel <= 1.0f && fadeLevel >= 0.0f);
+
+  if (!r.get_rects(&rects))
+    return;
+
+  if (parent->getRect().width() != width() || parent->getRect().height() != height())
+    throw std::runtime_error(_("Parent PixelBuffer size mismatch"));
+
+  if (format != parent->getPF())
+    throw std::runtime_error(_("Parent PixelBuffer format mismatch"));
```
CendioOssman, line 86: If these are actually programmer bugs, then asserts might be more appropriate. Or `std::logic_error`. At which point we also don't translate it.

common/rfb/FadingPixelBuffer.cxx
```diff
@@ -0,0 +1,178 @@
...
+
+  try {
+    pixmanFormat = rfbToPixmanFormat(format);
+  } catch (std::runtime_error&) {
```
CendioOssman, line 90: Exceptions aren't ideal for expected stuff. Is there no value we can use instead to indicate "unknown"?

common/rfb/FadingPixelBuffer.cxx
```diff
@@ -0,0 +1,178 @@
...
+  dstImg = pixman_image_create_bits(pixmanFormat, width(), height(),
+                                    (uint32_t *)fadedBuffer, width() * bpp);
+  for (core::Rect &rect : rects) {
+    pixman_image_composite32(PIXMAN_OP_SRC, srcImg, nullptr, dstImg,
+                             rect.tl.x, rect.tl.y, 0, 0, rect.tl.x,
+                             rect.tl.y, rect.width(), rect.height());
+
+    pixman_image_composite32(PIXMAN_OP_OVER, overlay, nullptr, dstImg,
+                             rect.tl.x, rect.tl.y, 0, 0, rect.tl.x,
+                             rect.tl.y, rect.width(), rect.height());
```
CendioOssman, line 113: Is it possible to use the mask and do this in one pass?

samhed:
Very nice!

I wonder if the animation could be tweaked a bit somehow? I get the feeling that you'd want things to feel even darker just the last second or so. I don't know exactly what I'm basing my expectations on, but it feels like it disconnects (with IdleTimeout) in the middle of the fading process.

common/rfb/VNCSConnectionST.cxx
```diff
@@ -59,6 +60,9 @@ using namespace rfb;
 static const unsigned LOGIN_GRACE_TIME = 120;
 // Number of seconds allowed to flush a closing socket
 static const unsigned CLOSE_GRACE_TIME = 5;
+// Number of seconds the framebuffer will fade to black before
```
CendioOssman, line 63: A blank line here could be helpful to group related constants.

common/rfb/VNCSConnectionST.h
```diff
@@ -202,6 +204,7 @@ namespace rfb {
     std::map<uint32_t, uint32_t> pressedKeys;
 
     core::Timer idleTimer;
+    core::Timer fadeTimer;
```
CendioOssman, line 207: We often try to group on function, not type. So you can put the buffer and timer together if you want.

common/rfb/VNCSConnectionST.cxx
```diff
@@ -492,9 +503,16 @@ void VNCSConnectionST::queryConnection(const char* userName)
...
-  if (rfb::Server::idleTimeout)
+  if (rfb::Server::idleTimeout) {
     idleTimer.start(core::secsToMillis(rfb::Server::idleTimeout));
 
+    // Trigger the timer just as we are about to start fading
+    if (idleTimer.getRemainingMs() < FADING_TIME * 1000.0f)
+      fadeTimer.start(1000 / rfb::Server::frameRate);
+    else
+      fadeTimer.start(idleTimer.getRemainingMs() - FADING_TIME * 1000.0f);
+  }
```
CendioOssman, line 514: Don't you need to update this timer every place where we change `idleTimer`? Which probably means we want a helper method.

common/rfb/VNCSConnectionST.cxx
```diff
@@ -858,6 +876,54 @@ void VNCSConnectionST::handleTimeout(core::Timer* t)
...
+        if (!fadedBuffer) {
+          fadedBuffer = new FadingPixelBuffer(server->getPixelBuffer());
+          needsUpdate = true;
+        }
+
+        oldLevel = fadedBuffer->getFadeLevel();
+        newLevel = (idleTimer.getRemainingMs() / (FADING_TIME * 1000.0f));
+
+        // FIXME: Don't hardcode 255
+        if ((int)(oldLevel*255) != (int)(newLevel*255)) {
```
CendioOssman, line 897: Why not a direct comparison?

common/rfb/VNCSConnectionST.cxx
```diff
@@ -858,6 +876,54 @@ void VNCSConnectionST::handleTimeout(core::Timer* t)
...
+      if (idleTimer.getRemainingMs() < FADING_TIME * 1000.0f) {
+        bool needsUpdate;
+        float oldLevel;
+        float newLevel;
+
+        needsUpdate = false;
+
+        if (!fadedBuffer) {
+          fadedBuffer = new FadingPixelBuffer(server->getPixelBuffer());
+          needsUpdate = true;
```
CendioOssman, line 890: Is this needed? Won't it sync relevant pixels later anyway?

common/rfb/VNCSConnectionST.cxx
```diff
@@ -1068,6 +1134,29 @@ void VNCSConnectionST::writeDataUpdate()
...
   }
 
+  if (fadedBuffer) {
+    float oldFadeLevel;
+    float newFadeLevel;
+
+    newFadeLevel = (idleTimer.getRemainingMs() / (FADING_TIME * 1000.0f));
+    oldFadeLevel = fadedBuffer->getFadeLevel();
+
+    if ((int)(oldFadeLevel * 255) != (int)(newFadeLevel * 255)) {
```
CendioOssman, line 1144: This seems a bit duplicated. Can't we just do this in one of the two places?

common/rfb/VNCSConnectionST.cxx
```diff
@@ -1068,6 +1134,29 @@ void VNCSConnectionST::writeDataUpdate()
...
+      updates.clear();
+      updates.add_changed(server->getPixelBuffer()->getRect());
```
CendioOssman, line 1146: I think this is dangerous to change this late. I don't think it will interact well with the cursor.

common/rfb/VNCSConnectionST.cxx
```diff
@@ -1068,6 +1134,29 @@ void VNCSConnectionST::writeDataUpdate()
...
+    }
+
+    try {
+      fadedBuffer->syncBuffers(ui.changed);
+    } catch (std::exception& e) {
+      vlog.error(_("Failed to sync faded buffer: %s"), e.what());
+      delete fadedBuffer;
+      fadedBuffer = nullptr;
+    }
+  }
```
CendioOssman, line 1158: What errors do we expect? Can't we let them propagate like other errors?

common/rfb/VNCSConnectionST.cxx
```diff
@@ -1068,6 +1134,29 @@ void VNCSConnectionST::writeDataUpdate()
...
+
+    if ((int)(oldFadeLevel * 255) != (int)(newFadeLevel * 255)) {
+      updates.clear();
+      updates.add_changed(server->getPixelBuffer()->getRect());
+      fadedBuffer->setFadeLevel(newFadeLevel);
+      updates.getUpdateInfo(&ui, req);
+    }
+
+    try {
+      fadedBuffer->syncBuffers(ui.changed);
```
CendioOssman, line 1152: What about `ui.copied` when fading?

common/rfb/VNCSConnectionST.cxx
```diff
@@ -1078,7 +1167,11 @@ void VNCSConnectionST::writeDataUpdate()
 
   writeRTTPing();
 
-  encodeManager.writeUpdate(ui, server->getPixelBuffer(), cursor);
+  if (fadedBuffer)
+    // FIXME: Fade the cursor as well
```
CendioOssman, line 1171: Should probably be a draft until this FIXME is done.

common/rfb/VNCSConnectionST.cxx
```diff
@@ -1155,8 +1248,23 @@ void VNCSConnectionST::writeLosslessRefresh()
 
   writeRTTPing();
 
-  encodeManager.writeLosslessRefresh(req, server->getPixelBuffer(),
-                                     cursor, maxUpdateSize);
+  if (fadedBuffer) {
+    try {
+      fadedBuffer->syncBuffers(req);
```
CendioOssman, line 1253: This is probably not needed in many cases since we are resending things when we are doing a lossless refresh. But tracking that is not trivial.

Perhaps put a FIXME for future improvement?

common/rfb/VNCServerST.h
```diff
@@ -117,6 +117,7 @@ namespace rfb {
     const char* getName() const { return name.c_str(); }
     unsigned getLEDState() const { return ledState; }
     bool isDesktopReady() const { return desktopStarted; }
+    int getIdleRemainingMs() { return idleTimer.getRemainingMs(); }
```
CendioOssman, line 120: It's a shame we have to add more connections between the objects. We can't make a reasonable guess? It will have the same trigger as `idleTimeout`, so can't we just compare the difference between that and `maxIdleTime`?

CendioOssman (changes requested):
Very nice! This fixes a big usability annoyance. Great work!

A quick test run shows some odd bug in the smoothness of the fade. It's really smooth to begin with, and then it starts slowing down and gets a very stuttering look at the end. A known issue?

CendioOssman:
> I wonder if the animation could be tweaked a bit somehow? I get the feeling that you'd want things to feel even darker just the last second or so. I don't know exactly what I'm basing my expectations on, but it feels like it disconnects (with IdleTimeout) in the middle of the fading process.

We might want to go completely black a bit before disconnecting/terminating. A normal monitor doesn't disappear, so that part is a bit unexpected.

common/rfb/FadingPixelBuffer.cxx
CendioHalim: `pixman_format_code_t` does not have any "unknown" types unfortunately. I'll change the method to return true/false instead.

CendioHalim:
> > I wonder if the animation could be tweaked a bit somehow? I get the feeling that you'd want things to feel even darker just the last second or so. I don't know exactly what I'm basing my expectations on, but it feels like it disconnects (with IdleTimeout) in the middle of the fading process.
> 
> We might want to go completely black a bit before disconnecting/terminating. A normal monitor doesn't disappear, so that part is a bit unexpected.

Yeah I added one second after the screen is completely black, it looks and feels much better.

common/rfb/FadingPixelBuffer.cxx
CendioHalim: Good point! Switched to `ManagedPixelBuffer`

CendioHalim: Did not play around with this. It likely would require shifting some of the logic to the `EncodeManager` which probably should be done anyway. I'll have a look at it once the logic is moved.

common/rfb/FadingPixelBuffer.cxx
CendioHalim: Fair, changed to asserts.

common/rfb/FadingPixelBuffer.cxx
CendioHalim: It is! 
```cpp
for (core::Rect &rect : rects) {
    pixman_image_composite32(PIXMAN_OP_SRC, srcImg, overlay, dstImg,
                             rect.tl.x, rect.tl.y, 0, 0, rect.tl.x,
                             rect.tl.y, rect.width(), rect.height());
  }
```
This did the trick, just had the inverse the alpha for the mask

common/rfb/VNCSConnectionST.cxx
CendioHalim: This was to guard against changing the fade level too often. It doesn't *really* make sense to keep it at 255 (there are way more distinct levels where at least *some* of the pixels will change). I think it would make more sense to just keep it tied to the timer's frequency regardless. 30 fps (or server's framerate / 2) looks smooth enough IMO

common/rfb/VNCSConnectionST.cxx
CendioHalim: We shouldn't throw anything from `FadingPixelBuffer` any more, so I'll just remove the block

common/rfb/VNCSConnectionST.cxx
CendioHalim: We had a chat about this, I don't think that this will be an issue. Did some quick testing with some `copied` changes and could not see any glitches. Will do a deeper investigation once this has been re-written a bit

Issue #2145, opened 2026-08-19
Viewer renders oversaturated on wide-gamut displays (macOS)


rubens-lopes:
## Environment

- TigerVNC viewer 1.16.2, built from source against FLTK 1.3.11
- macOS 26.0 (Darwin 25.5.0), Apple Silicon
- Display: Alienware AW3225QF, QD-OLED, roughly 99% DCI-P3
- Server: wayvnc 0.10.1 on Arch/Hyprland, serving ordinary sRGB content

## What happens

Everything the viewer draws comes out more saturated than it should be. Skin
tones and muted colours are where you notice it. Fully saturated primaries look
fine, which is part of why it took me a while to trust that anything was wrong.

The server is not involved. Capturing the remote framebuffer directly with
`grim` returns the authored sRGB values exactly, so the pixels are correct right
up until the viewer draws them.

## Measurement

I put a solid sRGB `(198,134,66)` on the remote desktop and screenshotted the
viewer with `screencapture -x -o`. That embeds the *display* profile, so the
file holds raw framebuffer values: unmanaged content sits on the authored
number, managed content is shifted.

| build | framebuffer value | comment |
|---|---|---|
| stock 1.16.2 | `(215,127,52)` | pushed 17 levels away from neutral |
| patched `cocoa_win_color_space` only | `(198,133,65)` | raw passthrough |
| patched FLTK only | `(198,133,65)` | raw passthrough |
| both patched | `(185,137,75)` | correct |

For reference, a plain FLTK program drawing the same triplet into a window on
this display produces `(185,138,76)`. The last row agrees with that to within
one level, and the ±1 throughout is JPEG encoding on the wire.

Note the first row. Stock is not merely skipping a conversion, it is applying
one in the wrong direction: the value ends up further from neutral than the
data it started with.

## Cause

`vncviewer/cocoa.mm`:

```objc
CGColorSpaceRef cocoa_win_color_space(Fl_Window *win)
{
  ...
  return [nscs CGColorSpace];   // the display's own profile
}
```

`Surface_OSX.cxx` uses that as the *source* colour space for the framebuffer
image. The framebuffer holds sRGB, so labelling it with the display profile
tells CoreGraphics the numbers are already in display space. The intended
sRGB-to-display conversion therefore never happens; what runs instead is
whatever transform CoreGraphics derives from those mislabelled endpoints, and
measurably that pushes values outward rather than leaving them alone.

## The obvious fix is not sufficient

[...]

PR #2157, opened 2026-09-14
Get macOS color space from graphics context

CendioOssman:
FLTK has an intermediate graphics context between us and the actual window. And since FLTK 1.3.6, that graphics context has a different color space. Make sure we grab the right information or we'll spend a lot of CPU converting between color spaces.


CendioOssman:
Should fix the most obvious bug from #2145.

